PR #154, opened 2013-03-12
Support @JsonAutoDetect to enable/disable methods and fields.

shawnlauzon:
The current system uses only public fields and methods, and there isn't
a very good way of changing this. This commit adds support for
@JsonAutoDetect, which is used by Jackson to configure what types of
fields and methods are supported.

Caveats:
* Annotation is only supported at class level
* The creatorVisibility is not supported
* Visibility.PUBLIC_ONLY is the default visibility, which is different
  from Jackson, but more consistent with the existing Swagger.

I considered instead using XmlAccessType, but since Swagger supports
both Jackson and JAXB annotations, decided to use the one that worked
better for us.

Finally, I did not add unit tests for this. If this seems like a good
direction, I will write some unit tests and include in this commit.


fehguy:
This is now fully handled via jackson in 1.5.x

Issue #360, opened 2013-10-24
Can't mix and match polymorphic annotation declarations


jknova:
If I have a base model declared in Java, with an ApiModel annotation, and some jackson annotations:

```
@JsonTypeInfo(use=JsonTypeInfo.Id.NAME, include=JsonTypeInfo.As.PROPERTY, property="type", defaultImpl=Animal.class)
@JsonSubTypes({  
    @Type(value = Cat.class, name = "cat"),  
    @Type(value = Dog.class, name = "dog") })
@ApiModel(value="Animal description")
public class AttributeSource
{
...
}
```

Then the model's 'subTypes' and 'discriminator' properties are empty. I need to explicitly define them in the ApiModel annotation or remove the ApiModel annotation

PR #362, opened 2013-10-25
Update swagger-core to return subType models for all dependencies and models defined in a Map

jknova:
Hello,
I wanted to address a few things in my fork, mainly around supporting model sub types and thought it might be useful to push them back to the main project. 

My changes are:
1. Fix up the annotation handling for a model's discriminator and subType fields
2. Allow all sub type models to be included if they are referenced as a dependent field
3. Include the model for Map[String,MyModel] field types, so the value type of Map would be returned as a model as well (i.e. MyModel)

For 1) this is my fix for #360 so the annotation JsonTypeInfo.property plays nice with ApiModel.discriminator, and JsonSubTypes plays nices with ApiModel.subTypes

For 2) If I had the models

```
@ApiModel(value="a model with subtypes", discriminator = "name", subTypes = Array(classOf[DomesticAnimal], classOf[WildAnimal]))
case class Animal (
  @(ApiModelProperty @field)(value = "name of animal", position = 1) name: String,
  @(ApiModelProperty @field)(value = "date added", position = 2) date: java.util.Date)

case class DomesticAnimal (
  @(ApiModelProperty @field)(value = "name of animal", position = 1) name: String,
  @(ApiModelProperty @field)(value = "animals are safe for children", position = 2) safeForChildren: Boolean,
  @(ApiModelProperty @field)(value = "date added", position = 3) date: java.util.Date)

case class WildAnimal (
  @(ApiModelProperty @field)(value = "name of animal", position = 1) name: String,
  @(ApiModelProperty @field)(value = "location found in", position = 2) foundInLocation: String,
  @(ApiModelProperty @field)(value = "date added", position = 3) date: java.util.Date)

case class Cage (
  @(ApiModelProperty @field)(value = "caged animal", position = 1) name: Animal)
```

Where I have a polymorphic field in Cage, and Cage is my top level model, then the sub type models for Animal were not generated. Maybe this is by design to prevent too many models, but for our purposes we wanted to include these sub types. bca90dacca7e655b470cfbb80e60339cac42dfb4 is my attempt to enhance this.

For 3) I also wanted the value type for maps to be included as well. Previously if we had a model with the field

```
myMap: Map[String, MyModel]
```

Then MyModel was not being included in the API listing models. I tried to enhance this as well.

### DIFF
--- a/modules/swagger-core/src/main/scala/com/wordnik/swagger/converter/ModelConverters.scala
+++ b/modules/swagger-core/src/main/scala/com/wordnik/swagger/converter/ModelConverters.scala
@@ -9,7 +9,7 @@ import scala.collection.mutable.{ ListBuffer, LinkedHashMap, HashSet, HashMap }
 
 object ModelConverters {
   private val LOGGER = LoggerFactory.getLogger(ModelConverters.getClass)
-  val ComplexTypeMatcher = "([a-zA-Z]*)\\[([a-zA-Z\\.\\-]*)\\].*".r
+  val ComplexTypeMatcher = "([a-zA-Z]*)\\[([a-zA-Z\\.\\-,]*)\\].*".r
 
   val converters = new ListBuffer[ModelConverter]() ++ List(
     new JodaDateTimeConverter,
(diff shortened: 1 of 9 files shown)


webron:
While this is great work, map support is done natively in swagger-core 1.5.X.

PR #413, opened 2014-01-07
Added non-working sample test for JBoss7 with empty web.xml

rage-shadowman:
A non-working sample using JBoss7 and RESTEasy with an empty web.xml file per issue https://github.com/wordnik/swagger-core/issues/412

Everything starts up fine, you can hit the resources (ie: localhost:port/sample/api/lowercase) and api-docs returns some information which does not include the Api resources.


fehguy:
I think we can close this out as the issue was reflections.

PR #522, opened 2014-04-04
fix incorrect log level

bigwheel:
This log text fits info level.


webron:
Please resubmit against the develop branch.

PR #531, opened 2014-04-15
Develop

MiguelGL:
Model Properties attributes `notes` and `access` where not properly taken into account. Changes in this request do. Thanks for all the swagger awesomeness!


fehguy:
This should be supported in 1.5.x, in the develop_2.0 branch.

PR #545, opened 2014-04-30
Use SVG version for badges

christianv:
Instead of using .png, use SVG version for badges


fehguy:
change made manually in develop_2.0 branch

PR #574, opened 2014-05-21
XmlEnumValue annotation doesn't override the default value

e28eta:
We're using @XmlEnumValue to override the string that our enum values serialize/deserialize as.

Swagger is using the Java enum names, and it seems to be ignoring the XmlEnumValue annotation. This pull request changes the test slightly to expose this issue. On my machine, this causes the test case to fail. I don't know what it'll take to fix.

Our particular XmlEnumValue annotations also change the case to lowercase, but I didn't think that was necessary here.


fehguy:
Hi, please try upgrading to 1.5.x. We're putting 1.3.x in maintenance mode.

PR #602, opened 2014-06-20
Feature/issue #542

stankevichevg:
Hi! Recently I've faced the problem with ignored annotation @ApiModelProperty, the same as it discussed [here](https://github.com/wordnik/swagger-core/issues/542). As I learned from the code the issue related to the logic in ModelPropertyParser where it is process annotation only for public fields and methods. My solution is to enable annotation proccessing for such private fields which have public acessor method.


fehguy:
closing as this this should be supported in 1.5.x

PR #626, opened 2014-07-17
Migrated to Play 2.3.x and Scala 2.10.4

bkowalik:
This upgrade breaks backward compatibility with Play because SimpleResult is now out from API. Because of that I've also increased vertsion to 1.4.0. Scala 2.10.4 does not break antyhing.


fehguy:
This has been done in 1.3.12

PR #635, opened 2014-07-31
Add getRootResource method to avoid multiple routes

chafikhnini:
This method avoids to write multiple repetitive routes for resources like
```
GET     /api-docs.json/pet            controllers.ApiHelpController.getResource(path = "/pet")
GET     /api-docs.json/cat            controllers.ApiHelpController.getResource(path = "/cat")
```
Thanks to it, adding this route will handle all controllers if the name follows the convention
```
GET     /api-docs.json/*path            controllers.ApiHelpController.getRootResource(path: String)
```


webron:
Thanks for the PR but it used to be for the 1.3.X branch which is no longer maintained. I hope we can provide better support for Play in general in the future.

PR #639, opened 2014-08-06
Adding Ruby Sinatra sample

devonberry:
A sample of swagger-ui hosted in Sinatra.  This is intended to provide
support for API driven design in Ruby environments.


fehguy:
Since this repo is primarily about java integrations, I suggest putting any ruby-specific examples in the swagger-spec readme.

PR #652, opened 2014-08-12
ApiListingResourceJSON supports Accepts: application/json

colin-yell:
Header "Accepts: application/json" as sent from swagger-ui is being rejected. Only "Accepts: application/json; charset=utf-8" accepted.


fehguy:

PR #654, opened 2014-08-16
Sample for exhibiting Issue606

mwhig:
https://github.com/wordnik/swagger-core/issues/606#issuecomment-48190836. 

Look at the classes in packages ending with issue606 under the java-jaxrs sample app. Earlier, adding the hidden=true to the @ApiOperation did nothing, but I can't reproduce it now.


fehguy:

PR #677, opened 2014-09-16
support inheritance api models defined by interfaces

andrew00x:
For example if I have two interfaces:
```java
@ApiModel(value = "Model A", subTypes = {B.class})
public interface A {
    @ApiModelProperty(value = "A")
    String getA();

    void setA(String a);
}
```
and
```java
@ApiModel(value = "Model B", parent = A.class)
public interface B extends A {
    @ApiModelProperty(value = "B")
    String getB();

    void setB(String b);
}
```
In UI for class B I can see just properties (with getters) that are defined directly in interface B:
```
B {
b (string, optional): B,
}
```
Implementations for interfaces generated automatically (for client and server side) and we aren't able to define annotations on implementations. This pull request fixes the issue.


webron:
This should be submitted against the develop-scale_2.10 branch. Can't promise it'll be merged as we not concentrate on critical issues for the 1.3.X releases and concentrate on adding features to the 1.5.X tree.

PR #715, opened 2014-10-07
Change ApiHelpController to trait for both testing and extension

osbornk:
I needed to extend getResource and getResources of ApiHelpController in order to add a custom Actions, but still use the parent methods. I also wanted the ability to do some unit tests with mock controllers. So, I made ApiHelpController into a trait with a Cake parameter. I also added a companion object so nothing existing will break.


webron:
This should be submitted against the develop_scale-2.10 in order to be merged. Can't promise it will be merged as we mostly handle critical issues for 1.3 now.

PR #732, opened 2014-10-29
Removed references to sun implementations

step76:
The sun implementations are not compatible with Google App Engine and
are useless in this context because accessing java.lang.reflect.* is
enough.

The tests pass and on Google App Engine the error "Use of (sometimes) restricted class sun.reflect.generics.reflectiveObjects" disappeared.


fehguy:
This is addressed in 1.5.x, there are no more references to sun impl.

PR #741, opened 2014-11-06
Added Swagger UI to Play2 sample project

rktoomey:
I used your `scala-play2` sample to get Swagger working in my own project.

But when I wanted to look at Swagger UI instead of the raw JSON, it was really difficult to find a working example of how to do it.  So I wanted to contribute back to your project how to make Swagger UI work in Play2.  

This pull request adds a working Swagger UI endpoint, `/api`, and updates the readme with an explanation of how to set it up in your own project.


fehguy:
let's let folks use webjars for play projects.

PR #783, opened 2014-12-08
Add support for scala Vector collection parsing

frosforever:
Support for scala's `Vector` should be native to swagger as it is the suggested collection when an algorithm will not just process `head`.

I'm not sure if this should be it's own supported type in `ModelPropertyParser ` and accompanying `TypeUtil` or if it should included in the check for `TypeUtil.isParameterizedList` and be rendered as a `List`. 

As this is my first time working with the swagger code base, please let me know if I'm attempting to merge to the wrong branch or placing this in the wrong spot.


fehguy:
We have moved the scala support into it's own independently-evolving module.  And there is a test to ensure this is working correctly, please see here:

https://github.com/swagger-api/swagger-scala-module/commit/db60af9f3abc51ed8ed4955c90838012b53b1fcb

Issue #762, opened 2014-11-20
invalid YAML for petstore


fehguy:
For some reason, the security definitions look wrong:

```yaml

securityDefinitions:
  api_key: in: apiKey
    name: "api_key"
    in: "header"
  petstore_auth: in: oauth2
    authorizationUrl: "http://petstore.swagger.wordnik.com/api/oauth/dialog"
    flow: "implicit"
    scopes:
      write:pets: "modify pets in your account"
      read:pets: "read your pets"

```

cc: @earth2marsh

PR #811, opened 2015-01-03
add swagger root and path specific parameters to model, created RefParam...

### LINKED ISSUE #705: [develop 2.0] add properties at swagger object
the swagger-models module doesn't support the `parameters` at the swagger object level:

https://github.com/wordnik/swagger-spec/blob/master/versions/2.0.md#path-item-object-

Helmsdown:
Proposing this as a first draft to fix #705.

* Added a map of Parameter objects to Swagger.java
* Added a List of Parameter objects to Path.java
* Created RefParameter.java to mirror RefProperty.java 
 * RefParameters only have one field of interest (i.e. $ref) which is not exposed in the Parameter interface
 * I am assuming that anyone iterating over a list of Parameter objects, will need to do an instance-of check if they want to handle refs (e.g. in codegen)
* Removed Jackson JsonTypeInfo and JsonSubTypes annotations from Parameter.java
 * I am a Jackson novice, but it appears after some googling that these annotations do not work for the RefParameter edge case
* Created a ParameterDeserializer class to handle the more complex deserialization logic
 * If there is a clever way to do this with Jackson annotations please advise and I will change it
* Added ParameterDeserializer to YAML.java and JSON.java ObjectMapper configuration logic

I am interested in getting Ref's working in swagger-codegen develop_2.0. Please advise.

### DIFF
--- a/modules/swagger-core/src/main/java/com/wordnik/swagger/util/Json.java
+++ b/modules/swagger-core/src/main/java/com/wordnik/swagger/util/Json.java
@@ -1,6 +1,7 @@
 package com.wordnik.swagger.util;
 
 import com.wordnik.swagger.models.Model;
+import com.wordnik.swagger.models.parameters.Parameter;
 import com.wordnik.swagger.models.properties.Property;
 
 import com.fasterxml.jackson.datatype.joda.JodaModule;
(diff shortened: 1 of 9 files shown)


Helmsdown:
this build error is also occurring on the develop_2.0 branch. can you confirm I didn't break it?

fehguy:
thanks!  The deserializer should also address #762 too, testing that.

PR #819, opened 2015-01-14
[develop_2.0] default values (for model properties) implemented

martoe:
according to http://json-schema.org/latest/json-schema-validation.html#anchor101,
https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#schemaObject


fehguy:
this should be supported in 1.5.4-M1-SNAPSHOT.

Issue #820, opened 2015-01-14
[develop_2.0] Scala tests fail on Windows due to EOL issues


martoe:
Some Scala tests are sensitive to the platform-specific line endings:

When cloning the repo on Windows (with `core.autocrlf=false`, which is the default and IMO recommended), the following tests will fail:

	  should read a model with required params and description(CompositionTest): "{[
	  should read a model with composition(CompositionTest): "{[
	  should convert a model(ModelConverterTest): "{[
	  should read an interface(ModelConverterTest): "{[
	  should read an inherited interface(ModelConverterTest): "{[
	  should convert a scala case class with nested models(ScalaModelTest): "{[
	  should read an interface(ScalaModelTest): "{[
  
When manually switching the line endings of those three files **from LF to CRLF**, the tests succeed.

---

One the other hand, some tests suffer from the opposite issue: When cloning with `core.autocrlf=true`, the following tests will fail:

	  should convert serialize a header key model to yaml(AuthSerializationTest): "---[
	  should serialize a with string array PathParameter(ParameterSerializationTest): "---[
	  should serialize a HeaderParameter(ParameterSerializationTest): "---[
	  should serialize a BodyParameter to yaml(ParameterSerializationTest): "---[

When manually switching the line endings of those two files **from CRLF to LF**, the tests succeed...

fehguy:
Fixed by #831

PR #822, opened 2015-01-14
[develop_2.0] Scala tests: Windows EOL issues fixed

martoe:
see https://github.com/swagger-api/swagger-core/issues/820


fehguy:
The replace can be moved to Json.prettyPrint(Object) but doing the replace on each of the string values is really ugly.  Any other way to address it?

martoe:
I agree that this is a very ugly hack - but both replacements are necessary to make the tests succeed on both Unix and Windows independent of the autocrlf style.
Two other possibilities to fix this:

* A change in the Json.prettyPrint() method (as you suggested), combined with the replacement of the test's multiline-strings with plain strings ("foo\nbar") - which of course makes the test less readable
* Change the assertion ("should equal") method to make it ignore whitespaces (can't do that because I know nothing about Scala)

fehguy:
@martoe looks like @whitlockjc has addressed this

PR #656, opened 2014-08-19
Feature/getresource encoding

mauhiz:
Apis description show garbled encoding on platforms where default charset is not UTF-8.

This PR fixes this issue by forcing the serialization charset to UTF-8, and declaring it in the returned content-type; it also fixes a variety of questionable or outdated coding practices in this file.

PR #827, opened 2015-01-21
Feature/getresource encoding

mauhiz:
See #656


webron:
Closing as won't be merged here, but linked to #918 to be taken into account in the new Play module.

PR #831, opened 2015-01-26
Fix tests on Windows

whitlockjc:
Only YAML tests were impacted because only YAML tests (as of now) were
using multi-line strings for comparison.  The approach taken was to
convert the object under test to an ObjectNode, read the multi-line
YAML into an ObjectNode and then use "should equal" for comparison.

This would make #822 unnecessary and could be used in the future to create a test utility for testing real objects instead of string outputs.


fehguy:
We still have a failing test but it's not related to this (and should be moved to v2.1.0-M2)

PR #843, opened 2015-01-31
Add support for @BeanParam to Jersey 2

### LINKED ISSUE #543: @BeanParam annotations should work on methods
the update to jersey2 support does work with @BeanParam objects, however it will only find fields.  It should also treat methods correctly.  Here is the offending code:

https://github.com/wordnik/swagger-core/blob/master/modules/swagger-jaxrs/src/main/scala/com/wordnik/swagger/jaxrs/JaxrsApiReader.scala#L268

whitlockjc:
* @BeanParam is a JAX-RS 2.0 thing and not specific to Jersey 2 but
  until swagger-core has a JAX-RS 2.0 module, it makes sense to be here
  and we can break this logic out when that time comes

Fixes #543


whitlockjc:
I tried to keep with your syntax and style as best as possible.

fehguy:
nice!  thanks

Issue #847, opened 2015-02-02
Model not created for an abstract base model with sub-types


gorgonblot:
The SwaggerSchemaConverter read method returns None as the model if the model does not define any fields even if the model defines sub-types.  For example, assume an API references an abstract base model without any fields.  It references one or more subTypes.

@ApiModel(description = "I am an Abstract Base Model without any declared fields and with Sub-Types",
  subTypes = Array(classOf[Thing3]))
trait AbstractBaseModelWithoutFields {
}

@ApiModel(description = "Thing3", parent = classOf[AbstractBaseModelWithoutFields])
case class Thing3(
                  @(ApiModelProperty @field)(value = "Additional field a") a: String,
                  @(ApiModelProperty @field)(value = "Additional field x") x: Int)
  extends AbstractBaseModelWithoutFields

The desired behavior is for ModelConverters.readAll(classOf[AbstractBaseModelWithoutFields]) to return a list of models including both the base model and the sub-type.

The current behavior is that readAll returns an empty list.

A pull request to address this issue is forthcoming.

cppexpert:
Guys, what's the ETA for the fix of this ?

MichaelDoumin:
Hope to fix today or tomorrow.

fehguy:
This should be addressed in #975

PR #848, opened 2015-02-02
Issue #847 - Support abstract base models without fields

gorgonblot:
Support an abstract base model that doesn't directly define any fields by returning a defined Model object from SwaggerSchemaConverter if either there are properties defined or subTypes defined.


fehguy:

PR #856, opened 2015-02-06
Support mixing Jackson annotation with @ApiModel annotations

tandrup:
In some cases we want to add for example a description to a class with existing @JsonSubTypes and @JsonTypeInfo annotations.

With this fix the @ApiModel annotation will not override discriminator and subtypes if they are not set.


fehguy:
Hi, thanks for sending this.  Two thoughts.

1) the develop branch for the swagger 1.3.x project is develop_scala-2.10.  Master is stable, so PRs should not be sent to it

2) we are on 1.5.0-M1 now, and the development for the 1.3.x branch is mostly for critical bug fixes.  I suggest transitioning to 1.5.0-M1 if you can, since that's where the efforts are being spent.

webron:
Closing this for now as not against the right branch. @tandrup - if you'd like to check if it's relevant for 1.5 and resubmit against that, it would be greatly appreciated.

PR #858, opened 2015-02-06
Fix complication of DefaultParameterExtension

iainh:
DefaultParameterExtension imports both java.lang.reflect.* and com.wordnik.swagger.models.parameters.* making the correct package for class Parameter ambiguous. java.lang.reflect.* is not required so remove it to fix complilation.


webron:
Can you change the PR to be against develop_2.0?

Issue #867, opened 2015-02-12
Java List<T> parameters with generic parameter which is inner class are incorrectly parsed by ModelPropertyParser


cppexpert:
Just found that generic parameters which are inner classes are incorrectly parsed by [ModelPropertyParser](https://github.com/swagger-api/swagger-core/blob/a770366706d0989bfb9e805733a406377c8e49a4/modules/swagger-core/src/main/scala/com/wordnik/swagger/converter/ModelPropertyParser.scala#L208-L227).

Problem lies in following patter matcher

```scala
 val ComplexTypeMatcher = "([a-zA-Z]*)\\[([a-zA-Z\\.\\-0-9_]*)\\].*".r
```

which chokes on `paramType` like `java.util.List<com.crazyrabbit.service.json.timesheet.TimesheetProjectTaskRolesService$RoleRecord>`

as result listing json has following reference `"$ref": "List[RoleRecord]"` for which later it can't find class `RoleRecord` therefore it's not visible in model UI of swagger-ui

todd-richmond:
I've seen the same issue with inner classes - good catch

cppexpert:
Guys (@webron @fehguy), would it possible to merge it and push update to Maven Repo ? Thanks!

cppexpert:
Guys (@webron @fehguy), would it possible to merge it and push update to Maven Repo ? Thanks!

fehguy:
I'm going to review this over the next couple days to make sure we're not regressing anywhere.

webron:
Closing as we're not going to release 1.3.13. Please upgrade to the latest version.

PR #870, opened 2015-02-13
Bugfix for 867 (incorrect parsing of generic class when generic parameter is inner class)

cppexpert:
Guys, I also replaced Class.getSimpleName calls across the module with TypeUtil.getClassSimpleName because it seems to be used as model ID everywhere. It took me a while to understand why inner class is not being referenced in Swagger-UI because of name difference (in my original fix I named model like "OuterClass$InnerClass")

Bugfix for: https://github.com/swagger-api/swagger-core/issues/867


fehguy:
Hi, I'm closing this out as we're not planning on more work on the 1.3.x branch.

PR #893, opened 2015-03-04
Fix for JavaDoc compilation failure with JDK8 (Issue 886)

cppexpert:
(cherry picked from commit 4bee4c0)


fehguy:
Looks like this causes an issue with non jdk8 builds.  Is there a way to make this flag conditional depending on jdk version?

cppexpert:
Not sure, I've never used Maven. Another option is to fix javadoc comments.

fehguy:
this has been fixed.

Issue #906, opened 2015-03-10
1.5.x: Class from HTTP body parameter is not added to list of models


cppexpert:
Class from HTTP body parameter is not added to list of models. Here is how endpoint is defined

```java
@POST
@Consumes(MediaType.APPLICATION_JSON)
@Produces(MediaType.APPLICATION_JSON)
@Path("testPostWithBody")
@ApiOperation(response = String.class, value = "Returns string with provided parameter")
public String testPostWithBody(@ApiParam(name = "body", value = "input parameters in json form")
                                   List<ClassWithString> items) {}
```

User cannot see `ClassWithString` when he click on `models` link in Swagger-UI because it's missing from list of model classes of resulting json.

Here is generated Swagger json

```json
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.0",
    "title": ""
  },
  "tags": [
    {
      "name": "authenticated",
      "description": "Authenticated routes"
    },
    {
      "name": "unauthenticated",
      "description": "Unauthenticated endpoints"
    }
  ],
  "schemes": [
    "http"
  ],
  "paths": {
    "/authenticated/testParams": {
      "get": {
        "tags": [
          "authenticated"
        ],
        "summary": "Returns string with provided parameter",
        "description": "",
        "operationId": "testParams",
        "consumes": [
          "text/plain"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "id",
            "in": "query",
            "description": "ID of timesheet",
            "required": false,
            "type": "integer",
            "format": "int32"
          }
        ],
        "responses": {
          "200": {
            "description": "successful operation",
            "schema": {
              "type": "string"
            }
          }
        }
      }
    },
    "/authenticated/testPostWithBody": {
      "post": {
        "tags": [
          "authenticated"
        ],
        "summary": "Returns string with provided parameter",
        "description": "",
        "operationId": "testPostWithBody",
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "in": "body",
            "name": "body",
            "description": "input parameters in json form",
            "required": false,
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/ClassWithString"
              }
            }
          }
        ],
        "responses": {
          "200": {
            "description": "successful operation",
            "schema": {
              "type": "string"
            }
          }
        }
      }
    },
    "/unauthenticated/getApiToken": {
      "get": {
[...]

PR #920, opened 2015-03-18
fixed adding of definitions for complex body input parameters 

olensmar:
to generated swagger definition - see #906

hope this was the right approach - let me know otherwise :-)

### DIFF
--- a/modules/swagger-jersey2-jaxrs/src/test/scala/ScannerTest.scala
+++ b/modules/swagger-jersey2-jaxrs/src/test/scala/ScannerTest.scala
@@ -30,4 +30,16 @@ class ScannerTest extends FlatSpec with Matchers {
     limit.getName() should be ("limit")
     limit.getDescription() should be ("maximum number of records to return")
   }
+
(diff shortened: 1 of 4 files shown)


modules/swagger-jersey2-jaxrs/src/test/scala/ScannerTest.scala
```diff
@@ -30,4 +30,16 @@ class ScannerTest extends FlatSpec with Matchers {
     limit.getName() should be ("limit")
     limit.getDescription() should be ("maximum number of records to return")
   }
+
+  it should "scan another resource" in {
+    val swagger = new Reader( new Swagger()).read( classOf[ResourceWithComplexBodyInputType])
+
+    Json.prettyPrint( swagger )
```
fehguy: You might dispose of the prettyPrint in the tests unless they're needed

fehguy:
Looks great--only comment is about removing the prettyPrint during tests to keep them less verbose

PR #921, opened 2015-03-19
maven-enforcer-plugin with dependencyConvergence

laat:
This pull-request forces every dependency, even the transitive dependencies, to converge to the same version. I have excluded the offending oldest dependencies.


fehguy:
Hi, I know this is confusing--and it's going to get a lot less confusing soon, but the develop branch is develop_2.0 for the current (1.5.x) version.  If you're looking for something for the previous (1.3.x), it should be develop_scala-2.10.  Sorry it's a maze, but again, will be cleaned up completely when 1.5.0 is released.

laat:
ohh, I see. And it is already enforced in develop_2.0.

Thank you =)

PR #923, opened 2015-03-20
Get build to run on windows from new checkout

jsimmonsc:
I was trying to do some research on the new java models on the develop_2.0 branch and tried to checkout and build and was unable to do so.

Swagger-core  was unable to build because it depended on swagger-parser, and swagger-parser would't build because it depended on swagger-core.  I know that swagger-core depended on a slightly older snapshot version, but, looking through the test it wasn't that hard to remove the dependency on swagger-parser.

I also upgraded scala test on swagger core to stop the maven warning about multiple scala versions.

### DIFF
--- a/modules/swagger-core/pom.xml
+++ b/modules/swagger-core/pom.xml
@@ -91,12 +91,6 @@
     </dependency>
 
     <dependency>
-      <groupId>io.swagger</groupId>
-      <artifactId>swagger-parser</artifactId>
-      <version>${swagger-parser-version}</version>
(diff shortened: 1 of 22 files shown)


jsimmonsc:
Also, this is a bit tangential but... I haven't done a lot of scala dev.  What's the best environment to verify that it works?  The travis build above failed, all the tests pass on redhat, and I get 24 failures on windows.

jsimmonsc:
Closing since a different commit was made to fix this

PR #933, opened 2015-03-26
changing apiteam@wordnik.com to apiteam@swagger.io in 10 files

emckean:
Hi! just changing the email addresses from apiteam@wordnik.com to apiteam@swagger.io


fehguy:
:+1:

Issue #937, opened 2015-03-28
[jersey2] QueryParam marked as required not working


Flamenco:
PathParam parameter works:
@ApiParam(required = true) @PathParam("userId") String username

![screen shot 2015-03-28 at 7 15 31 am](https://cloud.githubusercontent.com/assets/819940/6880864/61961ec8-d51a-11e4-94a1-ecfa5ce87913.png)

QueryParam does not work:
@ApiParam(required = true) @QueryParam("username") String username
![screen shot 2015-03-28 at 7 15 56 am](https://cloud.githubusercontent.com/assets/819940/6880865/67e9da9e-d51a-11e4-8d55-0ff8f7ec04ee.png)

webron:
Using which version of swagger-core?

Flamenco:
1.5.2-M1

fehguy:
I believe this is fixed in 1.5.3-M1. Can you please retest?

Flamenco:
with 1.5.3-M1 same issue.  Also note that default values are not working.  

	@GET
	@Path("/goodbye/{name}")
	@ApiOperation(value = "goodbye")
	public String goodbye(@ApiParam(defaultValue = "the-name", required = true) @PathParam("name") String name) {
		return "<Goodbye> Goodbye " + name + " </Goodbye>";
	}

	@GET
	@Path("/goodbye2")
	@ApiOperation(value = "goodbye2")
	public String goodbye2(@ApiParam(defaultValue = "the-name", required = true) @QueryParam("name") String name) {
		return "<Goodbye> Goodbye " + name + " </Goodbye>";
	}
![screen shot 2015-03-28 at 2 43 10 pm](https://cloud.githubusercontent.com/assets/819940/6882375/26232b80-d559-11e4-9359-61664766eebf.png)

fehguy:
Thank you for testing, will assure this is fixed

Flamenco:
Please also notice that the default values don't appear either :(  

The @Default jax-rs annotation does work however, but the docs state that default will be used if that value is not specified.

I would fix this myself but my Scala skills need some work.  For me it's easy and fast to write, but hard to read.

shivangshah:
+1 on "required" not working other than @PathParam

PR #940, opened 2015-03-29
Removing System.out.println from a previous bugfix under #937

shivangshah:
@fehguy Please have a look when you get a chance. It's an extremely minor fix but the server logs were really getting overloaded by the following message after your bug fix for #937 :
INFO  [2015-03-29 08:11:50,664] org.reflections.Reflections: Reflections took 81 ms to scan 1 urls, producing 22 keys and 41 values 
true
true
true
true
true
true
true
true
true
true
true
true
true
true
true
true
true
true
true
true
true


fehguy:
thanks

PR #945, opened 2015-03-30
Fixed tag descriptions in samples

webron:


fehguy:
I believe the swagger-ui e2e tests are going to look for the text that used to be there (irrespective of order).  Can you please keep the same text values for now?

webron:
Then just don't merge or close it for now. The problem is that the descriptions are wrong and they look funny in the online petstore sample.

I can look into the swagger-ui e2e tests too, though it may be too late with the weekend release.

fehguy:
Just change the descriptions to what they should be, swapping the tag name so they're correct.  Push that and the PR should update.

PR #973, opened 2015-04-09
Mule

Amyntas:
Mule module and updated sample


Amyntas:
Tony, if you look at https://github.com/swagger-api/swagger-core/pull/973/files#diff-6b04f0da967ddc159fc20e0c415942bfR47 and let me know if you'd like me to cut one of the config types, or add some code to prefer one over the other but try both, ill add that to the PR

PR #975, opened 2015-04-10
Fix for #847. Implemented findSubtypes.

MichaelDoumin:
Since the method findSubtypes was not implemented after jackson-swagger-module had been moved to the  swagger-core, ModelConverter could not find subtypes of passed class. Fixed it. Added unit-test.

### DIFF
--- a/modules/swagger-core/src/main/java/com/wordnik/swagger/jackson/AbstractModelConverter.java
+++ b/modules/swagger-core/src/main/java/com/wordnik/swagger/jackson/AbstractModelConverter.java
@@ -1,5 +1,7 @@
 package com.wordnik.swagger.jackson;
 
+import com.fasterxml.jackson.databind.introspect.AnnotationIntrospectorPair;
+import com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector;
 import com.wordnik.swagger.annotations.ApiModelProperty;
 import com.wordnik.swagger.converter.ModelConverter;
 import com.wordnik.swagger.converter.ModelConverterContext;
@@ -35,7 +37,10 @@ public abstract class AbstractModelConverter implements ModelConverter {
             new SimpleModule("swagger", Version.unknownVersion()) {
               @Override
               public void setupModule(SetupContext context) {
-                context.insertAnnotationIntrospector(new SwaggerAnnotationIntrospector());
+                context.insertAnnotationIntrospector(
+                        new AnnotationIntrospectorPair(
+                                new SwaggerAnnotationIntrospector(),
+                                new JacksonAnnotationIntrospector()));
               }
             });
           _mapper = mapper;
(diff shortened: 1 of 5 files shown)


webron:
Related to #847.

tomtit:
Looks like this snippet:
```
+    final JsonSubTypes jst = a.getAnnotation(JsonSubTypes.class);
+    if (jst != null) {
+      final JsonSubTypes.Type[] subTypes = jst.value();
+      final List<NamedType> names = new ArrayList<NamedType>(subTypes.length);
+      for (JsonSubTypes.Type subType : subTypes) {
+        names.add(new NamedType(subType.value()));
+      }
+      if (!names.isEmpty()) {
+        return names;
+      }
+ }
```
duplicates FasterXML internals.
Deep inside we have instance of the `com.fasterxml.jackson.databind.introspect.AnnotationIntrospectorPair` class:
```
    @Override
    public List<NamedType> findSubtypes(Annotated a)
    {
        List<NamedType> types1 = _primary.findSubtypes( a );
        List<NamedType> types2 = _secondary.findSubtypes( a );
        if (types1 == null || types1.isEmpty()) return types2;
        if (types2 == null || types2.isEmpty()) return types1;
        ArrayList<NamedType> result = new ArrayList<NamedType>(types1.size() + types2.size());
        result.addAll(types1);
        result.addAll(types2);
        return result;
    }
```
where `_primary` holds our SwaggerAnnotationIntrospector and `_secondary` holds some internal class that handles JSON annotations.
At least as we can see from our [CompositionTest.scala](https://github.com/swagger-api/swagger-core/blob/develop_2.0/modules/swagger-core/src/test/scala/CompositionTest.scala) JSON annotations are already supported.

MichaelDoumin:
The last fix considers Alexey's remark above: JsonSubTypes processing is re-used from JacksonAnnotationIntrospector.

modules/swagger-core/src/main/java/com/wordnik/swagger/jackson/AbstractModelConverter.java
```diff
@@ -35,7 +37,10 @@ protected AbstractModelConverter(ObjectMapper mapper) {
             new SimpleModule("swagger", Version.unknownVersion()) {
               @Override
               public void setupModule(SetupContext context) {
```
tomtit: I think this change is not required at all.

modules/swagger-core/src/main/java/com/wordnik/swagger/jackson/AbstractModelConverter.java
```diff
@@ -35,7 +37,10 @@ protected AbstractModelConverter(ObjectMapper mapper) {
...
-                context.insertAnnotationIntrospector(new SwaggerAnnotationIntrospector());
+                context.insertAnnotationIntrospector(
+                        new AnnotationIntrospectorPair(
+                                new SwaggerAnnotationIntrospector(),
+                                new JacksonAnnotationIntrospector()));
```
tomtit: I think this change is not required at all.

fehguy: I agree, I believe that the JacksonAnnotatinIntrospector is embedded in SwaggerAnnotationIntrospector

PR #980, opened 2015-04-13
Configure Jackson to only scan fields and setters of @BeanParam classes

griswold:
Thank you so much for all the amazing work on swagger-core!

When using swagger-core on a recent project, I came across an NPE where ParameterUtils.isMethodArgumentAnArray was receiving a null paramClass argument. It turns out that this was because a BeanParam class declared a method with a name like getX, where X did not directly correspond to a field name. Because the method name began with "get," Jackson was returning a BeanPropertyDefinition corresponding to it whose field and setter were null.

After looking at the JAX-RS documentation for BeanParam, I wasn't able to think of a case where introspecting on the getter methods of a BeanParam would be useful, so I configured the ObjectMapper to only consider fields and setters. These seem to be the only places where JAX-RS annotations are used on BeanParam classes.

To demonstrate the issue, I modified one of the BeanParamClasses to include a method beginning with "get" that does not directly correspond to a field.

### DIFF
--- a/modules/swagger-jersey2-jaxrs/src/main/java/com/wordnik/swagger/jersey/SwaggerJersey2Jaxrs.java
+++ b/modules/swagger-jersey2-jaxrs/src/main/java/com/wordnik/swagger/jersey/SwaggerJersey2Jaxrs.java
@@ -8,6 +8,8 @@ import com.wordnik.swagger.jaxrs.utils.ParameterUtils;
 import com.wordnik.swagger.models.parameters.Parameter;
 import com.wordnik.swagger.util.Json;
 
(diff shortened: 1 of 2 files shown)


griswold:
Sorry, need to rethink the approach here. I didn't notice that the mapper is a singleton, so limiting the visibility of that mapper affects other things that use the singleton. I considered creating the mapper with Json.create(), but this also affects the behavior of Json.mapper because it assigns the static mapper in addition to returning a new instance. Back to the drawing board.

webron:
@griswold - keep in mind there are two ObjectMappers - one in `Json` and one in `Yaml` and they both need to be taken into account.

PR #984, opened 2015-04-15
use better pattern to compare to empty resourcePackages

### LINKED ISSUE #983: Instable String comparison in BeanConfig
In Version 1.5.1-M2 we find the following code around lines 159ff:

    public Set<Class<?>> classes() {
        ConfigurationBuilder config = new ConfigurationBuilder();
        Set<String> acceptablePackages = new HashSet<String>();

        if(resourcePackage != "") {
            String[] parts = resourcePackage.split(",");
            // ... rest omitted

It would be better to write `if(!"".equals(resourcePackage))` instead, and not compare by String identity.
And best check for `null` as well -> `if(resourcePacke != null && !"".equals(resourcePackage))`

ctolkmit:
fixes #983 

this should fix my reported issue (and is my first ever pull request on github, yeah :))


webron:
Technically, if you compare a literal string, there's no need to check if the compared value is null first as it won't throw NPE (just return false). But as far as a PR goes, it looks good. Congrats on the first one :)

ctolkmit:
You are right, but the following line will throw an NPE if the string is NULL (inside the if-braces). the comparison itself is null-safe

webron:
Right you are!

Issue #802, opened 2014-12-22
[develop 2.0] ensure composition support is correct


fehguy:
The support for `allOf` in model definitions is partially complete--need to ensure that the support follows the design of the 2.0 spec.

See here:

https://github.com/swagger-api/swagger-core/blob/develop_2.0/modules/swagger-core/src/test/scala/CompositionTest.scala

tomtit:
The CompositionTest unit test is enabled now and checks the presence of the 'allOf' entry in the model definition. What else should be done to resolve the issue?

tomtit:
The first test for the Human class seems to be incorrect as it accepts Pet as a descendant of Human:
```
"Pet" : {
    "allOf" : [ {
        "$ref" : "#/definitions/Human"
    }
```
So, here is an improvement #987 for handling subtypes stricter.

PR #987, opened 2015-04-15
fix for 802: Subtypes hadnling has been improved: classes inheritance has been taken into consideration.

tomtit:


tomtit:
Fixes #802

PR #994, opened 2015-04-20
Fixes #394. Support of class-based sub-resources has been added.

### LINKED ISSUE #394: Problem with subresource detection when using dependency injection in Jersey 2
This works fine:
```
@Path("/{objectId: [a-fA-F0-9]{32}}")
@ApiOperation(value = "subresource", response = SomeSubResource.class)
public SomeSubResource continueInSubResource() {
  return new SomeSubResource();
}
```

The problem occurs when using Jersey dependency injection for creating the resources. The code is then changed to this:
```
@Path("/{objectId: [a-fA-F0-9]{32}}")
@ApiOperation(value = "subresource", response = SomeSubResource.class)
public Class<SomeSubResource> continueInSubResource() {
  return SomeSubResource.class;
}
```

That is, we only tell Jersey which type of class we want, and Jersey dependency injection instantiates the correct class (passing in the dependencies to the constructor automatically). However, this seems to confuse swagger, which does not detect this subresource at all.

tomtit:
Support of class-based sub-resources has been added.

### DIFF
--- a/modules/swagger-jaxrs/pom.xml
+++ b/modules/swagger-jaxrs/pom.xml
@@ -7,10 +7,8 @@
     <relativePath>../..</relativePath>
   </parent>
   <modelVersion>4.0.0</modelVersion>
-  <groupId>com.wordnik</groupId>
   <artifactId>swagger-jaxrs</artifactId>
   <packaging>jar</packaging>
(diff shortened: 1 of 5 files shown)


tomtit:
Fixes #394.

PR #995, opened 2015-04-20
Processing of @Api and @ApiOperation's consumes() and produces() es has ...

### LINKED ISSUE #952: JaxRS Reader - @Api and @ApiOperation's consumes/produces are ignored
`@Api` and `@ApiOperation` have `consumes()` and `produces()` values that can be used to override the values of `@Consumes` and `@Produces` respectively. 

Currently these values are read by the Reader, but not used to override the annotations value and do not propagate to the operation definitions.

lugaru1234:
...been added to Reader. Fixes #952


fehguy:
There was a build error but I've addressed it.

PR #999, opened 2015-04-21
Fixed #957: Add missing 'range' support for @ApiParam's allowableValues

### LINKED ISSUE #957: Add missing 'range' support for @ApiParam's allowableValues

iushankin:
Fixed #957: Add missing 'range' support for @ApiParam's allowableValues

Also was changed:
1. Removed unused variables
2. Moved common functionality into a base class (AbstractSerializableParameter)
3. Implemented support of exclusiveMinumum/exclusiveMaximum in ranges


iushankin:
Also fixed #1004 issue

fehguy:
This is a monster commit so it may take some time to get reviewed.

modules/swagger-core/src/test/scala/BeanValidatorTest.scala
```diff
@@ -29,9 +29,9 @@ class BeanValidatorTest extends FlatSpec with Matchers {
     password.getMaxLength() should be (20)
 
     val minBalance = properties.get("minBalance").asInstanceOf[DoubleProperty]
-    minBalance.getExclusiveMinimum should be(0.1)
+    minBalance.getExclusiveMinimum().booleanValue() should be (true)
```
fehguy: Why did you change the check for the value (0.1) to convert it to a boolean here?

modules/swagger-models/src/main/java/com/wordnik/swagger/models/parameters/AbstractSerializableParameter.java
```diff
@@ -0,0 +1,153 @@
...
+
+  @JsonProperty("default")
+  protected String defaultValue;
+
+  public T property(Property property) {
+    this.setProperty(property);
+    return (T) this;
+  }
+
+  public T parameterType(String type) {
```
fehguy: This will break the signature for the model.  While it may be more clear, I think we are too far along in the release of 1.5 to do this.

fehguy:
Overall this is a good commit.  A couple of comments.

1) Please try to break up further PRs into smaller ones.  This one is very large and contains both functionality changes, refactoring, and style changes.  Best to do more, smaller PRs or reviewing them is onerous (and takes too long)

2) If you see something broken that isn't following the JSON schema spec, for example, please call it out explicitly.  The `exclusiveMinimum` is an example.  It would be much easier to reference your comments than go hunting around in the spec to see why you changed the expected values from the tests.

3) There are model signature changes that shouldn't be there unless absolutely necessary.  Please go through the ParameterType models and keep the signatures consistent unless there is no option.

If you can address issue (3) or make it clear why it has to change, I'll merge this.

Thank you for the contribution!

modules/swagger-models/src/main/java/com/wordnik/swagger/models/parameters/AbstractSerializableParameter.java
tomtit: @fehguy could you please explain this? `parameterType` is just a helper to call `setType(String)`, while the `type` is reserved word in scala.

PR #1010, opened 2015-04-27
Collection<Enum<?>> as a field on a transfer object yields spec with ClassLoader et al.

kungfoo:
Specs that are returned when accessing swagger.json are not valid as per https://github.com/swagger-api/swagger-js/issues/268

### DIFF
--- a/modules/swagger-models/src/main/java/com/wordnik/swagger/models/Path.java
+++ b/modules/swagger-models/src/main/java/com/wordnik/swagger/models/Path.java
@@ -3,6 +3,8 @@ package com.wordnik.swagger.models;
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.wordnik.swagger.models.parameters.Parameter;
 
@@ -12,6 +14,7 @@ import java.util.List;
 import java.util.Map;
 
 @JsonPropertyOrder({ "get", "post", "put", "delete", "options", "patch"})
+@JsonInclude(Include.NON_NULL)
 public class Path {
   private Operation get;
   private Operation put;


fehguy:
I don't see why this is necessary.  What provider are you using for serializing the `Swagger` object to JSON?  We have built one that will do this already, and if you're not using it, you should use the logic that it has.

The raw `ObjectMapper`:

https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/com/wordnik/swagger/util/Json.java

The provider:
https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/json/JacksonJsonProvider.java

kungfoo:
I am using ApiListingResource in a JAX-RS application.

fehguy:
@kungfoo great--but you also need to use the provider I mentioned.  Can you please do that?  It won't affect any of your other operations, it's only used for the swagger mapping.

kungfoo:
Do I configure the jackson provider in web.xml?

fehguy:
Many ways to configure it, but the easiest--assuming you're using web.xml--is to add the path to your scanning properties:

https://github.com/swagger-api/swagger-core/blob/master/samples/java-jaxrs/src/main/webapp/WEB-INF/web.xml#L12

kungfoo:
Does this also work with resteasy, since we're actually using resteasy
3.0.9 to get jax-rs 2.0 features...

webron:
@kungfoo - if you're using jax-rs 2.0, you'd need the swagger-jersey2-jaxrs dependency. Yes, it would work for resteasy as well.

kungfoo:
I have added that dependecy as it is done in the sample project.
I'll check there for the appropriate web.xml. Must've missed that.

webron:
Also, FWIW, https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/listing/SwaggerSerializers.java will give you a more complete solution than https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/json/JacksonJsonProvider.java so you can use that s a Provider instead (it also supports YAML).

fehguy:
Yes, @webron is right, I should have pointed you to that file.

@webron should we delete the old provider?

webron:
@fehguy - I think so. We have it for easier migration, but I'm planning on making it clear in the migration guide that the ApiListingProvider and SwaggerSerializers providers are the only ones needed, making it much more easier to configure without caring which jax-rs module is used.

kungfoo:
Oayk, so adding `SwaggerSerializers` as a provider to web.xml and adding `ApiListingResource` to the class that extends `javax.ws.rs.core.Application` actually works wonders.
I will update the pull request with an adapted sample if it is missing there.

kungfoo:
Hmm, still not quite working: Swagger also serializes java properties that will never be in the JSON such as `Package` and `ClassLoader`.

```json
 "Package" : {
      "properties" : {
        "specificationTitle" : {
          "type" : "string"
        },
        "specificationVersion" : {
          "type" : "string"
        },
        "specificationVendor" : {
          "type" : "string"
        },
        "implementationTitle" : {
          "type" : "string"
        },
        "implementationVersion" : {
          "type" : "string"
        },
        "implementationVendor" : {
          "type" : "string"
        },
        "name" : {
          "type" : "string"
        },
        "annotations" : {
          "type" : "array",
          "items" : {
            "$ref" : "#/definitions/Annotation"
          }
        },
        "declaredAnnotations" : {
          "type" : "array",
          "items" : {
            "$ref" : "#/definitions/Annotation"
          }
        },
        "sealed" : {
          "type" : "boolean"
        }
      }
    },
```

How does one prevent this?

webron:
Need to see the code that serializes into that.

kungfoo:
It actually is an enum like this:

```java
public enum Action {
  SNIP,
  SNAP,
  SNOOP,
  APPROVE;
}
```
and that enum is used in a the return type of a method that looks like this:

```java
public class Result {
 private List<ResultItem> items;
 private List<Notification> notifications;
 // getters and setters for bean properties
 // some more properties
}

public class ResultListItem {
 private List<Action> actions;

 // getters and setters for bean property
}
```

webron:
Any chance you can share your full project or a sample project that reproduces that behavior? I can't see how the above snippet would cause what you're experiencing.

kungfoo:
@webron I will try to extract the code that produces this while still protecting the guilty. :)
I hope I can reuse one of the poms of the samples.

webron:
No worries, thanks. You can share it with us in private if it helps.

kungfoo:
You think it matters that we're using JBoss AS 7.3 and resteasy 3.0.9 instead of Jetty and Jersey?

webron:
It shouldn't matter, but it's better to have the resteasy dependency in the sample to be on the safe side.

kungfoo:
Okay, I will create a sample that uses maven for the build (we're using a
specialized/bastadrized version of ant/ivy (don't ask)) and create a
repository for it. Stand by.

webron:
Thank you, much appreciated.

kungfoo:
This is taking me longer than expected because I have to monkeypatch resteasy to 3.0.9 and I'm not really a maven maven (no pun inteded). Tomorrow, the sample will be ready.

kungfoo:
I've managed to reproduce the problem:

https://github.com/kungfoo/swagger-resteasy-sample

Since I've got to monkey patch resteasy, I added a rakefile to do that, since honestly that kinda thing in maven. :)

Anyhow, if you navigate to the documentation and look at the Model, or also the generated swagger.json, you can see the `declaringClass` and son on definitions.

webron:
Thanks. I'll probably get around to look at it tomorrow.

kungfoo:
You can look at the generated json here once it runs:

    http://localhost:8080/resteasy-sample/rest/swagger.json

kungfoo:
It seems that the field with the Set<Action> trips swagger. It really doesn't matter that it's a Set, the same also happens, when the field is declared as a List.

I will add a testcase to my fork that proves this later on.

kungfoo:
Just tested my sample on OSX as opposed to linux at work and it is not running on JDK 8, since 7.1.1 is not supporting that.
Are you guys running JDK 8? I will have to add a branch that uses Wildfly 8 to run the sample...

kungfoo:
I created a branch `java-8-wildfly` that is using wildfly to deploy the app.
It reproduces the bug without patching resteasy, because widlfy includes a 3.x release of resteasy.

webron:
@kungfoo - swagger-core itself uses Java 7. We use whatever is needed.

I'm a bit confused by your finding. Can you share the specific line of code you think is causing the problem?

kungfoo:
It is both reproducable on both branches, but JBoss AS 7.1.1 is not compatible with Java 8.

- JBoss AS 7.1.1, resteasy 3.0.9 (patched) and Java 7
- Wildfly 8.0.0, resteasy 3.x (delivered with Wildfly) and Java 8

The line that causes the issue is this one:

https://github.com/kungfoo/swagger-resteasy-sample/blob/master/resteasy-sample/src/main/java/ch/ergon/sample/rest/model/Pet.java#L28

If the transfer object includes a `Collection` of `Enum` values, the problem occurs.

kungfoo:
I would love to create a test-case in the code for this, but I can't seem to get the build to pass on my machine here. On my laptop it worked this morning, very weird.
Running

    mvn verify

should actually work, right?

webron:
should work to do what? ;)

webron:
Not really sure what you're trying to build though.

kungfoo:
I'm actually trying to build swagger-core so I can add a testcase with a Set<Enum<?>>.

webron:
Which branch?

kungfoo:
Tried the master branch.

webron:
Just built it locally, went fine. Try running `mvn clean install`.

kungfoo:
Freshly cloned: no luck, it still fails.

```
mvn -version
Apache Maven 3.3.3 (NON-CANONICAL_2015-04-29T11:04:34_root; 2015-04-29T10:04:34+02:00)
Maven home: /opt/maven
Java version: 1.8.0_45, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-jdk/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "3.19.3-3-arch", arch: "amd64", family: "unix"
```

webron:
Not sure what to tell you.

```
[INFO] Reactor Summary:
[INFO]
[INFO] wordnik-swagger-project ............................ SUCCESS [  2.010 s]
[INFO] swagger-annotations ................................ SUCCESS [  9.194 s]
[INFO] swagger-models ..................................... SUCCESS [  5.911 s]
[INFO] swagger-core ....................................... SUCCESS [ 27.123 s]
[INFO] swagger-jaxrs ...................................... SUCCESS [ 17.405 s]
[INFO] swagger-jersey2-jaxrs (Jersey 2.x support) ......... SUCCESS [ 12.728 s]
[INFO] swagger-jersey-jaxrs ............................... SUCCESS [  4.495 s]
[INFO] swagger-hiberate-validations ....................... SUCCESS [ 11.868 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 01:31 min
[INFO] Finished at: 2015-04-30T17:21:24+03:00
[INFO] Final Memory: 57M/559M
[INFO] ------------------------------------------------------------------------
```

```
mvn -version
Apache Maven 3.3.3 (7994120775791599e205a5524ec3e0dfe41d4a06; 2015-04-22T14:57:37+03:00)
Maven home: /usr/local/Cellar/maven/3.3.3/libexec
Java version: 1.8.0_45, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_45.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.10.3", arch: "x86_64", family: "mac"
```

kungfoo:
I will try to run it again at home. It worked on the laptop...

fehguy:
okie.  This PR has drifted into different issues during it's short life, but it did expose one very important bug--arrays of enums (your Set<Action> property) was causing expansion problems.  Thank you for bringing that up, and https://github.com/swagger-api/swagger-core/commit/605180e75a0246adec0f97aeef0fc380f092a5c3 should address it.

This does raise, unfortunately, another issue.  The latest swagger-jersey2-jaxrs module has added an import for `org.glassfish.jersey.media.multipart.FormDataParam`, which is specific to the jersey implementation.  Thus it cannot be excluded from your pom.xml and you get an ugly runtime error.

That said, I'm going to close this ticket and open a new one to track the jersey2 dependency

kungfoo:
Sorry that this has drifted so far, but I still think it's great we found and fixed that bug with the enums. :)

PR #1014, opened 2015-04-28
Fixed #953: protocols() value in @Api and @ApiOperation should change and be parsed

### LINKED ISSUE #953: protocols() value in @Api and @ApiOperation should change and be parsed
Right now, `protocols()` is a string in both `@Api` and `@ApiOperation`. We should consider the following changes:

- Change the type from string to string[], to avoid having to split values (easier input). This is a breaking change since it changes the annotation, but in 1.3.X this value was never used. If we do this change, it should be done in both annotations. Another possible change - instead of array of strings, array of Scheme Enum with fixed values (HTTP, HTTPS, WS, WSS) as given by the spec. We may not want to restrict it though.
- JaxRS Reader reads the value from `@ApiOperation` but does not process the value from `@Api`. It should read the value from `@Api` and assign the value to all operations, unless explicitly overridden by `@ApiOperation`.

iushankin:
Fixed #953: protocols() value in @Api and @ApiOperation should change and be parsed

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
@@ -6,6 +6,7 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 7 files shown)


webron:
Generally looks good. A few comments/questions:
- Please remove the javadoc changes for the annotations. I have a pending PR with a whole rework of the javadocs as a whole, including that.
- What happens if a user chooses to add an invalid scheme (say, `ftp`). Dry running it makes it seem like it would add `null` to the final schemes list. Possibly should add a test for that as well.
- For completeness of testing, there should be a test with no protocols at either level, with an output of no schemes property at all (as opposed to a schemes property with an empty array for example).
- The change to the schemes in the `Swagger.java` class is a bit out of scope for this PR as it has nothing to do with the `@Api` or `@ApiOperation` protocols. It's ok to keep it though.
- Can't think of a reason to use LinkedHashSet over HashSet in this use case. Not really important, just a general thing.

PR #1051, opened 2015-05-08
Fixed #1027. ApiModelProperty#access support has been added

### LINKED ISSUE #1027: "access" param handling for model properties
using swagger-core-1.5.1-M2.

I am trying to filter parameters/properties based on the access param.

`@ApiParam`/`@ApiImplicitParam` and swagger-models `Parameter` work just fine. There, both sides support the property and it can be mapped (e.g. in swagger-jaxrs `ParameterProcessor`).

However, `@ApiModelProperty`s access flag/param is not mapped to swagger-models `Property`. `AbstractProperty`/`Property` are lacking a corresponding access property, I think?

Not having the access param available in swagger-models `Property` makes it hard, if not impossible, to filter model properties based on the access param with a `SwaggerSpecFilter`. At least I did not find a clean way.

related: https://github.com/swagger-api/swagger-core/issues/996

thanks, zyro

lugaru1234:
Fixed #1027. ApiModelProperty#access support has been added

### DIFF
--- a/modules/swagger-core/src/main/java/com/wordnik/swagger/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/com/wordnik/swagger/jackson/ModelResolver.java
@@ -394,6 +394,10 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
         if(property != null) {
           property.setName(propName);
 
+          if (mp != null && !mp.access().isEmpty()) {
+            property.setAccess(mp.access());
+          }
+
           Boolean required = md.getRequired();
           if(required != null)
             property.setRequired(required);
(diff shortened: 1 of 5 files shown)


lukecampbell999:
Any chance that this could be backward patched into 1.5 stream also - it's a big nuisance for us in terms of the spec showing fields that we've already marked as non-visible.

lukecampbell999:
ie we have a class with a property that looks like this :-

@JsonView(Views.Internal.class)
@ApiModelProperty(required = false, readOnly = true, access = "private")
private String apiKey;

at the moment, when we use a SwaggerSpecFilter, access always comes back as null, therefore although it's marked as ignored in the view that we generate in the output when the call is made, in the spec itself, the field shows up. 

The best we can do for right now is to use readOnly or a combination of required and readOnly, which isn't ideal.

webron:
If it's not working for you, it's a bug. Just file a new ticket.

lukecampbell999:
Added: https://github.com/swagger-api/swagger-core/issues/2067

PR #1065, opened 2015-05-13
Fixes #943: Handling of method parameters without @ApiParam has been added.

### LINKED ISSUE #943: Request Body not picked up by Scanners
The previous scanners were able to pickup Body parameters without any @ApiParam now I had to do something like this for it to pickup body parameters:
BEFORE:
```
public Response create(@ApiParam(value = "someHeader", required = true)
            @HeaderParam("someHeader") String header, String jsonBody) {
```
AFTER
```
public Response create(@ApiParam(value = DATASOURCE_HEADER, required = true)
                        @HeaderParam("someHeader") String header,
            @ApiParam(value = "jsonBody", required = true) String jsonBody) {
```
Discussion with @fehguy can be found here #925

tomtit:
Fixes #943: Handling of method parameters without @ApiParam annotations has been added.

### DIFF
--- a/modules/swagger-core/src/main/java/com/wordnik/swagger/converter/ModelConverters.java
+++ b/modules/swagger-core/src/main/java/com/wordnik/swagger/converter/ModelConverters.java
@@ -100,21 +100,23 @@ public class ModelConverters {
   }
 
   private boolean shouldProcess(Type type) {
-    boolean process = true;
     if(type instanceof Class<?>){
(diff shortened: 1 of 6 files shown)


tomtit:
Ignoring of whitespaces will make reviewing easier https://github.com/tomtit/swagger-core/commit/1861659a917c7a64309f5ff7e06f997ceeecd64f?w=1

Issue #1072, opened 2015-05-15
Inherited annotations not observed


okigan:
Same as issue  #400 (see for more context), but occurring in latest version 1.5m2: 

I'm using com.wordnik:swagger-jaxrs_2.10:1.3.1 and I'm observing that @Api... annotations in parent objects are not observed.

I would like to use @Api... annotations on both my Java interface as well as on an (abstract) base class to be observed when the API documentation for my actual resource class is generated.

okigan:
Particularly painful as cannot define interfaces separately and reuse that with automatic client generation (ex. using https://github.com/Netflix/feign)

webron:
Fixed in develop_2.0, using #1122.

PR #1074, opened 2015-05-16
Add @Inherited to @ApiOperation

okigan:
Fix https://github.com/swagger-api/swagger-core/issues/1072

### DIFF
--- a/modules/swagger-annotations/src/main/java/com/wordnik/swagger/annotations/ApiOperation.java
+++ b/modules/swagger-annotations/src/main/java/com/wordnik/swagger/annotations/ApiOperation.java
@@ -28,6 +28,7 @@ import java.lang.annotation.Target;
  */
 @Target(ElementType.METHOD)
 @Retention(RetentionPolicy.RUNTIME)
+@Inherited
 public @interface ApiOperation {
   /** Brief description of the operation  */
   String value();


fehguy:
Hi @okigan what is the purpose of this?

olensmar:
according to the javadoc for @Inherited - it is only applicable to annotations for types/classes - so I'm not sure why you want to add it here. https://docs.oracle.com/javase/7/docs/api/java/lang/annotation/Inherited.html

okigan:
@olensmar hmm, good point, the issue still remains: methods from base class/interface annotated with ApiOperation do not show up in the documentation....

olensmar:
ok - I agree that's a problem, but there might already be an issue for it - @webron, can you shed some light?

okigan:
@webron @olensmar I think then this part is relevant: I *think* the javax.ws.rs.GET annotation is not picked up when a method is derived overridden, so extractOperationMethod needs to traverse up to the base base method while looking for javax.ws.rs.* annotations, or there is another way ?  

	public String extractOperationMethod(ApiOperation apiOperation, Method method, Iterator<SwaggerExtension> chain) {
		if(apiOperation.httpMethod() != null && !"".equals(apiOperation.httpMethod()))
		  return apiOperation.httpMethod().toLowerCase();
		else if(method.getAnnotation(javax.ws.rs.GET.class) != null)
		  return "get";
		else .....
		}
		else if(chain.hasNext())
		  return chain.next().extractOperationMethod(apiOperation, method, chain);
		else
		  return null;
	}

webron:
https://github.com/swagger-api/swagger-core/tree/develop_2.0/samples/java-resteasy-spring uses annotations on the interfaces and loads just fine.

Perhaps there are other paradigms that are not covered and those should be handled.

olensmar:
@webron unfortunately different frameworks have different levels of support in regard to traversal, annotations on interfaces, etc.

If we add support for this we need to do it consistently for all annotations - and we need to figure out how to handle merging/overriding annotations at different levels

webron:
Just to be clear, when you say different frameworks, what frameworks are you talking about?

olensmar:
different implementations of jax-rs

webron:
There's a Jersey sample as well exhibiting the behavior as well. It *could* depend on whether we get the resources from the jax-rs implementation or whether those are fed directly to swagger-core.

@okigan - can you share a sample project in which you encounter the problem? It'd definitely help us investigate it further and provide a fuller solution.

okigan:
@webron i'll try to modify java-jersey-jaxrs sample

webron:
Thanks. If there is a problem, we definitely need to fix it. Keep in mind that in the develop_2.0 branch there's a Jersey sample with Spring that also uses annotations on the interfaces.

okigan:
Here is a branch with modification: https://github.com/okigan/swagger-core/tree/base-resource-swagger

before modification: 
screen shot 2015-05-16 at 10 11 30 am


after, Note, *pet interface does not show up anymore*, but it is detected (see next image): 

screen shot 2015-05-16 at 10 15 46 am


swagger.json file after modification, note interface detected, but no methods (btw, would be nice if swagger.js would still show it): 
screen shot 2015-05-16 at 10 17 19 am

dilipkrish:
I believe for methods you need to walk up the object class hierarchy. Which might be alright for methods. But I know that in springfox Weve refrained from supporting annotated method parameters. We've left that as an extensitibility option.

okigan:
@fehguy as we still in discussion, would you re-open this issue?


Also any more info needed from my side?

webron:
@okigan - since this is a PR, I'd suggest opening an issue rather than reopen the PR.

okigan:
Oh I see

webron:
Also, unfortunately the sample you shared doesn't exhibit the problem as there's a small bug in it. You forgot to actually extend the base resource. I've submitted a small PR that fixes it - https://github.com/okigan/swagger-core/pull/1. With that change, it works as expected.

okigan:
@webron hmm, yeah seems to work, odd i'll see what's the difference between this sample and my actual project

okigan:
@webron ok, forgot the key change -- to actually override the method, and now it does not work (i am sure of it)

branch: https://github.com/okigan/swagger-core/tree/base-resource-swagger

before and after snapshot (note the GET method is gone):

screen shot 2015-05-16 at 8 36 44 pm
screen shot 2015-05-16 at 8 37 12 pm

okigan:
@webron bump ;)

Issue #1085, opened 2015-05-20
PathParam in Class @Path Fails to Create Required Parameters for Resource Methods


nkoterba:
We have quite a few REST Resources that utilize a PathParam value in the class's `@Path` definition.

For example:
```java
@Path("/api/workspaces/{workspaceId}/items")
@Api(value = "items")
@Produces(MediaType.APPLICATION_JSON)
@Consumes(MediaType.APPLICATION_JSON)
public class TestResource extends EndpointBase {

    public TestResource(@PathParam("workspaceId") long workspaceId) throws WorkspaceException
	{
		// Configures a class-wide resource using workspaceId/PathParam, e.g. DAO
	}
...
}
```

In that class, we have a get method:
```java
    @GET
	@ApiOperation(value = "Get all items")
	@ApiResponses(value = { @ApiResponse(code = 400, message = "Invalid ID supplied") })
	public Collection<Item> getItems(@QueryParam("skip") Integer skip, @QueryParam("take")
	Integer take) {
              // Uses item (e.g. DAO) configured in constructor with PathParam to get items
        }
```

Unfortunately, the swagger info generated does not recognize that `workspaceId` is required for the get method and doesn't create it in the UI:
```javascript
---
swagger: "2.0"
info:
  version: "1"
basePath: "/app/rest"
tags:
- name: "items"
  description: ""
paths:
  /api/workspaces/{workspaceId}/items:
    get:
      tags:
      - "items"
      summary: "Get all items"
      description: ""
      operationId: "getItems"
      consumes:
      - "application/json"
      produces:
      - "application/json"
      parameters:
      - name: "skip"
        in: "query"
        required: false
        type: "integer"
        format: "int32"
      - name: "take"
        in: "query"
        required: false
        type: "integer"
        format: "int32"
      responses:
        200:
          description: "successful operation"
          schema:
            type: "array"
            items:
              type: "object"
        400:
          description: "Invalid ID supplied"
```

And the UI:
![image](https://cloud.githubusercontent.com/assets/2508944/7717591/e7fce192-fe58-11e4-919d-114d552c1889.png)

Thus, it's impossible to *test* the API, because there's no input box for `workspaceId` on the **GET** method **AND** the test URL will also be incorrectly formatted (doesn't replace workspaceId with a valid value), e.g.: 
```javascript
GET https://localhost:8443/app/rest/api/workspaces/%7BworkspaceId%7D/items?skip=0&take=100 404 (Not Found)
```

[...]

PR #1088, opened 2015-05-20
Extension annotations

olensmar:
This PR adds a number of features that allows for customisation of the generated Swagger definition:
- SwaggerConfig and related annotations that allow one to define metadata at the definition level (info, consumes, produces, tags, externalDocs)
- Extension-related annotations allowing one to add vendor-extensions at the info, tags and operation level
- a ReaderListener interface that allows for customisation of the actual Swagger generation process
- support for all these in the Reader class
- a unit test to validate the correct reading of the annotations into the Swagger model


modules/swagger-annotations/src/main/java/com/wordnik/swagger/annotations/SwaggerConfig.java
```diff
@@ -0,0 +1,110 @@
...
+ * - Security Definitions
+ * - Security Requirements
+ * - Parameters
+ * - Responses
+ */
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+public @interface SwaggerConfig {
```
olensmar: Perhaps Definition or SwaggerDefinition would be better names for this - thoughts anyone?

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
```diff
@@ -70,9 +102,31 @@ public Reader(Swagger swagger, ReaderConfig config) {
     this.config = new DefaultReaderConfig(config);
   }
 
-  public Swagger read(Set<Class<?>> classes) {
+  public synchronized Swagger read(Set<Class<?>> classes) {
+
+    listeners.clear();
```
tomtit: I would suggest to convert `listeners` into local variable as we reinitialize it on each read.

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
```diff
@@ -70,9 +102,31 @@ public Reader(Swagger swagger, ReaderConfig config) {
...
+
+    for(Class<?> cls: classes) {
+      if( ReaderListener.class.isAssignableFrom( ReaderListener.class ) && listeners.containsKey( cls )){
```
tomtit: This condition looks strange. The first part is always true, the second one is always false as the `listeners` map is empty after the `clean()` call.

olensmar: wow - thanks for catching that one - will fix this and the above..

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
```diff
@@ -586,6 +817,19 @@ else if(!responseClass.equals(java.lang.Void.class) && !"void".equals(responseCl
     return operation;
   }
```
tomtit: This method is obsolete, it was moved into the ContainerWrapper class.

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
olensmar: it's used in the other read method to detect of the class being read is read independently...

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
olensmar: ok - removing..

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
```diff
@@ -279,9 +364,155 @@ protected Swagger read(Class<?> cls, String parentPath, String parentMethod, boo
...
+
+        if( !name.startsWith("x-")){
+          name = "x-" + name;
+        }
+
+        if( !map.containsKey( name )) {
+          map.put(name, new HashMap<String, Object>());
+        }
+
+        map = (Map<String, Object>) map.get(name);
```
tomtit: We can speed up these lines for free :)
```
Map<String, Object> value = map.get(name)
if (value == null) {
  value = new HashMap<String, Object>();
  map.put(name, value)
}
```

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
```diff
@@ -70,9 +102,31 @@ public Reader(Swagger swagger, ReaderConfig config) {
     this.config = new DefaultReaderConfig(config);
   }
 
-  public Swagger read(Set<Class<?>> classes) {
+  public synchronized Swagger read(Set<Class<?>> classes) {
+
+    listeners.clear();
+
+    for(Class<?> cls: classes) {
+      if( ReaderListener.class.isAssignableFrom( cls ) && listeners.containsKey( cls )){
```
tomtit: Second condition still lacks for '!' operator.
Do you think it's safe to instantiate REST related resources? Please take a look at the resource from the #1085. To add a listener for the TestResource class the author will need to rework his/her class and add a default constructor.

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
olensmar: I don't think performance will be an issue here...

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
olensmar: makes sense - the only objection I have is that you need to know at code-time which listeners you want; I was aiming at a solution where listeners can be added later without having to modify the source code (since the main scan method will find any class implementing this interface, those classes could be added later, for example by scanning a 3rd party library that adds additional Swagger-related features).

tomtit: My point was wrong. I didn't realize initially that we don't need to add implementation of the ReaderListener into our REST resources.
So, having this understanding I have a concern about the `read(Class<?> cls)` method. Inside this method method we don't add our listener into the `listeners` map. If we have two subsequent calls of this method it's possible to activate listener for itself only:
```
reader.read(MyListener.class);
reader.read(MyREST.class); // MyREST will be hidden from MyListener
```
As for now we always call this method from `read(Set<Class<?>> classes)`. Perhaps it makes sense to declare  `read(Class<?> cls)` as private or protected.

olensmar: yes you're right - i'm not sure if this method can be made private or protected - it is called directly by the tests (which we could change) - let me check the actual code...

modules/swagger-annotations/src/main/java/com/wordnik/swagger/annotations/SwaggerConfig.java
webron: We have a SwaggerConfig interface (https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/com/wordnik/swagger/config/SwaggerConfig.java) so it's probably better to rename. If this is an interface, I'd just use the full name SwaggerConfiguration.

jsiqueiraRF:
Hi, I've had a chance to test the current branch and it works for my use case, which is simply to be able to add a custom vendor field to the Operation.

A comment, what is the rationale behind the two levels of annotations (ie, `@Extension` takes an array of `@ExtensionProperty`)? For a simple case like mine only `@ExtensionProperty` would be needed and I'm wondering what other use cases there might be.

olensmar:
Thanks!

The rationale is to allow for more complex vendor extensions - for example

```javascript
"x-something" : {
    "p1" : "value",
    "p2" : "value",
    "p3" : "value"
}
```

instead of

```javascript
"x-something-p1" : "value",
"x-something-p2" : "value",
"x-something-p3" : "value"
```

Hope that makes sense :-)

PR #1090, opened 2015-05-21
Develop 2.0

Amyntas:
Added java-servlet module and sample with proper class reading for servlets with @ApiImplicitParameterS


webron:
@Amyntas - A huge thanks for the work here, but I'm afraid we'll have to rework it due to some changes. Someone is already looking into it and will use your PR as a reference. Again, the effort is greatly appreciated.

PR #1111, opened 2015-05-28
Issue 905 implicit params

olensmar:


webron:
Perhaps add a test for body param as well?

olensmar:
done... waiting for build..

Issue #918, opened 2015-03-17
Migrate play module


fehguy:
Using the swagger-scala module, need to migrate play support into a new module.  This is probably best done in a new artifact, and outside of swagger-core.

seantbrady:
+1

joantune:
+1 please, if possible, take into account  https://github.com/swagger-api/swagger-core/issues/900

jcranky:
Since the integration will be reworked, it would be nice if more information like path and query params could be automatically extracted, so that we don't have to use ```jax-rs```'s ```PathParam``` and ```QueryParam``` annotations. This is information is already available in the ```routes``` file.

alpeb:
Also it seems models inside scala objects aren't properly supported.
I modified `/samples/scala-play2/app/models/Pet.scala` (using tag `v.1.3.12`) like this:
```
object Models {
  @ApiModel("Pet")
  case class Pet(
    @(ApiModelProperty @field)(position=1, value="id")id: Long,
    @(ApiModelProperty @field)(position=2)category: Category,
    @(ApiModelProperty @field)(position=3)name: String,
    @(ApiModelProperty @field)(position=4)photoUrls: List[String],
    @(ApiModelProperty @field)(position=5)tags: List[Tag],
    @(ApiModelProperty @field)(position=6)status: String
  )

  @ApiModel("Tag")
  case class Tag (
    @(ApiModelProperty @field)(position=1)id: Long,
    @(ApiModelProperty @field)(position=2)name: String)
}
```

And the `Tag` model disappeared from the json in `http://localhost:9000/api-docs/pet`

Can someone please confirm this is indeed a bug?

fehguy:
@alpeb if you're looking at how this works in the 1.5.x version, you should look at the [swagger-scala-module](https://github.com/swagger-api/swagger-scala-module) which will support scala

alpeb:
Are there any instructions on how to use https://github.com/swagger-api/swagger-scala-module ? All I see there is the `SwaggerScalaModelConverter` class.

I would love to make this work now, that's why I was trying with the `v.1.3.12` tag.

fehguy:
I'll get that updated on the project, please track https://github.com/swagger-api/swagger-scala-module/issues/2

fehguy:
ok @alpeb please take a look at the homepage for the scala module for instructions on adding it.

alpeb:
[...]

PR #1115, opened 2015-05-29
Develop scala 2.11

manishgoyaler:
Adding swagger UI to complete this sample app.


webron:
Thanks for the PR. Since we're not actively developing the 1.3 branch anymore, I'll be closing it, but provided a link from #918 to be taken into account when the Play module is migrated.

PR #1126, opened 2015-06-03
Fixed #1085: PathParam in Class @Path Fails to Create Required Parameters for Resource Methods

### LINKED ISSUE #1085: PathParam in Class @Path Fails to Create Required Parameters for Resource Methods

iushankin:
Implemented types of injection:
* Class field
* Constructor parameters

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
@@ -1,11 +1,14 @@
 package com.wordnik.swagger.jaxrs;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -17,19 +20,23 @@ import java.util.Map;
 import java.util.Set;
 
 import javax.ws.rs.Consumes;
+import javax.ws.rs.HeaderParam;
 import javax.ws.rs.HttpMethod;
+import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 
-import com.wordnik.swagger.jaxrs.utils.ReflectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.JavaType;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.type.TypeFactory;
+import com.google.common.collect.Collections2;
 import com.wordnik.swagger.annotations.Api;
+import com.wordnik.swagger.annotations.ApiImplicitParam;
 import com.wordnik.swagger.annotations.ApiOperation;
+import com.wordnik.swagger.annotations.ApiParam;
 import com.wordnik.swagger.annotations.ApiResponse;
 import com.wordnik.swagger.annotations.ApiResponses;
 import com.wordnik.swagger.annotations.Authorization;
@@ -39,6 +46,7 @@ import com.wordnik.swagger.jaxrs.config.DefaultReaderConfig;
 import com.wordnik.swagger.jaxrs.config.ReaderConfig;
 import com.wordnik.swagger.jaxrs.ext.SwaggerExtension;
 import com.wordnik.swagger.jaxrs.ext.SwaggerExtensions;
+import com.wordnik.swagger.jaxrs.utils.ReflectionUtils;
 import com.wordnik.swagger.models.Model;
 import com.wordnik.swagger.models.Operation;
 import com.wordnik.swagger.models.Path;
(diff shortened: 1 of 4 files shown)


modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
```diff
@@ -53,16 +61,25 @@
...
   Swagger swagger;
   private final ReaderConfig config;
 
+  static {
+    final Set<Class<? extends Annotation>> fieldAnnotations = new HashSet<Class<? extends Annotation>>();
+    fieldAnnotations.add(PathParam.class);
+    fieldAnnotations.add(QueryParam.class);
+    fieldAnnotations.add(HeaderParam.class);
+    fieldAnnotations.add(ApiParam.class);
+    fieldAnnotations.add(ApiImplicitParam.class);
```
olensmar: don't we need to add FormParam as well?

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
```diff
@@ -710,6 +735,31 @@ private static boolean isResourceClass(Class<?> cls) {
     return cls.getAnnotation(Api.class) != null;
   }
 
+  private List<Parameter> collectGlobalParameters(Class<?> cls) {
```
olensmar: This method looks like a candidate to put in a ReaderUtils class (or perhaps even ReflectionUtils) ?

modules/swagger-jaxrs/src/test/scala/ReaderTest.scala
```diff
@@ -80,6 +81,24 @@ class ReaderTest extends FlatSpec with Matchers {
...
+    resourceParameters.get(1).asInstanceOf[QueryParameter].getName() should equal ("fieldParam")
+    resourceParameters.get(2).asInstanceOf[QueryParameter].getName() should equal ("methodParam")
+
+    val subResourceParameters = swagger.getPaths().get("/resource/{id}/subresource").getGet().getParameters()
+    subResourceParameters should not equal (null)
+    subResourceParameters.size() should equal (3)
+    subResourceParameters.get(0).asInstanceOf[PathParameter].getName() should equal ("id")
+    subResourceParameters.get(1).asInstanceOf[QueryParameter].getName() should equal ("fieldParam")
+    subResourceParameters.get(2).asInstanceOf[QueryParameter].getName() should equal ("subResourceParam")
+  }
```
olensmar: Perhaps add tests for all the other types as well?

olensmar:
please rebase also...

modules/swagger-jaxrs/src/test/scala/ReaderTest.scala
iushankin: Okay

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
iushankin: Seems no. This method depends on the private method Reader#getParameters

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
iushankin: From the spec: "The @FormParam annotation is special and may only be utilized on resource and sub-resource methods. This is because it extracts information from a request entity. "
Please look: https://jersey.java.net/documentation/latest/jaxrs-resources.html#d0e2731
Also look old implementation: https://github.com/swagger-api/swagger-core/blob/399bb429803c03cff9aa474057c72df39026787f/modules/swagger-jaxrs/src/main/scala/com/wordnik/swagger/jaxrs/JaxrsApiReader.scala#L216-L219

modules/swagger-jaxrs/src/main/java/com/wordnik/swagger/jaxrs/Reader.java
olensmar: ok - that could also be moved out perhaps - but lets not touch these for now... thanks

olensmar:
i've merged this in https://github.com/swagger-api/swagger-core/pull/1136

webron:
Merged manually.

PR #1130, opened 2015-06-04
Fixed #1120: Models handling has been reworked.

### LINKED ISSUE #1120: @ApiResponse.response of primitive types is reported incorrectly
Swagger version is 1.5.3-M2-SNAPSHOT (branch develop_2.0)
This test resource:
```java
@Api
@Path("xxx")
public class ResourceXXX {

  @GET
  @Path("test")
  @ApiResponses({ @ApiResponse(code = 400, message = "Message for URL", response = URL.class),
      @ApiResponse(code = 401, message = "Message for Long", response = Long.class),
      @ApiResponse(code = 402, message = "Message for String", response = String.class) })
  public Response test() {
    return null;
  }
}
```
is reported as follows:
```json
{
  "swagger" : "2.0",
  "paths" : {
    "/xxx/test" : {
      "get" : {
        "operationId" : "test",
        "parameters" : [ ],
        "responses" : {
          "400" : {
            "description" : "Message for URL",
            "schema" : {
              "$ref" : "#/definitions/string"
            }
          },
          "401" : {
            "description" : "Message for Long"
          },
          "402" : {
            "description" : "Message for String"
          }
        }
      }
    }
  },
  "definitions" : {
    "string" : {
      "type" : "object",
      "properties" : {
        "protocol" : {
          "type" : "string"
        },
        "host" : {
          "type" : "string"
        },
        "port" : {
          "type" : "integer",
          "format" : "int32"
        },
        "file" : {
          "type" : "string"
        },
        "authority" : {
          "type" : "string"
        },
        "ref" : {
          "type" : "string"
        },
        "path" : {
          "type" : "string"
        },
        "query" : {
          "type" : "string"
        },
        "userInfo" : {
          "type" : "string"
        },
        "defaultPort" : {
          "type" : "integer",
          "format" : "int32"
        },
        "content" : {
          "type" : "object"
        }
      }
    }
  }
}
```
As you can see `URL` was reported as composite class of type `string`, `Long` and `String` were not reported at all.

tomtit:
Fixed #1120: Models handling has been reworked.

Actually the #1120 has to separate issues. The first one that swagger doesn't fill the [Response.schema](https://github.com/swagger-api/swagger-core/blob/6875dfe0cc5e7f3ec08e49f4d0eb9c5e6832bfd3/modules/swagger-models/src/main/java/io/swagger/models/Response.java#L11) property for primitive types and the second that `URL` class is treated as complex type `"$ref" : "#/definitions/string"` (https://github.com/swagger-api/swagger-core/issues/1120#issue-83516910) while it should be reported as a primitive `string` with the `url` format.

The first issue rises here https://github.com/swagger-api/swagger-core/blob/6875dfe0cc5e7f3ec08e49f4d0eb9c5e6832bfd3/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java#L820-840 - primitive types are not processed at all.
There are 6 similar places where we instantiate implementation of the `Property` interface to reference a type:
1. https://github.com/swagger-api/swagger-core/blob/6875dfe0cc5e7f3ec08e49f4d0eb9c5e6832bfd3/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java#L97-118
2. https://github.com/swagger-api/swagger-core/blob/6875dfe0cc5e7f3ec08e49f4d0eb9c5e6832bfd3/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java#L122-173
3. https://github.com/swagger-api/swagger-core/blob/6875dfe0cc5e7f3ec08e49f4d0eb9c5e6832bfd3/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java#L656-662
4. https://github.com/swagger-api/swagger-core/blob/6875dfe0cc5e7f3ec08e49f4d0eb9c5e6832bfd3/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java#L741-776
5. https://github.com/swagger-api/swagger-core/blob/6875dfe0cc5e7f3ec08e49f4d0eb9c5e6832bfd3/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/ParameterProcessor.java#L128-159
6. https://github.com/swagger-api/swagger-core/blob/6875dfe0cc5e7f3ec08e49f4d0eb9c5e6832bfd3/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/ParameterProcessor.java#L162-187
All these pieces of code do the same thing but differently. It would be nice to extract this functionality into some utility class and re-use from everywhere. But it appeared that it's enough to call `ModelConverters.getInstance().readAsProperty(...)` to instantiate  implementation of the `Property` interface and `ModelConverters.getInstance().readAll(...)` to collect all referenced models.
The only difference for `ParameterProcessor` is that `Property` has to be converted into `Model`, that is why the `io.swagger.models.properties.PropertyBuilder.toModel(...)` method has been added.

The second issue arises here https://github.com/swagger-api/swagger-core/blob/6875dfe0cc5e7f3ec08e49f4d0eb9c5e6832bfd3/modules/swagger-core/src/main/java/io/swagger/converter/ModelConverters.java#L101, so `ModelConverters.read(...)` and `ModelConverters.readAll(...)` calls return empty maps for `java.lang.String` and non-empty maps for `java.net.URL` and `java.net.URI` but should not.
So, added the `io.swagger.jackson.TypeNameResolver.isStdType(JavaType)` method and used it in `io.swagger.jackson.ModelResolver` to force null models for primitive types.

After all these changes two old tests failed:
* EnumTest was expecting non null model for enum class but enums are treated as strings and no models should be created for them
* ModelWithTuple2 was incorrect as it merged `Pair<String, String>` and `Pair<ComplexLeft, String>` into a single model

So, both tests were reworked.

### DIFF
--- a/modules/swagger-jaxrs/src/test/scala/SimpleScannerTest.scala
+++ b/modules/swagger-jaxrs/src/test/scala/SimpleScannerTest.scala
@@ -2,12 +2,13 @@ import io.swagger.jaxrs.Reader
 import io.swagger.jaxrs.config.DefaultReaderConfig
 import io.swagger.models.parameters.{BodyParameter, PathParameter, QueryParameter, SerializableParameter}
 import io.swagger.models.properties.{MapProperty, _}
-import io.swagger.models.{Model, ModelImpl, RefModel, Swagger}
-import io.swagger.util.Json
+import io.swagger.models.{ArrayModel, Model, ModelImpl, RefModel, Swagger}
 import models.TestEnum
 import org.junit.runner.RunWith
 import org.scalatest.{FlatSpec, Matchers}
 import org.scalatest.junit.JUnitRunner
+
+import models.TestEnum
 import resources._
 
 import scala.collection.JavaConverters._
@@ -195,7 +196,11 @@ class SimpleScannerTest extends FlatSpec with Matchers {
 
   it should "correctly model an empty model per 499" in {
     val swagger = new Reader(new Swagger()).read(classOf[ResourceWithEmptyModel])
-    Json.prettyPrint(swagger)
+    swagger.getDefinitions.size should be (1)
+    val empty = swagger.getDefinitions.get("EmptyModel").asInstanceOf[ModelImpl]
+    empty.getType should be ("object")
+    empty.getProperties should be (null)
+    empty.getAdditionalProperties should be (null)
   }
   
   it should "scan a simple resource without annotations" in {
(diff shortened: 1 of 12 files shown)


pom.xml
```diff
@@ -515,7 +515,7 @@
     <jackson-guava-version>2.4.2</jackson-guava-version>
     <logback-version>1.0.1</logback-version>
 
-    <junit-version>4.8.1</junit-version>
+    <junit-version>4.8.2</junit-version>
```
tomtit: Need this for `org.junit.Assert.assertThat(...)` and `org.hamcrest.CoreMatchers.instanceOf(Class<?>)`

modules/swagger-jaxrs/src/test/scala/SimpleScannerTest.scala
```diff
@@ -195,7 +196,11 @@ class SimpleScannerTest extends FlatSpec with Matchers {
 
   it should "correctly model an empty model per 499" in {
     val swagger = new Reader(new Swagger()).read(classOf[ResourceWithEmptyModel])
-    Json.prettyPrint(swagger)
```
tomtit: `Json.prettyPrint(swagger)` checks nothing

olensmar:
to be honest - this is somewhat above my head :-) What problem does this PR solve?

tomtit:
Added detailed explanation on my changes. Hope it'll help.

tomtit:
Rebased.

olensmar:
great - thanks for the explanation - I don't think I would have figured that out myself :-)

fehguy:
Looks great--samples are passing validation with this change, I'm merging.

PR #1145, opened 2015-06-09
Improved accept header for swagger request.

MarvGilb:
1. Outsourced code for swagger creation, so that duplicate code is
reduced.
2. Changed path annotations of getListingJson and getListingYaml and a
new methods that supports swagger with .yaml or .json at the end.
Through this swagger now supports 3 ways to request the documentation.
/swagger , /swagger.json & /swagger.yaml
This way URL-based negotiation (http://docs.jboss.org/resteasy/docs/2.3.7.Final/userguide/html/JAX-RS_Content_Negotiation.html#media_mappings) of RESTeasy works, if a media type mapping from .json to application/json is configured.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/listing/ApiListingResource.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/listing/ApiListingResource.java
@@ -78,67 +78,74 @@ public class ApiListingResource {
         return swagger;
     }
 
+    private Swagger process(
+            @Context Application app,
+            @Context ServletConfig sc,
+            @Context HttpHeaders headers,
+            @Context UriInfo uriInfo) {
+        Swagger swagger = (Swagger) context.getAttribute("swagger");
+        if (!initialized)
+            swagger = scan(app, sc);
+        if (swagger != null) {
+            SwaggerSpecFilter filterImpl = FilterFactory.getFilter();
+            if (filterImpl != null) {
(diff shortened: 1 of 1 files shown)


webron:
@MarvGilb - thanks for the PR, looks interesting and can also solve #1169 elegantly (as @tomtit kindly pointed out to me).

Two points - 
1. This needs to be submitted against the develop_2.0 branch, so please resubmit it.
2. With the new [coding standards](https://github.com/swagger-api/swagger-core/wiki/Coding-Standards#block-less-loops-and-if-statements) we now employ, please add blocks for the if statements, even if they have one line in them.

modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/listing/ApiListingResource.java
```diff
@@ -78,67 +82,74 @@ protected synchronized Swagger scan(Application app, ServletConfig sc) {
...
-    @GET
-    @Produces(MediaType.APPLICATION_JSON)
-    @Path("/swagger.json")
-    @ApiOperation(value = "The swagger definition in JSON", hidden = true)
-    public Response getListingJson(
+    private Swagger process(
             @Context Application app,
             @Context ServletConfig sc,
             @Context HttpHeaders headers,
             @Context UriInfo uriInfo) {
```
tomtit: I think the `@Context` annotations can be removed as well.

MarvGilb:
New PR: Improved accept header for swagger request. #1172

PR #1158, opened 2015-06-11
Properly resolve a inner model with a self/cyclic reference.

msimons:
Properly resolve a inner model with a self/cyclic reference.

In the following example the implementation was lacking to generate a property reference from Model B to Model B. 

[Model A] -> [Model B] -> [Model B] 

I extented the 'ReferenceTest ' to cover this situation.


webron:
@msimons - Development is done in the develop_2.0 branch, so please submit your PR against it.

msimons:
@webron Thanks for informing me. I will submit it in develop_2.0 branch.

PR #1159, opened 2015-06-11
Fixed #1155: Names inspection has been fixed.

### LINKED ISSUE #1155: ModelResolver throws an exception if a property name equals "is" or "get"
If the entire name of a property is just "is" or "get", you'll get an exception from line 249 or 253 of `ModelResolver` because it's trying to parse it as a bean accessor name without length-checking. (Seen on develop_2.0 branch)

tomtit:
Fixed #1155: Names inspection has been fixed.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
@@ -245,13 +245,13 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
                 if (member != null) {
                     String altName = member.getName();
                     if (altName != null) {
-                        if (altName.startsWith("get")) {
-                            if (!Character.isUpperCase(altName.charAt(3))) {
+                        final int length = altName.length();
+                        for (String prefix : Arrays.asList("get", "is")) {
+                            final int offset = prefix.length();
+                            if (altName.startsWith(prefix) && length > offset
+                                    && !Character.isUpperCase(altName.charAt(offset))) {
                                 propName = altName;
-                            }
-                        } else if (altName.startsWith("is")) {
-                            if (!Character.isUpperCase(altName.charAt(2))) {
-                                propName = altName;
+                                break;
                             }
                         }
                     }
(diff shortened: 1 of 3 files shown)


modules/swagger-core/src/test/scala/ModelConverterTest.scala
```diff
@@ -240,8 +241,9 @@ class ModelConverterTest extends FlatSpec with Matchers {
 
   it should "override the property name" in {
     val schemas = ModelConverters.getInstance().readAll(classOf[ModelWithAltPropertyName])
-    val model = schemas.get("sample_model").asInstanceOf[ModelImpl]
-    Json.prettyPrint(model)
+    val properties = schemas.get("sample_model").asInstanceOf[ModelImpl].getProperties
+    properties.get("id") should be (null)
+    properties.get("the_id") should not be null
```
tomtit: Let's check renamed property instead of printing to console.

modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
```diff
@@ -245,13 +245,13 @@ public Model resolve(JavaType type, ModelConverterContext context, Iterator<Mode
...
-                            }
-                        } else if (altName.startsWith("is")) {
-                            if (!Character.isUpperCase(altName.charAt(2))) {
+                        final int length = altName.length();
+                        for (String prefix : Arrays.asList("get", "is")) {
+                            final int offset = prefix.length();
+                            if (altName.startsWith(prefix) && length > offset
+                                    && !Character.isUpperCase(altName.charAt(offset))) {
                                 propName = altName;
+                                break;
```
webron: This makes me sad.

tomtit: Could you clarify this?

webron: `break` statements in Java make me sad :) I have a couple of questions about the PR, but they'll wait for next week.

PR #1162, opened 2015-06-12
Add HTTP "head" method support in Path model

emailchaos:


webron:
Thanks for the submission. Two notes though:
1. Code changes need to be made against the develop_2.0 branch, so please resubmit it against it.
2. In terms of `@JsonPropertyOrder` I would push it at the end, or at least after `options`.

bofalot:
I am interested in this change, thanks for fixing. Just my $0.02, but I would keep ```head``` after ```get``` in the ```@JsonPropertyOrder``` as that is the order listed at http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html.

modules/swagger-models/src/main/java/io/swagger/models/Path.java
```diff
@@ -49,6 +53,11 @@ public Path get(Operation get) {
         return this;
     }
 
+    public Path head(Operation get) {
```
bofalot: The function argument should be ```head``` not ```get```. This typo prevents this change from actually working.

webron:
I'm afraid I have to close it as it's submitted against the wrong branch. Feel free to resubmit against develop_2.0.

bofalot:
I've included this into my pull request https://github.com/swagger-api/swagger-core/pull/1217

PR #1163, opened 2015-06-12
[Develop_2.0] Properly resolve a inner model with a self/cyclic references

msimons:
Properly resolve a inner model with a self/cyclic reference.

In the following example the implementation was lacking to generate a property reference from Model B to Model B.

[Model A] -> [Model B] -> [Model B]

I extented the 'ReferenceTest ' to cover this situation.


modules/swagger-core/src/test/scala/1_3/converter/SnakeCaseConverterTest.scala
```diff
@@ -95,7 +95,9 @@ class SnakeCaseConverter extends ModelConverter {
         val name = (model.asInstanceOf[ModelImpl]).getName
         if (model.isInstanceOf[ModelImpl]) {
```
tomtit: Here we have another issue that doesn't relate to the initial one. The [ComposedModel](https://github.com/swagger-api/swagger-core/blob/400662d7253c0c2efb14316f8948ddc61a60069e/modules/swagger-models/src/main/java/io/swagger/models/ComposedModel.java) is not supported here. So, this converter will update the references to composed models but leave model names untouched.

PR #1166, opened 2015-06-14
Develop 2.0 - support binary input and output

boazsapir:
See my proposed spec https://github.com/swagger-api/swagger-spec/issues/50

These changes in swagger-models are a precondition for an implementation in swagger-codegen of the proposed spec update, as detailed in the above mentioned swagger-spec issue #50, to support binary input and output


boazsapir:
See also https://github.com/swagger-api/swagger-codegen/pull/868

webron:
We still need to make a decision on that issue before merging this PR.

webron:
@boazsapir - finally clearing out the PRs (as much as I can). Based on our discussion, I imagine this is no longer relevant (that is, it needs to be modified)?

boazsapir:
I agree.
BTW is the new proposal confirmed or are we still waiting for input from more people?

webron:
I believe I talked with Tony about it offline, but let me just make sure :)

PR #1171, opened 2015-06-16
Fixed #1154: Handling of BigDecimal and BigInteger classed has been improved.

### LINKED ISSUE #1154: BigDecimal and BigInteger are misrepresented in the output
`BigDecimal` --> `type: number`
`BigInteger` --> `type: integer`

Shouldn't have `format` in either case.

tomtit:
Fixed #1154: Handling of BigDecimal and BigInteger classed has been improved.
Mapping of a small set of "well-known" classes into Swagger properties is implemented via hard-coded string identifiers of supported types. These string identifiers are defined in two classes: [TypeNameResolver](https://github.com/swagger-api/swagger-core/blob/cfd8eda836da8bd49b117259099d914b6ad4b19b/modules/swagger-core/src/main/java/io/swagger/jackson/TypeNameResolver.java#L48-78) and [AbstractModelConverter](https://github.com/swagger-api/swagger-core/blob/cfd8eda836da8bd49b117259099d914b6ad4b19b/modules/swagger-core/src/main/java/io/swagger/jackson/AbstractModelConverter.java#L91-121). This approach has two main drawbacks:
1. String identifiers have to be synchronized between two classes;
2. String identifiers are visible via `@ApiModelProperty.dataType()` and `@ApiImplicitParam.dataType()` annotations, so some internal type identifiers like `uri`, `url` and `uuid` are published indirectly as standard primitive types.
To overcome mentioned drawbacks and fix handling of BigDecimal and BigInteger types I added the `io.swagger.util.PrimitiveType` enumeration.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/jackson/AbstractModelConverter.java
+++ b/modules/swagger-core/src/main/java/io/swagger/jackson/AbstractModelConverter.java
@@ -13,17 +13,7 @@ import io.swagger.annotations.ApiModelProperty;
 import io.swagger.converter.ModelConverter;
 import io.swagger.converter.ModelConverterContext;
 import io.swagger.models.Model;
-import io.swagger.models.properties.BooleanProperty;
-import io.swagger.models.properties.DateProperty;
-import io.swagger.models.properties.DateTimeProperty;
-import io.swagger.models.properties.DoubleProperty;
-import io.swagger.models.properties.FloatProperty;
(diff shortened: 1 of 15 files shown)


webron:
Can you rebase this one?

tomtit:
Rebased.
The `io.swagger.converter.PrimitiveType` enum from the PR #1152 was removed as duplicate.

PR #1172, opened 2015-06-16
Improved accept header for swagger request.

MarvGilb:
1. Outsourced code for swagger creation, so that duplicate code is
reduced.
2. Changed path annotations of getListingJson and getListingYaml and a
new methods that supports swagger with .yaml or .json at the end.
Through this swagger now supports 3 ways to request the documentation.
/swagger , /swagger.json & /swagger.yaml This way URL-based negotiation
(http://docs.jboss.org/resteasy/docs/2.3.7.Final/userguide/html/JAX-RS_Content_Negotiation.html#media_mappings)
of RESTeasy works, if a media type mapping from .json to
application/json is configured.


webron:
@MarvGilb - unfortunately the build is failing. Can you have a look?

PR #1182, opened 2015-06-21
Added Jackson Annotation driven deserialization.

gdilipkumar:
Added custom deserializers for swagger polymorphic types (Property, Parameter, Model and SecuritySchemeDefinition). These will work with Jackson Annotation without requiring any object mapper configuration.

NOTE: After adding these annotations to Swagger Models (Paramter, Model, Property and SecuritySchemeDefinition), some test cases in Swagger-core are failing. As i don't have experience in scala i haven't fixed those test cases.


webron:
@gdilipkumar - a few thoughts: 
- I'm not really sure what problem you're trying to solve with this PR.
- PRs need to be made against the develop_2.0 branch.
- We will not accept PRs that are not well tested, and not ones that break existing tests. You will have to fix those if you want the PR to be merged.

gdilipkumar:
* **Problem Statement:**
   In our product we are depending only on Swagger-models. While processing models Jackson Object Mapper can't deserialize Polymorphic types like Property, Parameter and Model. For models like Parameter we can use in built Jackson feature (@JsonTypeInfo and @JsonSubTypes) but it had serialization bug (Id property duplicated) and for other models we don't have discriminator to decide. 

   To solve this problem we have two options:
   i. We have to create CustomDeserializer and configure Object mapper to resolve polymorphic sub types. "Swagger-core" using this approach. Problem is additional configuration required wherever we needed. In our case we are having different object mappers for different purposes.
  ii. Annotation + Custom Type Resolver (*My Approach*): With this approach we don't need any configuration. Here i created custom type resolver and configured using Annotations.

* I properly tested my changes and added required junit test cases. 
* Existing test cases failing because it was dependent on custom configuration after adding my changes those custom deserializers won't work, sub type already resolved with Annotation. Hence those test cases can be removed.   

* If your ok with my feature i'll push my changes to *develop_2.0* branch.

Let me know if you found any problems in changes.

webron:
Thank you for the explanation.

However, I still fail to see what you're trying to do here and how you're using the swagger-models project. We have the [swagger-parser](https://github.com/swagger-api/swagger-parser) which does exactly that, deserializes Swagger definitions into the Swagger objects (and you probably want to use that).

Disregarding the deserialization process you added (which I didn't dive into, though noticed you chose to ignore the default value for some reason), some of the test samples are invalid and wrong. Also, the tests don't include any negative use cases. We have yet to decide which Java testing framework we're going to migrate to, so would much rather receive the tests in Scala like the others. Obviously, we cannot have a failing build and will not accept anything that breaks the build. Looking at the some of the failing tests, this is not a matter of deleting the tests but actual output that's wrong.

Furthermore, it appears that there are some deviations from our coding standards as well. This PR is big and would take a while to review, and it looks like some thing in it are wrong. It would have helped if you provided comments (on the commits, specific lines of code, not in code itself) as to why you did what you did. As it stands, I'm not sure when we'll be able to review it.

PR #1195, opened 2015-06-23
Fixes #1153:parsing pathparams with regex has been fixed

### LINKED ISSUE #1153: Jaxrs pathparam with regex generates incorrect swagger json
For example, take this complicated regex which should result in a path like "/api/{itemId}".

    @Path("/{itemId: [0-9]{4}/[0-9]{2}/[0-9]{2}/[0-9]{2}/[0-9]{2}/[0-9]{2}/[0-9]{3}/[A-Za-z0-9]+}")

Instead, the json created ends up interpretting this as a long path:

    "/api/{itemId}/[0-9]{2}/[0-9]{2}/[0-9]{2}/[0-9]{2}/[0-9]{2}/[0-9]{3}/[A-Za-z0-9]+}": { "get": { ...

So, instead of the swagger-ui showing something like: /"api/{itemId}", it ends up showing "/api/{itemId}/a/bunch/of/path/items/that/are/really/itemid/regex"

Hopefully that makes sense. Poking around in the code, I think it's an assumption that the first close brace it finds closes the parameter opening brace, but that's not true when the regex is there:
https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java#L262

lugaru1234:
Fixes #1153:parsing pathparams with regex has been fixed

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -36,6 +36,7 @@ import io.swagger.jaxrs.config.ReaderListener;
 import io.swagger.jaxrs.ext.SwaggerExtension;
 import io.swagger.jaxrs.ext.SwaggerExtensions;
 import io.swagger.jaxrs.utils.ReaderUtils;
+import io.swagger.jaxrs.utils.PathUtils;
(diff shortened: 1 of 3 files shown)


webron:
Can you rebase please?

lugaru1234:
Done

PR #1204, opened 2015-06-26
Fixed #1203: Migrate tests from the Scala to the plain Java

### LINKED ISSUE #1203: Migrate tests from the Scala to the plain Java
Rework all test to the Java with the TestNG framework

- [x] add required dependencies to the pom.xml files
- [ ] swagger-core #1206 
- [x] swagger-jaxrs #1207 
- [x] swagger-jersey2-jaxrs #1208 
- [ ] swagger-hibernate-validations #1209

iushankin:
- Added configuration for the TestNG
- Created sample test

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/BeanValidatorTestNG.java
+++ b/modules/swagger-core/src/test/java/io/swagger/BeanValidatorTestNG.java
@@ -1,0 +1,37 @@
+package io.swagger;
+
+import io.swagger.converter.ModelConverters;
+import io.swagger.models.Model;
+import io.swagger.models.properties.DoubleProperty;
+import io.swagger.models.properties.IntegerProperty;
+import io.swagger.models.properties.Property;
+import io.swagger.models.properties.StringProperty;
+import models.BeanValidationsModel;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.Map;
+
+public class BeanValidatorTestNG extends Assert {
+
+    @Test(testName = "read bean validations")
+    public void readBeanValidatorTest() {
+        final Map<String, Model> schemas = ModelConverters.getInstance().readAll(BeanValidationsModel.class);
+        final Model model = schemas.get("BeanValidationsModel");
+        final Map<String, Property> properties = model.getProperties();
+
+        final IntegerProperty age = (IntegerProperty) properties.get("age");
+        assertEquals(age.getMinimum(), 13.0);
+        assertEquals(age.getMaximum(), 99.0);
+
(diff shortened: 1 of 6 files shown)


modules/swagger-core/src/test/java/io/swagger/BeanValidatorTestNG.java
```diff
@@ -0,0 +1,37 @@
...
+import io.swagger.models.properties.IntegerProperty;
+import io.swagger.models.properties.Property;
+import io.swagger.models.properties.StringProperty;
+import models.BeanValidationsModel;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.Map;
+
+public class BeanValidatorTestNG extends Assert {
```
olensmar: I think just "BeanValidatorTest" is fine... :-)

pom.xml
```diff
@@ -47,7 +47,6 @@
     <build>
         <sourceDirectory>src/main/scala</sourceDirectory>
         <scriptSourceDirectory>src/main/scripts</scriptSourceDirectory>
-        <testSourceDirectory>src/test/scala</testSourceDirectory>
```
olensmar: do we really want to remove the scala tests before all of them are migrated?

iushankin: The ScalaTest compiles tests from the src/test/scala path by default.
We compile tests from both paths: src/test/scala, src/test/java.
Please look at build.log file - https://travis-ci.org/swagger-api/swagger-core/builds/68503343#L3622-L3623

modules/swagger-core/src/test/java/io/swagger/BeanValidatorTestNG.java
iushankin: Yes, but we have some java tests in scala (i.e. SimpleGenerationTest, ModelPropertyOverrideTest, etc.). And these tests will be run two times (first run by the ScalaTest and second run by the TestNG).
It's best workaround when we have two types of tests.

olensmar: ok - we can rename them once the Scala tests have removed... thanks

olensmar:
thank you!

modules/swagger-core/src/test/java/io/swagger/BeanValidatorTestNG.java
tomtit: I don't like the idea with renaming and TestNG ending. There is nothing wrong if some tests will run twice (of course until they will not fail). This just means that these tests are first candidates for migration.
But my main concern here that we introduce implicit naming convention for Java tests for other community members.

PR #1229, opened 2015-07-01
Update RefModel, RefParameter and RefProperty to allow relative refs

Helmsdown:
Let me know what you think. There was some duplicated code between the RefXXX classes and I thought it would be nice to centralize that.

I added tests to swagger-models. Let me know if you are ok with that (there were no tests previously in the module, not sure if that was intentional). 

Normally I would test RefModel, RefParameter and RefProperty using JMockit (to assert that it was correctly delegating to GenericRef). Assuming you are ok with these changes. What are your thoughts on adding JMockit as a dependency to the swagger-models module (assuming it doesnt break the build)?

Please note I added the getRefFormat() to the RefXXX classes because I'm pretty sure that will come in handy inside swagger-parser:SwaggerResolver when we update it to resolve relative paths.


wing328:
@Helmsdown FYI the PR couldn't pass the CI build.

modules/swagger-models/src/main/java/io/swagger/models/refs/RefType.java
```diff
@@ -0,0 +1,9 @@
+package io.swagger.models.refs;
+
+/**
+ * Created by russellb337 on 7/1/15.
+ */
+public enum RefType {
+    DEFINITION,
+    PARAMETER
+}
```
olensmar: To be fully in line with in line with https://github.com/swagger-api/swagger-spec/blob/master/guidelines/REUSE.md - shouldn't we have OPERATION and RESPONSE refs as well (with corresponding changes in the other classes) ?

Helmsdown:
@wing328  The test failure is due to the coverage expectations not being met. These are the first tests for swagger-models module. I thought I had the coverage expectations to configured to allow the current state of coverage to pass (but its still tripping up on the number of uncovered classes). I'll tweak with the coverage expectations for swagger-models module to get it to pass.

@olensmar I absolutely agree. This will require creating some new RefXXX classes and appropriate deserializers (nothing too hard). I would like to suggest a different name for this section, https://github.com/swagger-api/swagger-spec/blob/master/guidelines/REUSE.md#operations. This section is really referring to paths, not operations. 

This is the example given, I would think of this as "ref path"
```json
{
  "paths": { //this parent object is not explicitly stated, but I assuming that is an oversight
    "/health": {
      "$ref": "http://localhost:8000/health.json#/health"
    }
  }
}
```

This is my opinion of what a "ref operation" would be
```json
{
  "paths": {
    "/health": {
      "get": {
        "$ref": "http://my.company.com/operations.json#/get"
      }
    }
  }
}
```

I think both are valid use cases and would like to implement both. If this is OK with you, I'll submit a change to the guidelines page (and perhaps the swagger 2.0 spec wiki page itself) when we get swagger-core changes merged in.

Helmsdown:
@olensmar I have some preliminary changes to support path/operation and response references on another [branch](https://github.com/Helmsdown/swagger-core/tree/path-operation-response-refs). These additional changes are more intrusive. What do you think merging this pull request first, and then reviewing a second pull request?

Helmsdown:
Thanks Tony!!!!!

webron:
@Helmsdown - 1.5.1-SNAPSHOT is now available in sonatype with the PR if you need to test anything.

PR #1235, opened 2015-07-03
Fixed #1234: The Mule module doesn't support BeanConfig

### LINKED ISSUE #1234: The Mule module doesn't support BeanConfig
Configuration from the BeanConfig not apply at the Mule module

https://github.com/swagger-api/swagger-core/blob/develop_2.0/modules/swagger-mule/src/main/java/io/swagger/mule/ApiListingJSON.java#L45-L49

iushankin:


iushankin:
Fix for swagger-api/intuit-support#70

olensmar:
Thanks - this looks fine to me but lets wait for @webron just to be sure we aren't missing something mule or Intuit-specific...

PR #1237, opened 2015-07-06
Fixes #1212: Deprecated annotations are now read from the class hiera…

### LINKED ISSUE #1212: @Deprecated is not inherited from the superclass [swagger-core 1.5.0, jaxrs 2.0]
io.swagger.jaxrs.Reader contains

```java
boolean isDeprecated = false;
annotation = method.getAnnotation(Deprecated.class);
if (annotation != null) {
    isDeprecated = true;
}
```

Instead of using the helper function getAnnotation(method, Deprecated.class), which would correctly search up the class hierarchy.

jacob-meacham:
…rchy.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -19,6 +19,7 @@ package io.swagger.jaxrs;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.type.TypeFactory;
 import com.google.common.collect.Collections2;
+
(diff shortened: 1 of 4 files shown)


modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
```diff
@@ -16,9 +16,40 @@
 
 package io.swagger.jaxrs;
 
+import java.lang.annotation.Annotation;
```
jacob-meacham: Ack, let me fix my import order

webron:
Can you rebase please?

jacob-meacham:
Done!

PR #1241, opened 2015-07-06
Fix Java 8 Javadoc 'error: self-closing element not allowed'

DavidBiesack:
Replace `&lt;p/&gt;` with `&lt;p&gt`; so that `mvn javadoc:javadoc` or `mvn install` runs cleanly with JDK 1.8


webron:
Can you resubmit it against develop_2.0 please? master is stable.

DavidBiesack:
closing to resubmit against  develop_2.0

PR #1245, opened 2015-07-07
Make sure `access` field is populated in BodyParam

colinmccabe:
Make sure the `access` field in an @ApiParam annotation gets copied to BodyParam.


colinmccabe:
This is submitted against `master`, just saw that it should be submitted against `develop_2.0`

Issue #1188, opened 2015-06-22
Add @ApiParam#hidden()


webron:
Was sure we had this one, but it doesn't seem to be the case. We should add `boolean hidden()` to `@ApiParam`, with the default value `false`. If set to `true`, that parameter should be dropped from the parameter list of the operation.

This would have helped #1187 and  #1173 (possibly making it obsolete).

t1:
I'd like to work on a patch for this issue.

who:
+1

t1:
I've created [pull request 1192](https://github.com/swagger-api/swagger-core/pull/1192)

t1:
I've updated [pull#1192](https://github.com/swagger-api/swagger-core/pull/1192) as requested. Can I improve anything else on it?

webron:
Fixed in develop_2.0. Thanks @t1!

who:
Thank you, @t1 !

PR #1246, opened 2015-07-07
Make sure access field is populated when constructing BodyParam obj

colinmccabe:
Make sure the access field in an @ApiParam annotation gets copied to BodyParam.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/ParameterProcessor.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/ParameterProcessor.java
@@ -104,11 +104,18 @@ public class ParameterProcessor {
         } else {
             // must be a body param
             BodyParameter bp = new BodyParameter();
+
             bp.setRequired(param.isRequired());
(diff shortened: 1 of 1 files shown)


webron:
Just curious, have you encountered a related issue?

colinmccabe:
I noticed this when trying to hide parameters with a SpecFilter that looked at the access field, as described here: http://stackoverflow.com/questions/22812365/how-to-hide-a-parameter-in-swagger. The parameter I was trying to hide happened to be BodyParam.

webron:
Fair enough. And one other question, are you trying to hide it completely or really based on some `access` value?

colinmccabe:
I'm trying to remove it completely from the generated swagger.json, with a custom SpecFilter similar to the one in the StackOverflow link above.

My API takes a session param, which Swagger is interpreting as a BodyParam. I'm aiming to hide it declare the session param an ApiImplicitParam instead.

webron:
In that case, be on the lookout for #1188 which would give you the same effect without having to configure a SpecFilter. 

Thanks for the PR!

Issue #1058, opened 2015-05-11
Petstore 1.2 api-docs JSON doesn't validate against the spec


earth2marsh:
Running the [Swagger JSON](http://petstore.swagger.io/api/api-docs) for the example petstore server through the swagger-tools validator turned up two issues:
1. `/pet` has an uploadFile operation with a 'File' type, which [should use paramType of 'form'](https://github.com/swagger-api/swagger-spec/blob/master/schemas/v1.2/parameterObject.json#L21).
2. `/user` has several operations whose scopes are `test:anything`. It might be failing validation because that hasn't been declared in the api-docs resource listing's authorizations definition?

webron:
Need to see how this is going to work with #1066. @fehguy, are we going to keep the 1.2 pet store specifically in the new repo and just modify it there?

fehguy:
OK, looking at the source of this issue, I do believe this is how the code has always been--the `File` datatype is automatically mapped to `body`:

https://github.com/swagger-api/swagger-core/blob/v1.3.12/modules/swagger-jersey-jaxrs/src/main/scala/com/wordnik/swagger/jersey/JerseyApiReader.scala#L178-L182

```scala
case e: FormDataParam => {
  mutable.dataType match {
    case "java.io.InputStream" => {
      mutable.name = readString(e.value, mutable.name)
      mutable.paramType = "body"
      mutable.dataType = "File"
    }
```

Looking at the UI code, it looks like the datatype `file` is also ignoring the paramtype:

https://github.com/swagger-api/swagger-ui/blob/master/src/main/javascript/view/OperationView.js#L218-L222

```js
      if (type && type.toLowerCase() === 'file') {
        if (!contentTypeModel.consumes) {
          contentTypeModel.consumes = 'multipart/form-data';
        }
      }
```

With that background, I suggest we do NOT change the underlying java code and release 1.3.13.  We should simply annotate the spec that the param type is ignored for datatype `File`.  I fear that releasing a new version of 1.3.x would not help.

Chalk this up to underspecification of the spec previously, which is addressed in 2.x.

webron:
I'm also not really looking forward to releasing 1.3.13 just to solve this thing. Would rather not update the spec either though. The issue with the file type is an old one, and I was sure we fixed it, but unfortunately we didn't.

fehguy:
OK I have created two PRs that will address this issue.  Let's decide if we want to merge them.

PR #1259, opened 2015-07-13
Feature/bug 1058

fehguy:
This addresses the petstore sample from #1058


fehguy:
I'm looking at the build error, it's unrelated to the change.

PR #1270, opened 2015-07-15
Fixes swagger-api/swagger-codegen#980 : Allowing inheritance without additonal properties

maxime-morand:
Fixes swagger-api/swagger-codegen#980

In swagger-codegen, a model which inherit from an other without defining its own properties will result in a NullPointerException. The error is caused by the ModelDeserializer which create a CompoundModel without any parent.
This fix would allow such a case.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/util/ModelDeserializer.java
+++ b/modules/swagger-core/src/main/java/io/swagger/util/ModelDeserializer.java
@@ -33,19 +33,20 @@ public class ModelDeserializer extends JsonDeserializer<Model> {
             List<Model> allComponents = model.getAllOf();
             if (allComponents.size() >= 1) {
                 model.setParent(allComponents.get(0));
-                model.setChild(new ModelImpl());
-            }
-            if (allComponents.size() >= 2) {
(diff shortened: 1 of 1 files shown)


webron:
@maxime-morand - is there a respective ticket opened on swagger-codegen?

maxime-morand:
@webron - no, i did not find any

webron:
@maxime-morand - can you please open one and link to the PR? This is for documentation purposes, especially with a cross-project dependency. Should end up in the release notes of the codegen when it's released with the fix.

maxime-morand:
Fixes Swagger/swagger-codegen#980

fehguy:
Looks good, thanks for contributing.

webron:
Heh, I believe @fehguy meant to merge it ;)

maxime-morand:
@webron thanks a lot, but i think you merged my first commit. There is still a NPE because the child is null (that's what my second commit corrected)

webron:
PR merges works on PRs as a whole, not specific commits. If something was left out, it means the PR wasn't created properly. Feel free to submit an additional PR to cover what's missing.

Issue #1258, opened 2015-07-13
SecurityRequirements: Mismatch between model & spec


FossiFoo:
The spec defines "security" as [SecurityRequirement] and those as a Map of names and scopes, so effectively a List<Map<String,List<String>>>. 

Swagger.java however contains a protected List<SecurityRequirement> securityRequirements; with those being just a name and list of scopes, missing one level of the structure.

I would propose to add a "Requirements" object which contains a map of the existing SecurityRequirements.

While this is touched anyway, i also noticed, that the Method to add these is called "addSecurityDefinitions(SecurityRequirements)" instead of "addSecurityRequirements" and a plain "securityRequirements" Helper is missing.

hgwood:
It is mentioned in #1271 that the fix for this requires a breaking change that will only be considered for the next major version. Is there any idea of the when of said major version?

webron:
@hgwood - not yet.

fehguy:
The helper `security` was added (not `securityRequirement` per your request) as it follows the JSON representation of attribute names.  There is als a `securityDefinition` helper for building a swagger with the SecurityDefinition object.  I'm not quite sure what the need is for a `Requirement` object though.

cbornet:
The field is still serialized as "securityRequirement". Do you know if it works with swagger-codegen and swagger-ui ?

cbornet:
Also I think the ```@JsonIgnore``` that was set on setSecurity() should be removed so that swagger-core can deserialize both security and securityRequirement but only serializes as securityRequirement. That would help for transitionning.

boillodmanuel:
Any news on this topic? The generated swagger is not valid due to 'security' attribute that's named 'securityRequirement'. Do you have an ETA for this issue?

I noticed that there were breaking changes in 1.5.8. Could the fix of this issue be included in 1.5.9?

Best regards

fehguy:
@boillodmanuel How are you getting `securityRequirement`?  The SwaggerSerializers for jackson should address this.  Perhaps you're not using them?  They are required.

boillodmanuel:
I'm not. I' e looked at the unit test and they are using ObjectMapper.
Thanks for pointing this out.

fehguy:
Ok, please report back if you're having troubles with this.  You can use the `Json.mapper()` for convenience.

boillodmanuel:
I wrote a simple test:

```java
    @Test
    public void testSwaggerExport() throws Exception {
[...]

Issue #1022, opened 2015-04-29
Add getOperations() method to Path model class that returns a Map<String, Operation>


Helmsdown:
Currently the Path model POJO allows you get access to it's operations in two ways:

1. Get a List<Operations>
2. Call a named function like getPost() and getDelete().

Both of these are useful, but sometimes when you are iterating over the operations for a Path you want all defined operations AND their http method. 

Can we get a function that looks like this:

```java
public Map<String, Operation> getOperations() {
    //if operation is defined put it in the map keyed by its http method name
}
```

lugaru1234:
I would suggest to add enum for http methods and use it as a key instead of strings.
`public Map<HttpMethod, Operation> getOperations()`
What do you think?

webron:
@Helmsdown - Any input regarding @lugaru1234's suggestion?

Helmsdown:
@webron I think it makes perfect sense. I ended up including a proposed change into pull request #1271.

#TwoBirdsOneStone :-)

Issue #1261, opened 2015-07-13
Multiple security requirements with annotations


cbornet:
According to the spec, it should be possible to declare multiple requirements with AND logic on a Map but this doesn't seem to be possible from the annotations (only the OR list is available)

When using the authorizations={@Authorizations...} in an operation, the following code is called
```java
public void addSecurity(String name, List<String> scopes) {
        if (this.security == null) {
            this.security = new ArrayList<Map<String, List<String>>>();
        }
        Map<String, List<String>> req = new HashMap<String, List<String>>();
        if (scopes == null) {
            scopes = new ArrayList<String>();
        }
        req.put(name, scopes);
        this.security.add(req);
    }
```
so the authorization is put in a single element map  which is then added to the list.
Could it be possible to have support for the "AND" logic in the annotations ?

cbornet:
Related to #1258 ?

webron:
Yup, it depends on it.

frantuma:
seems to be resolved by https://github.com/swagger-api/swagger-core/pull/1271

PR #1271, opened 2015-07-16
Extend swagger-models to support RefPath and RefResponse per Swagger Reuse Guidelines

Helmsdown:
The [reuse guideline](https://github.com/swagger-api/swagger-spec/blob/master/guidelines/REUSE.md) mentions two types of references which are not currently supported by swagger-core. 

[This](https://github.com/swagger-api/swagger-spec/blob/master/guidelines/REUSE.md#operations) section makes note of referencing operations. 

The example(s) given for "operation refs" are:
```json
{
  "/health": {
    "$ref": "http://localhost:8000/operations.json#/health"
  }
}
```

```json
{
  "health": {
    "get": {
      "tags": [
        "admin"
      ],
      "summary": "Returns server health information",
      "operationId": "getHealth",
      "produces": [
        "application/json"
      ],
      "parameters": [],
      "responses": {
        "200": {
          "description": "Health information from the server",
          "schema": {
            "$ref": "http://localhost:8000/models.json#/Health"
          }
        }
      }
    }
  }
}
```
When I read this section I think *"shouldn't we call these path references?"*. The first json file example has a REST path of "/health" as the key.

If anything we shouldn't we update the first example to be the following?

```json
{
  "swagger": "2.0"
  "paths": {
    "/health": {
      "$ref": "http://localhost:8000/operations.json#/health"
    }
  }
}
```
To this end, I went ahead and implemented PathRefs (and not OperationRefs), assuming I am correctly interpreting the spirit of the Reuse guide.

Additionally, [this](https://github.com/swagger-api/swagger-spec/blob/master/guidelines/REUSE.md#responses) section of the reuse guideline notes that response definitions should be reusable through references.

To this end I modified the swagger-models classes in question to fit this proposed [design](http://www.gliffy.com/go/publish/8487689). I tried to follow the established example of Model and RefModel.

This change will have non trivial impact on other swagger libraries (and anyone that has written custom code that interacts with the Path and Response objects). 

Specifically the following code will no longer compile:

```java
Path p = new Path();
Response r = new Response();
```
and will have to be changed to:
```java
Path p = new PathImpl();
Response r = new ResponseImpl();
```
 Any code, swagger or otherwise, that processes Path and Response objects will need to be aware there might be RefPath and RefResponse instances lurking about. Having written custom code that iterates over Swagger objects, I am sensitive to the impact these kinds of changes can have. However, I think the long term benefits of being able to compose specs from previously defined Paths and Response (in additional to the already supported ref types) outweighs any short term pain.

Feedback on my code is appreciated.


webron:
Since we're in the process of migrating to TestNG (and migrating all the tests), can you maybe convert the *new* tests to TestNG? It'd just help us with less migration.

Helmsdown:
Absolutely. Shouldn't take me too long.

webron:
Much appreciated. And from your note, once this is merged it resolved #1022 as well, right?

Helmsdown:
Yes indeed.

Helmsdown:
@webron The tests have been updated to Java/TestNG. Let me know if there are additional changes you'd like.

webron:
Thanks for the work, @Helmsdown. I'm going to leave the review to @fehguy on this one though.

Helmsdown:
Sounds good.

fehguy:
As discussed, we should avoid changing the pattern for models until a 2.0 release, since it'll break every usage of the `SwaggerModel` artifact.  While this is much better, I suggest we work around the interfaces and push the renaming into next major.

Helmsdown:
@fehguy Did as we we discussed. 

1. RefPath extends Path (until we can do a major release and make Path an interface)
2. RefResponse extends Response (*)
3. Added vendor extensions to the following swagger-models classes
 * AbstractModel.java
 * AbstractProperty.java
 * Response.java
 * SecurityRequirement.java
 * ExternalDocs.java
 * SecuritySchemeDefinition.java

Please note I had to do some "tomfoolery" in Json.java to support correct deserialization of RefPath/Path and RefResponse/Response. If you know of a better way to achieve the same result I would be very interested. Also, while I was messing around in Json.java I decided to do some light refactoring to central ize code that was duplicated between Json.java and Yaml.java.

As always, feedback on my code is greatly appreciated.

modules/swagger-core/src/main/java/io/swagger/core/filter/SpecFilter.java
```diff
@@ -1,11 +1,7 @@
 package io.swagger.core.filter;
 
 import io.swagger.model.ApiDescription;
-import io.swagger.models.Model;
-import io.swagger.models.Operation;
-import io.swagger.models.Path;
-import io.swagger.models.Swagger;
-import io.swagger.models.Tag;
+import io.swagger.models.*;
```
fehguy: In general, we've moved away from wildcard imports.  Not a blocker to this PR of course, just making sure you've seen the [style guides](https://github.com/swagger-api/swagger-core/wiki/Coding-Standards)

fehguy:
Well this looks really good.  Let's get @webron's input as well as his spec changes first and we can see if this can be merged for the next minor release.

Helmsdown:
@fehguy Auto-organize imports strikes again! Just reverted that particular change.

modules/swagger-core/src/main/java/io/swagger/core/filter/SpecFilter.java
Helmsdown: Auto-organize imports strikes again! Just reverted that particular change.

webron:
After updating the spec, working out the proper places for vendor extensions, here is the full list:

- Swagger (i.e. the root object)
- Info
- Contact
- License
- Paths
- Path Item
- Operation
- External Docs
- Parameter
- Items
- Responses
- Response
- Header
- Tag
- Schema (which includes Property)
- XML
- Security Scheme
- Scopes

All other objects cannot contain vendor extensions. 

I suspect there may be an issue with the security-related objects in the Java implementation. Not sure if it's related to the change here or not, but perhaps should be taken into account. Have a look at #1258 and #1261.

Helmsdown:
@webron Want me to refactor to satisfy #1258 and #1261? Are breaking changes ok? (since it apparently doesnt match the spec)

webron:
Since it's breaking, I'd rather have @fehguy's opinion. It's a bit tricky.

Helmsdown:
@fehguy Let me know what, if anything, you want me to do anything additional regarding #1258 and #1261...I have the time.

webron:
Thanks for all the effort, @Helmsdown

fehguy:
We can't break the models.  We can add to them though.  We can add new accessors or fields, but not break old ones :(

webron:
How would we resolve #1258 and #1261 then?

fehguy:
if we are breaking the old models then we cannot.  If we add new ones, then great.

Helmsdown:
So #1261 sounds like it just needs a logic change. We need to add to this list of scopes...not override it with each successive call to *addSecurity()*...right?

#1258 sounds like it is a break change and we will have to wait till the next major version.

#1261 doesnt *need* to go in this pull request, but I can fix it if you like.

Helmsdown:
You know on second read...I am confused by #1261...what exactly is the problem.

webron:
#1261 shouldn't be part of this PR but requires #1258 to be solved. 
Not resolving #1258 means that there can't be proper security representation for tools that depend on models, and swagger-core can't be used in general to output such definitions.

Helmsdown:
After re-reading #1258 again and looking at SecurityRequirement.java, I understand it now. 

This is what is allowed right now in Java
```java
        Swagger swagger = new Swagger();
        swagger.addSecurityDefinition(
                new SecurityRequirement("oauth")
                        .scope("scope1")
                        .scope("scope2"));
        swagger.addSecurityDefinition(
                new SecurityRequirement("api_key"));
```
and it gets serialized as
```json
{
  "swagger" : "2.0",
  "securityRequirement" : [ {
    "scopes" : [ "scope1", "scope2" ]
  }, { } ]
}
```
There are several issues at play:
1. The spec says this element should be called "security" not "securityRequirement" (can't fix this without a breaking chance)
2. The name of the requirement is not serialized due to a JsonIgnore annotation in SecurityRequirement.java
3. Even if the name of the requirement was serialized it wouldn't match the spec due to the the fact that the spec documents the format to be {"api_key": []} not {"name": "api_key", "scopes": ["scope1", "scope2"]
4. The java function to add a securityRequirement to the root swagger object is called *addSecurityDefintion()*

I have some proposed changes to remedy most of these with non breaking changes (and hopefully get us set up for cleaning this stuff up in the next major version). Will post the diff in a moment.

modules/swagger-models/src/main/java/io/swagger/models/SecurityRequirement.java
```diff
@@ -1,19 +1,26 @@
...
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 public class SecurityRequirement {
     private String name;
     private List<String> scopes;
+//    private final Map<String, Object> vendorExtensions = new HashMap<String, Object>();
```
Helmsdown: can't have two @JsonAnyGetter/@JsonAnySetter maps. Being compliant with the primary purpose of this object seems more important than having vendor extensions in this object at this point.

modules/swagger-models/src/main/java/io/swagger/models/SecurityRequirement.java
```diff
@@ -21,19 +28,40 @@ public SecurityRequirement scope(String scope) {
         return this;
     }
 
+    public SecurityRequirement requirement(String name, List<String> scopes) {
```
Helmsdown: support multiple definition references in this requirement

Helmsdown: support multiple references to security definitions in this requirement

modules/swagger-models/src/main/java/io/swagger/models/Swagger.java
```diff
@@ -19,7 +20,7 @@
     protected List<Scheme> schemes;
     protected List<String> consumes;
     protected List<String> produces;
-    protected List<SecurityRequirement> securityRequirements;
+    protected List<SecurityRequirement> security;
```
Helmsdown: rename the private variable to bring in line with the spec definition

modules/swagger-models/src/main/java/io/swagger/models/Swagger.java
```diff
@@ -115,6 +116,11 @@ public Swagger model(String name, Model model) {
         return this;
     }
 
+    public Swagger security(SecurityRequirement securityRequirement) {
```
Helmsdown: helper requested by #1258

modules/swagger-models/src/main/java/io/swagger/models/Swagger.java
```diff
@@ -267,19 +273,39 @@ public void addSecurityDefinition(String name, SecuritySchemeDefinition security
...
     public void addSecurityDefinition(SecurityRequirement securityRequirement) {
-        if (this.securityRequirements == null) {
-            this.securityRequirements = new ArrayList<SecurityRequirement>();
+        if (this.security == null) {
+            this.security = new ArrayList<SecurityRequirement>();
+        }
+        this.security.add(securityRequirement);
+    }
+
+    @JsonIgnore //remove JsonIgnore when getSecurityRequirement() method is deleted in next major version
```
Helmsdown: don't want to serialize the security requirements twice

Helmsdown:
Ok I think I have things, with respect to #1258, as good as they are going to get without a breaking change. Its not completely compliant with the spec (i.e. The list of security requirements is going to get serialized/deserialized as "securityRequirements" but the spec says it should be "security"). 

I @Deprecated all the methods that are wrong, but we can't delete because it would be breaking. 

We now should be able to model what is described in the spec. However, I *did not* change any of the code in swagger-core that actually acts on the @Authorization annotations (e.g. Reader.java).

@fehguy & @webron take a look at let me know if you want anything additional.

fehguy:
looks good @Helmsdown

Helmsdown:
Danke!

FossiFoo:
Nice. Not breaking the old behaviour is fine with me, as long as we can implement the correct usage in generators with both "security" and "securityRequirements" for now. Then the "securityRequirements" one can be deprecated later in the next major.

PR #1278, opened 2015-07-21
upgrade jackson version

Helmsdown:
version 2.4.2 of jackson-dataformat-yaml has a bug that prevents
successful serialization of an object to yaml (i.e. when you try and serialize an object to YAML it actually gets serialized as JSON).

I upgraded all of swagger core's Jackson libraries to 2.4.5 to pull in the fixed version. 

I also added a unit test that tests YAML is getting serialized correctly.

### DIFF
--- a/modules/swagger-core/pom.xml
+++ b/modules/swagger-core/pom.xml
@@ -148,6 +148,12 @@
             <artifactId>testng</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>${commons-io-version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <properties>
         <!-- TODO increase coverage -->
--- a/modules/swagger-core/src/test/java/io/swagger/util/YamlTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/util/YamlTest.java
@@ -1,0 +1,31 @@
+package io.swagger.util;
+
+import io.swagger.models.*;
+import io.swagger.models.properties.BooleanProperty;
+import io.swagger.models.properties.RefProperty;
+import org.apache.commons.io.IOUtils;
+import org.testng.annotations.Test;
+
+import java.io.InputStream;
+
+import static org.testng.Assert.assertEquals;
+
+
+public class YamlTest {
+
+    @Test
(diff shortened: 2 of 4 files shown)


webron:
@Helmsdown - I'm a bit surprised by this as we pretty much produce YAML successfully already, so I don't follow the use case.

Also, if we're to upgrade it, it's better to upgrade all related Jackson versions (such as jackson-guava-version).

Helmsdown:
@webron -Try taking the content of the unit test I wrote and running it in a main(), do you get YAML?. I totally missed the jackson-guava dependency...will fix that momentarily.

webron:
I'll try to do it a bit later (in the middle of a few things) but I'm basing my knowledge on http://petstore.swagger.io/v2/swagger.yaml which looks pretty YAML to me ;) However, perhaps you're talking about a different use case, which is what I'm trying to understand.

Helmsdown:
Ah so this is interesting.

With 2.4.2 of the Jackson YAML library...

```java
//this produces yaml
Yaml.mapper().writeValueAsString(foo);
//this produces json
Yaml.pretty().writeValueAsString(foo);
```

With verison 2.4.5 of the library...
```java
//this produces yaml
Yaml.mapper().writeValueAsString(foo);
//this produces yaml as well now
Yaml.pretty().writeValueAsString(foo);
```

I've gotten in the habit of calling pretty() when I am marshaling something that a human is going to look at. So there is some bug related to the yaml library when you use the DefaultPrettyPrinter class in conjunction with the YAMLFactory, but you can get around it by just not using the pretty() function in YAML. WDYT, upgrade or keep it as is?

webron:
I'm inclined to keep it as-is for now. There *may* be a fix in Jackson 2.6 that would be helpful and we'd upgrade to it, but we've had issues with some of our users (you may know a few of them) even with Jackson 2.4.2 (though that's the bare minimum we have to use).

On the other hand, 2.4.5 is a minor bump so it may not be an issue. If you can push the guava update as well, that'd be good. Give us a couple of days and we may merge it still (if only to solve the pretty() bug).

Thanks again for all the work!

Helmsdown:
@webron Done.

webron:
:+1:

webron:
@Helmsdown - we actually upgraded to 2.4.5 via a different PR, but thanks :)

Helmsdown:
Great. All's well that ends well.

PR #1297, opened 2015-07-28
Allow VendorExtensions to process @Context params

mwinkels:
Currently, parameters that have @Context annotation are ignored by the Reader. This is fine for automatically processed parameters, but with Vendor Extensions, it would be nice to be able to extend the model with additional parameter information for these parameters.


webron:
This PR broke the 1.5.1 release and forces us to release 1.5.2 to fix it. 

We can easily add the necessary check (parameter *can* be null, so it throws NPE). However, the whole use case for this PR is unclear, and should be furthered explained. If you want this to make in, make sure you provide us with more information, and add additional tests to the code to make sure the edge cases are covered.

Issue #1202, opened 2015-06-26
swagger-models - ObjectProperty children are lost - no way to parse nested object definitions that are not $refs


benken-parasoft:
I am using swagger-models to marshall Swagger back and forth.

When consuming a Swagger having a JSON schema object definition where one of the properties is another object definition, all properties in the nested object are lost.  The nested object definition becomes an ObjectProperty but there is no method to get the children.  Marshalling the ModelImpl back to JSON (round trip) shows the properties of the nested object being lost.

There appears to be no way to use swagger-models to parse the nested object definition!  This causes a problem for any application that wishes to accurately consume Swagger documents and their embedded JSON Schemas.

Below is a JUnit that demonstrates the problem with ObjectProperty:

<pre><code>import static org.junit.Assert.*;

import com.fasterxml.jackson.core.*;
import com.fasterxml.jackson.databind.*;

import io.swagger.models.*;
import io.swagger.models.properties.*;
import io.swagger.util.*;

import java.io.*;
import java.util.*;

import org.junit.*;

public class SwaggerTest {
    @Test
    public void testMinMaxItems() throws JsonProcessingException, IOException {
        ObjectMapper mapper = Yaml.mapper();
        JsonNode rootNode = mapper.readTree(SAMPLE_SWAGGER);
        assertNotNull(rootNode);
        Swagger doc = mapper.convertValue(rootNode, Swagger.class);
        Map<String, Model> definitions = doc.getDefinitions();
        assertEquals(3, definitions.size());

        // What happened to children of submodel?!?
        Model model = definitions.get("Model");
        assertEquals(ModelImpl.class, model.getClass());
        ModelImpl modelImpl = (ModelImpl) model;
        Map<String, Property> properties = modelImpl.getProperties();
        assertEquals(2, properties.size());
        Property property = properties.get("simple1");
        assertEquals(IntegerProperty.class, property.getClass());
        property = properties.get("submodel");
        assertEquals(ObjectProperty.class, property.getClass());
        //ObjectProperty objectProperty = (ObjectProperty) property;
        //objectProperty.getProperties()
        assertEquals("{\"type\":\"object\",\"required\":[\"simple1\",\"submodel\"],\"properties\":{\"simple1\":{\"type\":\"integer\",\"format\":\"int32\"},\"submodel\":{\"type\":\"object\"}}}",
                mapper.valueToTree(model).toString());
        System.out.println(SAMPLE_SWAGGER);
    }

[...]

PR #1314, opened 2015-07-31
Adding children to ObjectProperty

bchazalet:
This is an attempt to add children to the ObjectProperty so that the Property tree can be fully traversed.


modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
```diff
@@ -88,6 +101,19 @@ Property propertyFromNode(JsonNode node) {
                 if (items != null) {
                     return new MapProperty(items).description(description);
                 }
+            } else {
+              detailNode = node.get("properties");
+              Map<String, Property> properties = new HashMap<String, Property>();
+              if(detailNode != null){
+                  for(Iterator<Map.Entry<String,JsonNode>> iter = detailNode.fields(); iter.hasNext();){
+                      Map.Entry<String,JsonNode> field = iter.next();
+                      if(!isReserved(field.getKey())){
```
bchazalet: I am not sure it is necessary to check for the reserved property names here, some input would be welcome

bchazalet:
for #1202

PR #1335, opened 2015-08-06
Develop 2.0 - support binary input and output 

boazsapir:
This replaces #1166 with a new implementation based on the conclusions of the discussion in https://github.com/swagger-api/swagger-spec/issues/50

These changes in swagger-model are a precondition to the changes that are needed in swagger-codegen in order to support binary input and output. Once this request is merged, it will be followed by another pull request to swagger-codegen.


BigSocial:
I guess this solves my issue swagger-api/swagger-core#1531 assuming core will map the InputStream parameter (see my example) to `"type" : "string", "format" : "byte"` and codegen and UI understand the same thing.

I'd make a couple points:

1. `"type" : "string", "format" : "byte"` is a confusing way to encode this. `"type" : "binary"` seems much more clear but I guess there are reasons...
2. @webron asked in swagger-api/swagger-spec#50 "Do you think that, in the case of "format" : "binary", the content type "octet-stream" should be mandatory?" I believe the answer is very much "no" since there are any number of MIME types that one might want to deal with this way e.g. all the image types, `application/pdf`,  `application/msword`, etc.
3. Every popular object storage REST API including S3, OpenStack, Box, DropBox, etc. require PUT or POST of binary data with no encoding to upload objects. I point this out to make it clear that this is not a fringe use case for anyone dealing with binary files such as images, documents, presentations, etc.

PR #1342, opened 2015-08-10
added file as a primitive type

agusgr:
Related info to this change in https://github.com/swagger-api/swagger-core/issues/1319

PR #1363, opened 2015-08-14
Fixes #1319: Added file as a primitive type

### LINKED ISSUE #1319: Error log when dataType is "file" in @ApiImplicitParam (swagger-jaxrs)
Hi all,

Working with version 1.5.1, I get an error trace[1] when I set "file" as the dataType in an ApiImplicitParam annotation.

Seems like the problem is in the method "typeFromString" in https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/utils/ReflectionUtils.java

Maybe "file" should be in PrimitiveType ?

Changing dataType to java.io.File the error dissapeared but the file input also dissapeared from the UI and I cannot perform multipart post.

[1] ERROR [io.swagger.jaxrs.utils.ReflectionUtils] Failed to resolve 'file' into class: java.lang.ClassNotFoundException: file from ...

agusgr:
Related info to this change in #1319 (see also #1342)


fehguy:
This could potentially create other issues and therefore we need to test the impact first.  Targeting 1.5.5

agusgr:
Fixed in #1519

PR #1393, opened 2015-08-24
allow relative refs that start with forward slash

Helmsdown:
A small update to the the logic that determines what kind the "type" of a ref is. 

We want a ref that starts with forward slash to be considered a "relative" ref because its not an internal ref nor is it a URL ref. 

Classpath refs are almost certainly going to start with a "/" to so you can concretely specify a path to to your file from the root of the classpath. For example, `/io/swagger/petstore/pets.json` would mean there is a petstore.json file in a jar on the classpath in the `io.swagger.petstore` package. 

In retrospect, the "third" type of ref should have been named "FILE" instead of "RELATIVE" because references that start with "/" are not really relative.


Helmsdown:
@fehguy / @webron  I would like to start work this week that requires this change. Do you guys have bandwidth to take a look at it?

fehguy:
@Helmsdown sorry for the delay.  Is there any sort of test that we can use to show what the effect of this is?  I'm just not sure how it'll affect things.

Helmsdown:
@fehguy, I just add some additional test cases to GenericRefTest to cover this.

This change should not affect anything negatively.

fehguy:
thank you @Helmsdown

PR #1396, opened 2015-08-25
added some useful log messages and added toString methods

F481:


fehguy:
Hi, I appreciate this PR but want to keep dependencies out of the models artifact.

PR #1397, opened 2015-08-25
multiple swaggers on the same tomcat

crazySocket:
Swagger has somewhat global presence in tomcat environment. If you want to partition your api, global nature hiders you. This commit aims to alleviate the issue.

Change description:
ScannerFactory does not act in accord with its name since this factory does not produce new instances. The class is used as the holder for Scanner instance. As I understand, there exist only one instance of Scanner by design so I find accurate embedding this fact in classes' name, hence ScannerSingleton.
For each part of my api I am gonna need proper SwaggerConfig. For instance to get proper baseUrl. Each instance of SwaggerConfig has to be easily identified in global scope. I could use ServletContext but I loose controll over object type and it would not be consistent with the way global presence of Scanner is done. Added SwaggerConfigMap.
It seems that ServletContext was used to store "reader" (that is obviously instance of SwaggerConfig) and "scanner". When "reader" is used, "scanner" seems to be abbandoned in favor of ScannerFactory. Moreover the "reader" actually sets global scanner. As a result "scanner" has no merit.
Since ScannerFactory is now ScannerSingleton I moved creation of new instance of Scanner from WebXMLReader to ScannerSingleton. BeanConfig replace this instance as it have done.
JerseyJaxrsConfig is present only due to compatibility (I guess) so I modified the class to reflect that.
ApiListingResource stores instances of Swagger using swagger.id what allows partitioning. When swagger.id is absent Swagger is stored under name "swagger-default" as a result previous functionality is supported.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/config/ScannerSingleton.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/config/ScannerSingleton.java
@@ -1,7 +1,9 @@
 package io.swagger.config;
 
-public class ScannerFactory {
-    private static Scanner SCANNER;
+import io.swagger.jaxrs.config.DefaultJaxrsScanner;
+
+public class ScannerSingleton {
+    private static Scanner SCANNER = new DefaultJaxrsScanner();
 
     public static Scanner getScanner() {
         return SCANNER;
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/config/SwaggerConfigMap.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/config/SwaggerConfigMap.java
@@ -1,0 +1,16 @@
+package io.swagger.config;
+
+import java.util.Map;
+import java.util.HashMap;
+
+public class SwaggerConfigMap {
+    private static Map<String, SwaggerConfig> map = new HashMap<String, SwaggerConfig>();
+
+    public static SwaggerConfig getConfig(String id) {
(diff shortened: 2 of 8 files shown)


danigonzalez:
Hi, I've tried and it works, but with this we can only configure swagger instances using init parameters (so web.xml is the only way). Isn't it?

I would like to be able to configure init-params externally (basepath for example) to be able to deploy my war to any of my servers without needing to change web.xml code.

I think it's not possible to override web.xml init-params values using variable substitution defined in other files (context.xml for example). I think it's not possible to override web.xml init-params extending the servlet class to override it in ServletConfig.

It's BeanConfig the only way to configure it programmatically (so in java we can retrieve environment variables defined in context.xml or wherever we want)?? Can I help you making BeanConfig multi instance?

crazySocket:
Sorry for late answer but I had urgent assignments.

You can always make your custom implementation of ServletConfig and make similar servlet to DefaultJaxrsConfig. If you want to use BeanConfig you have to remove last line from setScan(boolean) (setting the scanner). You would need to make some modifications in ApiListingResource.scan() too (choosing scanner).

I remember I found changes necessary to make BeanConfig multiinstance extensive (some time ago). But since swagger team is not interested in merging these changes, maybe no code change is too extensive.

Try to modify BeanConfig as suggested above. I will look into it anyway.

fehguy:
fixed by https://github.com/swagger-api/swagger-core/commit/8031d514f2487e26f73b080a339c993b1964ec61

Issue #1133, opened 2015-06-05
Swagger object missing some features


jahlborn:
You can define common responses at the top-level of a swagger document (for reference in the various operations).  The Swagger object should support this.  Also, i believe the security section is currently missing.

webron:
I assume you're looking for it as a general Swagger models support and not specific for JAX-RS generation?

jahlborn:
yes, we generate our model info on the fly from internal product configuration.

webron:
Fair enough.

fehguy:
This was added in https://github.com/swagger-api/swagger-core/commit/5928abd2668448d29bf29ce6147d83ec1200eb59

Issue #993, opened 2015-04-20
Custom Formats not parsable?


RobWin:
Hello, I use the swagger-parser in https://github.com/RobWin/swagger2markup.

I have added an emailAddr with format "email" to the Swagger Petstore example:

```
"Order": {
            "properties": {
                "id": {
                    "type": "integer",
                    "format": "int64"
                },
                "petId": {
                    "type": "integer",
                    "format": "int64"
                },
                "quantity": {
                    "type": "integer",
                    "format": "int32"
                },
                "shipDate": {
                    "type": "string",
                    "format": "date-time"
                },
                "status": {
                    "type": "string",
                    "description": "Order Status"
                },
                "complete": {
                    "type": "boolean"
                },
                "emailAddr": {
                    "type": "string",
                    "format": "email"
                }
            }
```

And it seems that the Swagger parser (and core) do not add the emailAddr to the Swagger model:

![image](https://cloud.githubusercontent.com/assets/5815046/7175767/81eefd7c-e418-11e4-936b-31405225e30a.png)

If I have a Model like:
```
"Order":{
 "properties": {
                "emailAddr": {
                    "type": "string",
                    "format": "email"
                }
            }
}
```

The properties Map is null:

![image](https://cloud.githubusercontent.com/assets/5815046/7175794/cf5db68e-e418-11e4-9c1c-315040ac8a35.png)

The JsonNode in the parser is correct:

![image](https://cloud.githubusercontent.com/assets/5815046/7176242/e2cbc25c-e41c-11e4-99ef-40cbb15d22f3.png)

The PropertyDeserializer can see the field correctly

![image](https://cloud.githubusercontent.com/assets/5815046/7177789/cc81ce94-e429-11e4-9a13-7df0038e61ed.png)

The problem is in the PropertyBuilder and there in StringProperty.isType

```
  //TODO: implement additional formats
  public static boolean isType(String type, String format) {
    if("string".equals(type) && (format == null || "uri".equals(format) || "byte".equals(format)))
      return true;
    else return false;
  }
```

There should be some CustomProperty which is used then, if it's a custom format.

Thanks in advance.
Robert

jasonh-n-austin:
+1 confirmed this issue today. This seems to violate spec (certainly in the perspective of JSON Schema), as parsers should ignore unknown custom formats:
https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#data-types
[...]

PR #1400, opened 2015-08-26
Feature/993 fall back to no format if format is missing

ePaul:
This is a possible solution for Issue #993.

It is a simple fall-back solution: In `PropertyBuilder.Processor.fromType`, we first try the given (type, format) pair. If this doesn't give a match, we have a custom format without special implementation support, so we simply ignore the format name and look again for Processors for (type, null).

Only if this doesn't give a result we return null.

The first commit also adds some tests for this problem (which fail without the fix in the second commit).

This seems to also solve [Issue #1133 from swagger-codegen](https://github.com/swagger-api/swagger-codegen/issues/1133) (after using a fixed version of swagger-models in swagger-codegen).

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/properties/PropertyBuilder.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/properties/PropertyBuilder.java
@@ -14,7 +14,7 @@ import java.util.List;
 import java.util.Map;
 
 public class PropertyBuilder {
-    static Logger LOGGER = LoggerFactory.getLogger(PropertyBuilder.class);
+    static final Logger LOGGER = LoggerFactory.getLogger(PropertyBuilder.class);
 
     /**
      * Creates new property on the passed arguments.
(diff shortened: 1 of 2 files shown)


tomtit:
I would suggest to remove format checks here at all:
https://github.com/ePaul/swagger-core/blob/feature/993-fall-back-to-no-format-if-format-is-missing/modules/swagger-models/src/main/java/io/swagger/models/properties/StringProperty.java#L54
and place this item https://github.com/ePaul/swagger-core/blob/feature/993-fall-back-to-no-format-if-format-is-missing/modules/swagger-models/src/main/java/io/swagger/models/properties/PropertyBuilder.java#L146 after all other items with the `string` format (i.e. `DATE`, `DATE_TIME`, `E_MAIL` and `UUID`).

ePaul:
@tomtit generalizing your proposal to the other types (the issue also affects number, integer, and possibly array + object), the idea would be to have the isType method for the base-type's processor ignore the format, and then sort the Processor enum items in a way that those ignoring processors come last?

tomtit:
I'm not sure if it's correct to allow any formats for `number`s and `integer`s.

webron:
@tomtit - custom formats are valid no matter what type is used.

tomtit:
Well, `INTEGER` and 'DECIMAL' are already located after all format sensitive builders, so they can be modified to accept any format.

ePaul:
I added a second pull request #1409 with the alternative implementation. Feel free to decline this one.

Issue #1388, opened 2015-08-23
RESPONSE refs can not be internal references. Why?


ChrisTitos:
Hi,

I'm having troubles using the latest swagger-codegen. I get this error:
```
RESPONSE refs can not be internal references 
(through reference chain: io.swagger.models.RefResponse["$ref"]) 
(through reference chain: io.swagger.models.Path["get"]->io.swagger.models.Operation["responses"]->java.util.LinkedHashMap["404"]) 
(through reference chain: io.swagger.models.Swagger["paths"]->java.util.LinkedHashMap["/companies/{id}"])
```

This exception is thrown by swagger-core here: 
https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/models/refs/GenericRef.java#L37

In the comment it says "PATH AND RESPONSE refs  can only be URL or RELATIVE", but my question is, where does it say this in the spec? The [Responses object](http://swagger.io/specification/#responsesObject) doesn't mention any limitations to a reference object.

I might be missing something, but this exception confused me (especially because the online codegen doesn't complain)

webron:
When you say latest swagger-codegen, do you mean the master branch or develop_2.0?

ChrisTitos:
At first I thought it was both master and develop, but after doing a clean first, now it only happens in develop_2.0.
I also get the exception when using swagger-parser v1.0.10

ChrisTitos:
The develop_2.0 branch has been merged into codegen master, so now I also get the exception when using the master branch.
But like I said, this exception shouldn't even be there right? The spec says nothing about what kind of reference is allowed or not (or am I misinterpreting something?).

ePaul:
This error message seems to have been introduced in #1271 (commit 59e6b10).
I didn't find any justification for this in the discussion of the Pull request, nor in the commit comment (which doesn't say much at all).

ePaul:
Related to this, there seems to be no `responses` property of the Swagger object. Maybe this was the reason?

fehguy:
this has been added and will be available in 1.5.4

PR #1405, opened 2015-08-27
Fix Issue #1388: allow path and response refs, too.

ePaul:
Assuming the discussion in #1388 doesn't result in there being a valid reason not to allow PATH and RESPONSE as internal refs, this is a solution which simply allows those.

This commit does those things:

* move the internal prefixes into the RefType enum (→ RefConstants class is not needed anymore)
* add prefixes for RESPONSE and PATH
* get rid of the exception when trying to use an internal reference of type RESPONSE or PATH
* add test cases in GenericRefTest for PATH and RESPONSE.

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/RefModel.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/RefModel.java
@@ -3,7 +3,6 @@ package io.swagger.models;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import io.swagger.models.properties.Property;
 import io.swagger.models.refs.GenericRef;
-import io.swagger.models.refs.RefConstants;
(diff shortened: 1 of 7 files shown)


ChrisTitos:
Any estimate on when this will be merged? :)

webron:
It would probably take a few more days.

ePaul:
@ChrisTitos @webron This needs a corresponding change in swagger-parser (at least in ResolverCache, I'll see if somewhere else too), it seems. I'm currently seeing if I can do this, too.

ePaul:
@ChrisTitos @webron Parsing response-$refs doesn't work currently, because there is no `responses` field in the swagger object – see #1436. I might do a new pull request for this next week, if noone arguments that this omission was intentional.

Issue #1361, opened 2015-08-13
Make 1.5.x swagger-core and swagger-jarxs OSGi-enabled


elakito:
As discussed in this following thread, I am attaching a PR to this issue ticket shortly.
https://groups.google.com/forum/#!topic/swagger-swaggersocket/6w-xuHmTzvQ

webron:
Where's the PR? ;)

elakito:
Just wanted to create the issue number to prepare for a commit ;-)
I wanted to think about one thing before submitting a PR. That is about the lower range for the guava dep.

Because of the current situation between SMX's org.reflections-0.9.9 bundle and guava-18.0 bundle not playing each other on OSGi (the problem that I mentioned in the thread), I am thinking of building swagger-core and swagger-jaxrs with guava 18.0 but explicitly setting the allowed lower range of guava down to 15.0 so that when using the out-of-the box org.reflections-0.9.9 bundle, everyone can wire to guava 15.0. This can be switched back to 18.0 once a newer version of org.reflection bundle becomes available.

Alternatively, if we keep guava's lower range at 18.0, we will need a version of org.reflections-0.9.9 that can be wired to guava-18.0, This can be either done dynamically by using karaf's special wrap installer to set/overwrite its imports manually or by physically creating a new version of org.reflections-0.9.9 bundle. I'll ping the SMX guy tomorrow, as I haven't got any reply to my message posted on their ML.

How do you think? We can decide it by tomorrow.

thanks.
regards, aki

elakito:
Hi Ron,
The SMX guys have just patched their org.reflections bundles. So, the next release versions 0.9.9_2 and 0.9.10_2 will work with guava-18.0. So, we can keep the lower range of swagger-core/jaxrs to the current default 18.0.
regards, aki

webron:
Cool, thanks for the update.

elakito:
Hi Ron,
could you merge the above PR?
So that we can use the OSGI-enabled version from the snapshot repo?
thanks.
regards, aki

webron:
Done, give it a try.

elakito:
Hi Ron,
great. It's working nicely. ;-)
I added a working demo sample in cxf's samples distro
https://github.com/apache/cxf/tree/master/distribution/src/main/release/samples/jax_rs/description_swagger2_osgi

Before closing this ticket, I have a questions (my PR mentioned this briefly).

[...]

PR #1406, opened 2015-08-27
 #1361 fix the pacakging issue caused by PR #1362

elakito:
fix the issue of wrongly including io.swagger.models in swagger-core.jar.


webron:
@elakito - I'm afraid there's another problem. `io.swagger.config` exists in both swagger-core and swagger-jaxrs, and the classes are repackaged in the jars.

elakito:
@webron 
okay. i'll look into it.

PR #1407, opened 2015-08-27
'expected' and 'actual' wrong way around

mattburns:
The syntax for the test method is:
`assertEquals(expected, actual);`

Swapping them around make no difference to whether the test will pass of fail, it just makes understanding it clearer if it ever fails. Eg, it will say something like:
    `Expected "error" but was "string"`
when it should have been:
    `Expected "string" but was "error"`


webron:
Thanks!

webron:
@mattburns - maybe I was too quick on the trigger. Looks like it's the other way around - http://testng.org/javadoc/org/testng/Assert.html#assertEquals%28boolean,%20boolean%29

mattburns:
Ahh, my mistake. I didn't notice you were using TestNG and I didn't know the order was different from JUnit. Sorry for the time wasted...

webron:
No problem, thanks for looking out.

PR #1409, opened 2015-08-27
Feature/993 remove format checks where null

ePaul:
An alternative implementation to #1400 for issue #993. It makes the non-specific Properties' isType methods not check the format anymore, and makes sure they come in the PropertyBuilder.Processor enum after their more specific colleagues.

This pull request contains the same test case commits as #1400.

### DIFF
--- a/modules/swagger-models/src/test/java/io/swagger/models/properties/PropertyBuilderTest.java
+++ b/modules/swagger-models/src/test/java/io/swagger/models/properties/PropertyBuilderTest.java
@@ -1,59 +1,138 @@
 package io.swagger.models.properties;
 
+import java.util.ArrayList;
+import java.util.EnumMap;
+import java.util.Iterator;
+import java.util.List;
+
 import org.testng.Assert;
 
+import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+
+import io.swagger.models.properties.PropertyBuilder.PropertyId;
+import io.swagger.models.properties.StringProperty.Format;
 
 public class PropertyBuilderTest {
 
-    @Test
-    public void testStringPredefinedFormats() {
-        for (final StringProperty.Format format : StringProperty.Format.values()) {
-            final String name = format.getName();
-            final Property property = PropertyBuilder.build(StringProperty.TYPE, name, null);
-            Assert.assertEquals(property.getType(), StringProperty.TYPE);
-            Assert.assertEquals(property.getFormat(), name);
+    private static final String STRING_FORMATS = "stringFormats";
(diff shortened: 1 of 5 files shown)


modules/swagger-models/src/test/java/io/swagger/models/properties/PropertyBuilderTest.java
```diff
@@ -1,26 +1,59 @@
...
+        Assert.assertEquals(customFormatProperty.getFormat(), customFormatName);
+    }
+
+    @Test
+    public void testNumberFloat() {
+        final String floatFormat = "float";
+        final Property floatProperty = PropertyBuilder.build(DecimalProperty.TYPE, floatFormat, null);
+        Assert.assertEquals(floatProperty.getType(), DecimalProperty.TYPE);
+        Assert.assertEquals(floatProperty.getFormat(), floatFormat);
+    }
```
tomtit: @ePaul what do you think on this test?
```
for (Property numeric : Arrays.<Property>asList(new DecimalProperty(), new FloatProperty(), new DoubleProperty(), new BaseIntegerProperty(), new IntegerProperty(), new LongProperty())) {
    final Property built = PropertyBuilder.build(numeric.getType(), numeric.getFormat(), null);
    Assert.assertNotNull(built);
    Assert.assertEquals(built.getClass(), numeric.getClass());
    Assert.assertEquals(built.getType(), numeric.getType());
    Assert.assertEquals(built.getFormat(), numeric.getFormat());
}
```
It will ensure that the `PropertyBuilder` handles all predefined numeric formats properly.
The same test can be added for string properties: `StringProperty`, `DateProperty`, `DateTimeProperty`, `EmailProperty` and `UUIDProperty`.

ePaul: Hmm, this test looks a bit implementation-specific: we are actually just testing that for each property type `‹PropertyClass›.isType(property.getType(), property.getFormat())` returns true (and for the other ones this is either not checked or returns false).

I'm trying with my test to come more from a specification point of view.

Though I guess I can add your test, too. (The `Arrays.asList` is not needed here, we can also directly iterate over that array.)

ePaul: I looked into parameterized Tests with TestNG, and implemented both as parameterized test. Now we have 40 tests (by 4 test methods).

I'll see if I can also cover some corner cases to increase the coverage (currently 67% for PropertyBuilder).

ePaul:
I rebased this again to fix any merge conflicts. Is there still anything to do?

Issue #1277, opened 2015-07-20
ComposedModel without implementation of clone method


rojeda-desp-lp:
What is the cause? Is it a bug?

webron:
What?

rojeda-desp-lp:
Sorry for my poor description.

I want use the clone method of all Swagger Models that implement "Model" interface. But i saw that the object "ComposedModel" doesn't implement this method. 
The "AbstractModel" returns null in this case. 

For this reason, I want know if there is a bug.

I am using the clone method for merging of more than one swagger document.

webron:
Fair enough. Feel free to submit a PR for it.

despegargcampelo:
Looks more like a bug than a feature...
+1

arosenber:
+1 and submitting this solution as it removes a bug that's impairing me.

webron:
Solution merged in master, thanks @arosenber.

PR #1419, opened 2015-09-01
Adding in a clone method.

arosenber:
This references issue 1277 https://github.com/swagger-api/swagger-core/issues/1277. 

Need this to fix an issue with the use of added in SwaggerSpecFilter caused a NullPointerException when using jackson sub types (https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java#L494).


arosenber:
Note that this "clone" isn't really cloning as ModelImpl and ArrayModel do not really clone...(their properties are not cloned immutably...couldn't say the consequences in terms of usage)

PR #1442, opened 2015-09-15
Added support for inheriting parameter annotations on overriden methods.

briandelaney:
One of the things I've run into using Jersey2 is that if I have a method with parameter annotations that is overriden, Swagger ends up giving the parameter a name of "body". This happens even if I also add a meaningful `@ApiParam` annotation to the parameter on the overriding method. You can see this by the following code I added in `ReaderTest.scanOverridenMethod` that fails without these changes:

```
        Swagger swagger = getSwagger(DescendantResource.class);
        Operation overriddenMethodWithTypedParam = getGet(swagger, "/pet/{petId1}");
        assertNotNull(overriddenMethodWithTypedParam);
        Parameter firstParameter = overriddenMethodWithTypedParam.getParameters().get(0);
        assertEquals(firstParameter.getDescription(), "ID of pet to return child");
        assertEquals(firstParameter.getName(), "petId1");
```

The implementation here is similar to the way `Reader.getAnnotation(Method,...` handles looking at overriden methods for method annotations, but in this case, it merges parameter annotations per parameter. It also allows parameter annotations to be overridden by preferring annotations on derived classes, when annotations have the same type.

I also fixed a problem I ran into that I commented on here: https://github.com/swagger-api/swagger-core/commit/818d23f

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -61,6 +61,7 @@ import io.swagger.models.properties.ArrayProperty;
 import io.swagger.models.properties.MapProperty;
 import io.swagger.models.properties.Property;
 import io.swagger.models.properties.RefProperty;
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -832,7 +833,8 @@ public class Reader {
         }
 
         Type[] genericParameterTypes = method.getGenericParameterTypes();
-        Annotation[][] paramAnnotations = method.getParameterAnnotations();
+        Annotation[][] paramAnnotations = getMergedParameterAnnotations(method);
+
         for (int i = 0; i < genericParameterTypes.length; i++) {
             final Type type = TypeFactory.defaultInstance().constructType(genericParameterTypes[i], cls);
             List<Parameter> parameters = getParameters(type, Arrays.asList(paramAnnotations[i]));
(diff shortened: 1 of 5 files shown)


webron:
@frantuma - can you review this please? If it makes sense, we can ask @briandelaney to rebase and we'll merge.

frantuma:
@webron @briandelaney 

I believe the direction is ok, but it doesn't address fully the issue, and also with current code it introduces a wrong behaviour when 2 contrasting JAX-RS param annotations are set on sub and super.

A possible idea is to better adhere to JAX-RS spec, implementing the same behaviour defined in the spec, i.e. inherit JAX-RS annotations if no JAX-RS annotation is defined in the child.

It would mean enhancing @briandelaney code handling JAX-RS annotations according to spec

Please let me know what you think

> JAX-RS annotations MAY be used on the methods and method parameters of a super-class or an implemented interface. Such annotations are inherited by a corresponding sub-class or implementation class method provided that method and its parameters do not have any JAX-RS annotations of its own. Annotations on a super-class take precedence over those on an implemented interface. If a subclass or implementation method has any JAX-RS annotations then all of the annotations on the super class or interface method are ignored. E.g.:

```
public interface ReadOnlyAtomFeed {
    @GET @Produces("application/atom+xml")
    Feed getFeed();
}

@Path("feed")
public class ActivityLog implements ReadOnlyAtomFeed {
    public Feed getFeed() {...}
}

```

> In the above, ActivityLog.getFeed inherits the @GET and @Produces annotations from the interface. Conversely:

```
@Path("feed")
public class ActivityLog implements ReadOnlyAtomFeed {
    @Produces("application/atom+xml")
    public Feed getFeed() {...}
}

```

> In the above, the @GET annotation on ReadOnlyAtomFeed.getFeed is not inherited by ActivityLog
> .getFeed

fehguy:
This PR has drifted a bunch from the codebase and I believe we've moved in a slightly different direction

PR #1449, opened 2015-09-22
update felix-plugin to generate the exported version numbers correctly

elakito:
Hi Ron,
it took a while because it confused me that the maven build of swagger was not automatically adding the exported version as in the other projects that I had. Then I noticed that swagger was using maven-bundle-plugin 2.3.4. whereas the other projects that I had were using a newer version either 2.3.7, 2.4.0, or 2.5.4 and they are all automatically generating the exported version number. After switching swagger to use 2.3.7, the version numbers are now correctly included. So, I needed just a single line change ;-)
regards, aki

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -481,7 +481,7 @@
         <joda-version>1.2</joda-version>
         <joda-time-version>2.2</joda-time-version>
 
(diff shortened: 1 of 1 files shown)


webron:
:+1:

PR #1468, opened 2015-10-02
Add support for specifying vendor extensions on ref props

roykachouh:
I hope you don't mind this sudden PR. 

I'm using the swagger spec 2.0 with a version of my own custom code generation.  There are situations where I want to add vendor extensions on a $ref object.

For example,

```javascript
definitions:
  Foo:
    type: object
    properties:
      myRefProp:
        x-my-special-meta-annotation: true
        $ref: '#/definitions/Bar'
```
It looks like currently the vendor extensions aren't supported on $ref properties.

This commit basically adds that support, hope you can approve.

Thanks.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
+++ b/modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
@@ -145,7 +145,9 @@ public class PropertyDeserializer extends JsonDeserializer<Property> {
 
         JsonNode detailNode = node.get("$ref");
         if (detailNode != null) {
-            return new RefProperty(detailNode.asText()).description(description);
+            RefProperty refProperty = new RefProperty(detailNode.asText()).description(description);
+            refProperty.setVendorExtensionMap(getVendorExtensions(node));
+            return refProperty;
         }
 
         if (ObjectProperty.isType(type)) {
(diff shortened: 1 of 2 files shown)


webron:
I'm afraid you can't add vendor extensions to $refs. By definition, any properties in the JSON Reference other than the $ref itself will be ignored. This is beyond our control.

roykachouh:
Ok, I was worried about that.

I probably can do something like this instead, do you agree?

```javascript
definitions:
  Foo:
    type: object
    x-fields-with-my-special-annotation:
      - fooRef
      - another one
    properties:
      fooString:
        type: string
        minLength: 5
        maxLength: 500
```

webron:
That's a valid structure in general, but I don't follow your intent :)

roykachouh:
Put simply, i wrote a generator that uses the swagger spec and has some of our custom approaches for dealing with value types.

I'd like to be able to detect items in the spec and add some custom annotations to a property of the field...let's say a custom javax.validation annotation.

I was able to do this using vendor extensions on all properties of a definition, but couldn't do it with a $ref. I'm thinking if the structure above is vaild, I can use that...similar to how I use the required fields in the spec to slap on some @NotEmpty annotations on my generated models.

webron:
Well, strictly speaking of structure, your second example is valid.

webron:
Closing this as it cannot be merged (explained in comments).

Issue #1373, opened 2015-08-17
Composed models duplicate properties from base models.


tomtit:
Swagger version is 1.5.2 (the same issue in 1.5.4-SNAPSHOT).
Here is the classes hierarchy:
```
@ApiModel("MyProperty")
@JsonSubTypes({@JsonSubTypes.Type(value = UrlProperty.class), @JsonSubTypes.Type(value = ValueProperty.class)})
public static abstract class AbstractProperty {
    final String type;

    protected AbstractApiProperty(String type) {
        this.type = type;
    }

    public String getType() {
        return type;
    }
}

@ApiModel
public static class UrlProperty extends AbstractProperty {
    final URL url;

    public UrlProperty(String type, String url) {
        super(type);
        try {
            this.url = new URL(url);
        } catch (MalformedURLException e) {
            throw new IllegalArgumentException(e);
        }
    }

    public URL getUrl() {
        return url;
    }
}

@ApiModel
public static class ValueProperty extends AbstractProperty {
    final String value;

    public ValueProperty(String type, String value) {
        super(type);
        this.value = value;
    }

    public String getValue() {
        return value;
    }
}
```
Swagger definitions for these classes looks as follows:
```
{
  definitions":{
    "MyProperty":{
      "type":"object",
      "properties":{
        "type":{
          "type":"string"
        }
      }
    },
    "UrlProperty":{
      "allOf":[
        {
          "$ref":"#/definitions/MyProperty"
        },
        {
          "type":"object",
          "properties":{
            "type":{
              "type":"string"
            },
            "url":{
              "type":"string",
              "format":"url"
            }
          }
        }
      ]
    },
    "ValueProperty":{
      "allOf":[
        {
          "$ref":"#/definitions/MyProperty"
        },
        {
          "type":"object",
          "properties":{
            "type":{
              "type":"string"
            },
            "value":{
              "type":"string"
            }
          }
        }
      ]
    }
  }
}
```
I would expected to get `UrlProperty` and `ValueProperty` models without the `type` property, as this property is defined in the base model.

PR #1488, opened 2015-10-09
Issue #1373: Composed models duplicate properties from base models

iushankin:


webron:
You forgot the magic word ;)

Issue #1440, opened 2015-09-15
The ApiOperation.nickname property does not work


kvnazar:
The reason: please check this class:  io.swagger.jaxrs.Reader, line 711
I think there should be: operationId = apiOperation.nickname();
instead of: operationId = method.getName();

Sincerely yours
Nazar

PR #1495, opened 2015-10-13
Testcase for issue #1440 and PR #1453

gobolt:
I started to fix a problem that apparently was already fixed in PR #1453 for issue #1440.  Looks like there wasn't a testcase added for the fix, so will offer that contribution.


webron:
Thanks!

PR #1501, opened 2015-10-16
Feature/example values

fehguy:
Adds example values for query, header, form, and path params.  The goal is to differentiate between the `default` value, which is intended to be "what the server will do by default if no value is passed".  The example construct is not supported in the 2.0 specification so it is being exposed using `x-example` for scalar parameters and `x-examples` for body parameters.

For example:

```java
@ApiOperation(value = "test")
@GET
@Path("external/info")
public void queryExample(@ApiParam(value = "test",
        example = "a,b,c") @QueryParam("tenantId") ArrayList<String> tenantId) {
    return;
}
```

This is represented as:
```json
{
  "name" : "tenantId",
  "in" : "query",
  "description" : "test",
  "required" : false,
  "type" : "a,b,c",
  "items" : {
    "type" : "string"
  },
  "x-example" : "a,b,c"
}
```

will allow the string array to have an example value of `a, b, c`

Likewise for a post body:

```java
@ApiOperation(value = "test")
@POST
@Path("external/info")
public void postTest(@ApiParam(value = "test",
        examples = @Example(value = {
                @ExampleProperty(mediaType="application/json", value="[\"a\",\"b\"]")
        })) ArrayList<String> tenantId) {
    return;
}
```

This is represented as:
```json
{
  "in" : "body",
  "name" : "body",
  "description" : "test",
  "required" : false,
  "schema" : {
    "type" : "array",
    "items" : {
      "type" : "string"
    }
  },
  "x-examples" : {
    "application/json" : "[\"a\",\"b\"]"
  }
}
```

### DIFF
--- a/modules/swagger-annotations/src/main/java/io/swagger/annotations/Example.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/annotations/Example.java
@@ -29,14 +29,6 @@ import java.lang.annotation.Target;
 @Target(ElementType.ANNOTATION_TYPE)
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Example {
-
-    /**
-     * An option name for these examples.
-     *
-     * @return an option name for these examples - will be prefixed with "x-"
-     */
-    String name() default "";
-
     /**
      * The examples properties.
      *
--- a/modules/swagger-annotations/src/main/java/io/swagger/annotations/Example.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/annotations/Example.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2015 SmartBear Software
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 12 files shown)


modules/swagger-annotations/src/main/java/io/swagger/annotations/Example.java
```diff
@@ -0,0 +1,47 @@
...
+@Target(ElementType.ANNOTATION_TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Example {
+
+    /**
+     * An option name for these examples.
+     *
+     * @return an option name for these examples - will be prefixed with "x-"
+     */
+    String name() default "";
```
webron: This is not used by the code.

modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
```diff
@@ -341,6 +382,13 @@ public ApiParam getAnnotation() {
         public boolean isHidden() {
             return apiParam.hidden();
         }
+
+        @Override
+        public String getExample() {
+            return apiParam.example();
+        };
+
+        public Example getExamples() { return apiParam.examples();};
```
webron: Shouldn't be a single line.

modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
```diff
@@ -408,5 +456,12 @@ public ApiImplicitParam getAnnotation() {
         public boolean isHidden() {
             return false;
         }
+
+        @Override
+        public String getExample() {
+            return apiParam.example();
+        };
+
+        public Example getExamples() { return apiParam.examples();};
```
webron: Shouldn't be a single line.

fehguy:
I've addressed the two comments from before.

fehguy:
Need to make sure the example value has the correct type, currently always `string`.

Issue #1244, opened 2015-07-07
Reorganize JAX-RS Module


webron:
We have 3 modules today:
1. jaxrs - generic JAX-RS 1.0 support.
2. jersey-jaxrs - including file upload support for Jersey 1.X.
3. jersey2-jaxrs - support for JAX-RS 2.0 (`@BeanParam`) and Jersey 2.X file uploads.

With the new extension mechanism in core 1.5.X, we should create mini-modules for more specific support:
1. jax-rs 1
2. jax-rs 2
3. Jersey 1/Jersey 2/RESTEAsy/Wink/CXF/... file upload support.

We can then create bundles for specific implementations which would include the relevant combination of jax-rs and file upload support. Eventually, most users would use these bundles.

This would also be an opportunity to drop the jax-rs and specific implementation (jersey/jersey2) as `compile` dependencies and make them `provided`. Swagger-Core doesn't need to bring in those dependencies as whoever uses swagger-core already has a working JAX-RS library they depend on, with their own version. This would ease dependency management.

webron:
May also want to take #997 into account here.

webron:
Related to #1031 as well.

elakito:
Hi Ron,
As a temporal workaround that does not break nor change anything, I changed my local copy of SwaggerJersey2Jaxrs to use reflection to get the jersey's FormDataParam's value. With this change, I could use this swagger-jersey2-jaxrs in my cxf-3.x use case having @BeanParam without having any jersey2 dependency. Would you consider this as an option for 1.5.4 until the package is refactored in the future?

regards, aki

elakito:
Hi @webron @fehguy 
I got another idea about this.

How about adding the jaxrs 2.0 BeanParam support into swagger-jaxrs using reflection. It seems we need reflection only for the annotation type check, the rest of the code has no jaxrs-2.0 reference, so it can be directly used there.

This means, we can keep the current build dependency for swagger-jaxrs, that is to use jsr311-api.
But in runtime, if we find jsr399 on the classpath, we can support BeanParam.

Here you can see the code in my repo
https://github.com/swagger-api/swagger-core/compare/master...elakito:work-1244b

Basically, I moved part of the code from SwaggerJersey2Jaxrs into DefaultParameterExtension.
(I haven't deleted the corresponding code from SwaggerJersey2Jaxrs yet, so technically speaking not "moved" but "copied" at the moment)

[...]

PR #1511, opened 2015-10-21
 #1244 use reflection to support BeanParam

elakito:
Hi,
I created a PR based on the current 1.5.5-SNAPSHOT.

I moved the part of jsr-399 specific handling code from swagger-jersey2-jaxrs to swagger-jaxrs using reflection and the relevant unit tests are still executed in swagger-jersey2-jaxrs as jsr-399 is not linked to swagger-jaxrs.

And I verified the updated swagger-jaxrs using cxf in standalone and in osgi using BeanParam.

I would appreciate if you could look at it and merge it if you you see no objection.

Thanks.
Regards, aki


frantuma:
merged in #1683

PR #1513, opened 2015-10-22
Method level reader extensions

kendzi:
It should be possible to enrich swagger model by vendor based extensions stored on operation level. To do that we need to add extension point when operation is processed and method context exists. This allow to process e.g. some vendor annotation which are stored on method level and could be used by documentation.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -856,7 +856,19 @@ public class Reader {
             Response response = new Response().description(SUCCESSFUL_OPERATION);
             operation.defaultResponse(response);
         }
+        
+        processOperationDecorator(operation, method);
+        
         return operation;
+    }
+
+    private void processOperationDecorator(Operation operation, Method method) {
+        final Iterator<SwaggerExtension> chain = SwaggerExtensions.chain();
+        if (chain.hasNext()) {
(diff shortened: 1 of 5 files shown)


webron:
Not entirely sure what's your end goal here, but you should definitely add a test for it (which would also explain your case).

kendzi:
I added example as you asked. In short lets assume that come company created set of annotation which could be used on method level e.g. "@VendorFunnyAnnotation". If it is used this indicates that that method could return some error code and we would like to add that error code to swagger documentation.

webron:
@frantuma can you review this please?

frantuma:
@webron @kendzi 
It looks good to me, but it needs to be rebased and to have `SimpleResourceWithVendorAnnotation` annotated with `@Path("/")` at class level to allow it to be scanned during test.

kendzi:
done

kendzi:
@webron @frantuma anything else should I rebase of fix?

sunlaud:
Upvote this pull request! It really simplifies unobtrusive tuning of documentation generation process.
Are there any reasons why not to merge?

iherasymenko:
+1

kendzi:
I re-base it once again. Please merge it.

sunlaud:
@webron @frantuma guys, are there any reasons why this PR can not be merged?

webron:
Time 😕 

@frantuma since you reviewed it in the past, feel free to merge if you feel confident about it.

PR #1515, opened 2015-10-23
Default value returns correct type

shadidchowdhury:
Currently parameter default value is always returned as String regardless of its type. So, if I generate swagger definition from ```@DefaultValue("true") @QueryParam("async") boolean async``` it will generate the definition with boolean value as String       ``` - name: "async"
        in: "query"
        ... default: "true"  ``` which is wrong.

This pr changes it to return ```getDefault()``` instead of String value, which returns the correct type instead of String.

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
@@ -3,16 +3,14 @@ package io.swagger.models.parameters;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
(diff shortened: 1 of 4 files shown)


modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
```diff
@@ -144,8 +142,8 @@ public void setProperty(Property property) {
         }
     }
 
-    public String getDefaultValue() {
-        return defaultValue;
+    public Object getDefaultValue() {
```
tomtit: The signature of this method should not be changed for backward compatibility. If you need to get object type instead of string value just use the `Object getDefault()` method.
Handling of default values has been fixed in version `1.5.4` (PR #1378, unit test for boolean values looks as [follows](https://github.com/swagger-api/swagger-core/pull/1378/files#diff-f64eab520f5cc6c19526a79d67f5d444R24)) - could you please check the version of the `swagger-core` artifact in your project?

shadidchowdhury: Yeah I confirm it is fixed in 1.5.4, closing this pr.

PR #1552, opened 2015-11-25
Include responses from a method's declared exceptions

jawher:
A common pattern in JAX-RS applications is to have resources throw
custom exceptions, and configure an ExceptionMapper to map them to a
specific HTTP status code.

This pattern does not mesh well with how swagger collects the APIs
info: swagger expects the developer to declare for every method the
list of possible responses, regardless of the exceptions it declares to
throw.

```java
@PUT
@ApiOperation("desc")
@ApiResponses({
        @ApiResponse(code = 204, message = "Everything is awesome")
        @ApiResponse(code = 500, message = "Internal error")
        @ApiResponse(code = 502, message = "Upstream error")
})
@Path("/op")
void operation(Params params) throws UpstreamServerError, InternalServerError;
```

This pull request proposes to also check the exceptions declared in the
throws clause of a method for `@ApiResponse` annotations, and include
them in the method possible responses.

So that the previous example would become:

```java
 @PUT
 @ApiOperation("desc")
 @ApiResponses({
        @ApiResponse(code = 204, message = "Everything is awesome")
 })
 @Path("/op")
 void operation(Params params) throws UpstreamServerError, InternalServerError;
 ```

By moving the other response codes to the exceptions:

```java
@ApiResponses({
       @ApiResponse(code = 502, message = "Upstream error")
}
public class UpstreamServerError extends RuntimeException {
  :
}
```

In my particular case, I have an app with hundreds of resource methods
which already declare the possible exceptions they throw.
This improvement would save me hundreds and hundreds of repetitions of
return statuses.

### DIFF
--- a/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiResponses.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiResponses.java
@@ -29,7 +29,7 @@ import java.lang.annotation.Target;
  *
  * @see ApiResponse
  */
-@Target(ElementType.METHOD)
+@Target({ElementType.METHOD, ElementType.TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface ApiResponses {
     /**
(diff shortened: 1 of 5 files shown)


webron:
Thanks, that's an interesting approach. Can you add another use case for your test that covers both an `@ApiResponse` annotation on the method and the custom exception? That would be more complete.

jawher:
@webron Thanks for the reply.

PR updated accordingly.

jawher:
@webron Any news on this ? :)

fehguy:
Hi @jawher this is great.  Sorry for the delay, @webron was waiting for my review, and I'm both slow and lazy.

jawher:
@fehguy Thank you for accepting this PR, and no worries for the delay :+1:

PR #1563, opened 2015-12-06
Code change for 1482 and 1514

Kvezon:
I was trying to fix 1482


fehguy:
This was handled in #1611

Issue #1343, opened 2015-08-10
Swagger scans methods on JAX-RS resources not annotated to be REST service methods


oweise:
Information copied from this [google group thread](https://groups.google.com/forum/#!topic/swagger-swaggersocket/AbnvBVrflDs).

On scanning JAX-RS resources for the REST interface Swagger not only scans those methods that are annotated to be service methods with `@GET/PUT/POST/DELETE` and `@Path`. It seems to scan all methods on the resource class. This leads to a scanning failure if those non-service methods return types that are not appropriate to be served via service.

Imagine the following resource class, which is a draft of a real class on a Web Service we use:

    @Path("/v1")
    @Api(value="root")
    @XmlRootElement(name=RootResource.RESOURCE_TYPE)
    @XmlDiscriminatorValue(RootResource.RESOURCE_TYPE)
    public class RootResource  {

      @Path(REFLIST_DBS + "/{dbkey}")
      @ApiOperation(value="Retrieve a database resource")
      public DatabaseResource getDatabase(@ApiParam(name="dbkey",value="Database key") @PathParam("dbkey") String dbKey) throws WGException {
          return new DatabaseResource(this, dbKey);
      }

      public RestApplication getApplication() {
         return _application;
      }
    }

The method getDatabase() annotated with `@Path` is a service method, therefor should be scanned by Swagger. The method getApplication() is not annotated and is for internal use. Via scanning this method Swagger descends onto further types and their methods not meant to be service resources until, in our case, it reaches a type org.apache.log4j.spi.LoggerRepository which provokes a scanning error:

    java.lang.IllegalArgumentException: Conflicting setter definitions for property "threshold": org.apache.log4j.spi.LoggerRepository#setThreshold(1 params) vs org.apache.log4j.spi.LoggerRepository#setThreshold(1 params)
    at com.fasterxml.jackson.databind.introspect.POJOPropertyBuilder.getSetter(POJOPropertyBuilder.java:293)
    at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:264)
    at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:151)
    ...
    at io.swagger.jaxrs.Reader.read(Reader.java:146)
    at io.swagger.jaxrs.config.BeanConfig.setScan(BeanConfig.java:170)
    at de.innovationgate.wga.services.rest.RestApplication.<init>(RestApplication.java:267)

[...]

PR #1588, opened 2016-01-04
Swagger scans methods on JAX-RS resources not annotated to be Swagger operations

titiu7:
This is related to issue #1343. The fix that was provided allowed Swagger to skip methods without REST annotations which was just one part of the problem. In Swagger-Core v1.5.4, all the methods annotated to be REST service methods are scanned even if they don't have the annotation @ApiOperation. I believe this shouldn't be the case according to the Swagger documentation. 
As an example, let's assume we have the following method in a REST resource class:

```
@GET 
@Path("/")
public String getString() {
 String str = "test method";
return str;
}
```

Swagger will include that method regardless the fact that it's not annotated with @ApiOperation. This pull request addresses that issue.

Thanks


fehguy:
I don't believe we can merge this, as it fundamentally changes how the processor works.  I'm going to keep it open so documentation can be improved but also ensure it's behaving as intended.

titiu7:
Hey @fehguy, do you think this issue will be addressed anytime soon in the upcoming versions?

webron:
@titiu7 - did you get a chance to test 1.5.8 for that functionality? I know @frantuma has done some work in the field.

titiu7:
@webron thanks for your reply. I tested out 1.5.8, but that functionality is still not there yet. @frantuma, can you share your thoughts on this. Thanks

pixels4nickels:
Is there any workaround for this? Is there any way to exclude packages or something? Servlets using Enterprise Web Objects are invoking this error for me:
`java.lang.IllegalArgumentException: Conflicting setter definitions for property "userInfo": com.webobjects.eocontrol.EOObjectStore#setUserInfo(1 params) vs com.webobjects.eocontrol.EOObjectStore#setUserInfo(1 params)`

fehguy:
We've clarified how the processing picks up annotations, please see here for details:

https://github.com/swagger-api/swagger-core/releases/tag/v1.5.9

PR #1603, opened 2016-01-14
Fix #1103. Do not rely on ServletContext to initialize swagger instance.

### LINKED ISSUE #1103: ServletContext null - Jersey + Grizzly2 Standalone - 1.5.x
When using Grizzly2 in standalone mode, ServletContext is not available.

NPE at 
- `com.wordnik.swagger.jaxrs.listing.ApiListingResource.scan(ApiListingResource.java:47)`
- `com.wordnik.swagger.jaxrs.listing.ApiListingResource.getListingJson(ApiListingResource.java:68)`

https://jersey.java.net/documentation/latest/jaxrs-resources.html#d0e2742 [3.6]

pierredeman:
I based my fix on PR#1535.
Then, I added a little object based on PR#1563 to maintain multi-tenancy of swagger definitions in a single app.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/config/BeanConfig.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/config/BeanConfig.java
@@ -8,11 +8,7 @@ import io.swagger.config.ScannerFactory;
 import io.swagger.config.SwaggerConfig;
 import io.swagger.core.filter.SwaggerSpecFilter;
 import io.swagger.jaxrs.Reader;
-import io.swagger.models.Contact;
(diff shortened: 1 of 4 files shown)


webron:
While generally speaking I'd like to be able to not depend on the ServletContext, we probably can't remove the dependency altogether. Some existing implementation depend on the reading the Swagger object from the ServletContext and this would be a breaking change for such implementations.

pierredeman:
Ok I see... I added a read / write from the ServletContext when scanning the resources.

fehguy:
I believe a similar support was just merged in https://github.com/swagger-api/swagger-core/commit/d319d9a55c2c77870aeff75bcfc2038f254b3697

Issue #1482, opened 2015-10-08
Use of static Scanner in ScannerFactory blocks use of multiple Swagger configs in the same JVM


tfeak:
I'm running 2 REST APIs in the same JVM on different ports. They both use Swagger. When calling BeanConfig.scan(true), the ScannerFactory instance is set to the BeanConfig instance making the call. The result is that the second call to this is the one that "wins".

ScannerFactory shouldn't be using a singleton static model. This is a perfect example of the unintended side effect that can occur when an object model leverages static.

As a workaround, I've tried setting up the BeanConfig to scan the packages from both apps. All of the API calls show up in the Swagger UI correctly, but because it's on the wrong port sample calls cannot be made.

EDIT: To clarify, this is part of the swagger-jaxrs packaging

fehguy:
Correct, this should be changed.  Targeting 1.5.5

fehguy:
fixed by https://github.com/swagger-api/swagger-core/commit/8031d514f2487e26f73b080a339c993b1964ec61

Issue #1384, opened 2015-08-21
run two swaggers in one tomcat context


crazySocket:
I have REST service big enough to distinguish admin and user parts. My administrator wants to have them separate. You can do that by deploying application two times with differently configured web.xml. Unfortunately, this solution is disliked.

So, I have one context on tomcat. I configure web.xml like this (necessary details only)
```
<servlet>
    <servlet-name>restServiceAdmin</servlet-name>
    <servlet-class>org.glassfish.jersey.servlet.ServletContainer</servlet-class>
        <init-param>
            <param-name>jersey.config.server.provider.packages</param-name>
            <param-value>some.package.admin</param-value>
        </init-param>
</servlet>

<servlet>
    <servlet-name>restServiceUser</servlet-name>
    <servlet-class>org.glassfish.jersey.servlet.ServletContainer</servlet-class>
        <init-param>
            <param-name>jersey.config.server.provider.packages</param-name>
            <param-value>some.package.user</param-value>
        </init-param>
</servlet>

<servlet-mapping>
    <servlet-name>restServiceAdmin</servlet-name>
    <url-pattern>/rest/admin/*</url-pattern>
</servlet-mapping>

<servlet-mapping>
    <servlet-name>restServiceUser</servlet-name>
    <url-pattern>/rest/user/*</url-pattern>
</servlet-mapping>

<servlet>
    <servlet-name>SwaggerBootstrapAdmin</servlet-name>
    <servlet-class>io.swagger.api.util.BootstrapAdmin</servlet-class>
    <load-on-startup>1</load-on-startup>
</servlet>
<servlet>
    <servlet-name>SwaggerBootstrapUser</servlet-name>
    <servlet-class>io.swagger.api.util.BootstrapUser</servlet-class>
    <load-on-startup>2</load-on-startup>
</servlet>
```

Although, boostrap servlets have appropriate base paths and package resources set, when I download swagger.json I get definition configured in BootstrapUser. The last SwaggerBootstrap to start is the one to set contents of swagger.json regardless of how I obtain it (via /rest/admin/swagger.json or /rest/user/swagger.json).

I would like /rest/admin/swagger.json or /rest/user/swagger.json to return definitions of resources some.package.admin, some.package.user accordingly. My question is: is it possible to achieve the result I intend in swagger at all?

webron:
As it is written right now it *may* be possible, but it's not trivial. @fehguy may have more info on that.

crazySocket:
[...]

PR #1611, opened 2016-01-19
Issue 1482 refactor

frantuma:
As discussed with @fehguy, PR aims at fixing and references issues #1482  #1384, PR #1397.

It basically allows to configure more than one "Swagger context" (logical concept including SwaggerConfig, Swagger initial info and used Scanner instance) in same JVM and same web application, one for each configured servlet/servletConfig.

An example of "multi context" setup can be find in samples project PR https://github.com/swagger-api/swagger-samples/pull/46
where a "swagger.context.id" initParam is provided for each separate "swagger context" instance (see sample web.xml)

Approach is sub-optimal, but as discussed quite some additional refactor would be needed to accomplish a nicer solution, with quite high risk related to compatibility.


webron:
As @fehguy mentioned, I assume this doesn't solve #1103. I think it's fine if it solves the other issues, and the solution to #1103 will be a separate one (there's a pending PR which may solve it, I think).

frantuma:
@webron it should actually possibly solve it, I just updated last remaining unchecked ref to context  (kind of quick fix, ReaderConfig still only configurable via context), but tests and samples would need to be run/checked to confirm

fehguy:
I've verified this works.  There is a slight change in usage though--as shown in the samples, this:

```
context.setAttribute("swagger", swagger);
```

has been replaced with this:
```
new SwaggerContextService().withServletConfig(config).updateSwagger(swagger);
```

We'll update the release notes

PR #1618, opened 2016-01-21
refs #1438 adds jsr-303 to param

### LINKED ISSUE #1438: Support for JSR-303 (Bean Validation) annotations for JAX-RS API parameters
JAX-RS supports Bean Validation integration and so should Swagger-Jaxrs.

Currently, Bean Validation support is implemented on model level (#981, #800), however, support for parameters (@QueryParam, @PathParam, etc.) is still missing.

This would make a very convenient addition to Swagger-JaxRS because now it's really painful to have duplicate information for JAX-RS parameters - one using @ApiParam, one using JSR-303 annotations, like here:

```java
@GET
	@Produces(MediaType.APPLICATION_XML)
	@ApiOperation(value = "Get (filtered) list of reports", notes = "This method enables comfortable server-side filtering of available reports")
	ReportDtoList getReports(
			@ApiParam(value = "Type of the component to get reports for"/* , required=true */) @NotNull @QueryParam("componentType") ComponentType componentType,
			@ApiParam(value = "ID of the component to get reports for"/* , allowableValues = "range[1, infinity]", required = true */) @NotNull @Min(1) @QueryParam("componentId") Long componentLogicalId,
			@ApiParam(value = "Report inclusion parameter"/* , required = true */) @NotNull @QueryParam("componentInclusion") ComponentInclusion componentInclusion,
			@ApiParam(value = "The requested report type") @QueryParam("type") ReportType type,
			@ApiParam(value = "The requested report range(s)"/* , allowMultiple = true */) @QueryParam("range") List<ReportRange> ranges);

```java

At least, these annotations should be supported (and any others which make sense for Swagger):
* @NotNull
* @Min
* @Max
* @Size

frantuma:
As discussed with @fehguy, fixes #1438 adding support for JSR-303 annotations @Min, @Max and @NotNull (@Size was already processed) for operation parameters.

Behvaiour changes include: 

* If @NotNull, parameter 'required' will always be true (e.g. also if @ApiParam (required = false))

* If @Min(x) or @Max(x), parameter 'minimum' or 'maximum' will be set to x only if @ApiParam (allowableValues) is not specified.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
+++ b/modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
@@ -26,6 +26,9 @@ import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
 
 public class ParameterProcessor {
@@ -73,6 +76,9 @@ public class ParameterProcessor {
             if (helper.getMaxItems() != null) {
                 p.setMaxItems(helper.getMaxItems());
             }
+            if (helper.isRequired() != null) {
+                p.setRequired(true);
+            }
 
             AllowableValues allowableValues = AllowableValuesUtils.create(param.getAllowableValues());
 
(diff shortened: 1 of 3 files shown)


fehguy:
@frantuma looks great.  Does this mean some of the code in the special module here:

https://github.com/swagger-api/swagger-core/tree/master/modules/swagger-hibernate-validations

can be removed?

frantuma:
@fehguy I believe they serve different purposes; this PR (and generally related code) handles jsr-303 annotations for non "body" operation parameters (therefore primitive), while hibernate BeanValidator is used during model/definition processing, e.g. for body parameters etc.

Nogyara:
Thank you guys for implementing #1438, it'll really makes life of some devs easier, helping to remove duplication for REST API defined fully by JEE standard annotations (JAX-RS + Bean Validation)!

However, I just tested the brand new 1.5.7 release and the implementation seems to be working only for some param types - we use both @PathParam and @QueryParam in our APIs and the implementation seems to be supporting @PathParam only. Should I reopen #1438 or is it better to create a separate ticket?

webron:
@Nogyara probably better to open a new ticket with details of what's missing. Thanks.

Issue #1557, opened 2015-11-30
After filtering swagger.json also remove no longer needed references


trallus:
Hi, I have two swagger.json files from which I only take parts and merge these to one file. 

For filtering I use a class that extends the AbstractSpecFilter, which works very well except that no longer needed references are still inside the merged json. At the moment i scan the parameters/responses of the operation in the isOperationAllowed method to find the ones that are still needed. At the moment this is based on a lot of casting and class filtering.

So I wanted to ask if it is possible to remove all no longer needed references in an easier way or if it would be possible to add a method to get all References for an Operation directly?

webron:
You have references that are no longer needed or models that are no longer referenced? Not sure how the first one is happening - if it does, would appreciate an example to understand better.

trallus:
Hi, thanks for your fast reply, it's the second case i have models that are no longer referenced and for me these are "no longer needed" inside the json, so I want to get rid of them.

webron:
That sounds like a bug, I'm afraid as those should be filtered automatically. We try to filter out what's not needed (for example, tags are being filtered if they have no operations).

webron:
Resolved in master.

trallus:
Hello, I'm sorry to bother you again with this topic but after testing the new "isRemovingUnreferencedDefinitions" method i found two problems:

1. I have two responses that are just an array of objects, for which the reference is still there but the definition is removed e.g `[ Entry{...} ]` becomes `[ undefined ]`
2. Definitions of Models are not removed when they are inside other models, even if these are removed e.g. `MyUserDetails{ User {...} }` "MyUserDetails" is removed but "User" is not

webron:
@trallus - would you mind opening a new ticket for those?

PR #1621, opened 2016-01-22
#1557 filter out unreferenced definitions if specified in filter

frantuma:
@fehguy 

solution for #1557, allowing to filter out unreferenced definitions from returned Swagger, if specified in filter by extending `AbstractSpecFilter` and overriding `isRemovingUnreferencedDefinitions()`:

```java
public class RemoveUnreferencedDefinitionsFilter extends AbstractSpecFilter {
    public boolean isRemovingUnreferencedDefinitions() {
        return true;
    }
}
```

Checks for the following references and compares with definitions:

swagger.parameters.parameter(BodyParam).schema
swagger.responses.Response.schema
swagger.paths.Path.parameters.parameter(BodyParam).schema
swagger.paths.Path.operations.Operation.responses.Response.schema
swagger.paths.Path.operations.Operation.parameters.parameter(BodyParam).schema

Method `isRemovingUnreferencedDefinitions` has not been added to interface `SwaggerSpecFilter` to avoid breaking compatibility with existent client filters directly implementing `SwaggerSpecFilter`.

Cleaning is performed after all other filters (in additional iteration) to possibly support future updates e.g. with multiple filters, etc, as kind of "final cleanup".

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/core/filter/AbstractSpecFilter.java
+++ b/modules/swagger-core/src/main/java/io/swagger/core/filter/AbstractSpecFilter.java
@@ -38,4 +38,8 @@ public abstract class AbstractSpecFilter implements SwaggerSpecFilter {
             Map<String, List<String>> headers) {
         return true;
     }
+
+    public boolean isRemovingUnreferencedDefinitions() {
+        return false;
+    }
 }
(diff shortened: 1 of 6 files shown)


fehguy:
Looks good!  At first I was not understanding why we aren't just doing the cleanup in the `SpecFilter.java` after cloning but thinking about it more, you may not want to clean up all the unused references.

Hypenate:
How do you use this in 2023? :)

Issue #1614, opened 2016-01-20
Interesting use of JAX-RS and Swagger annotations confuses BeanConfig scanning


Helmsdown:
@olensmar @fehguy 

The following will make the most sense if you check out our example project we are using to expose the problem: https://github.com/jithumenon/swagger-karyon. 

Note: The swagger-karyon project README describes how to run the app.

We ran into a problem with BeanConfig's scanning logic in the following scenario:

1. You have a "root" resource class that (in the example I reference below this class is RootResource.java). This class is annotated with JAX-RS's @Path as well as Swagger's @Api.
2. RootResource exposes child resources by an instance method, annotated with an @Path, that returns an instance of a "child" JAX-RS annotated class (for this example the child resource will be called PetResource.java)
3. PetResource.java has a class level @Api annotation and its methods implement the actualy JAX-RS endpoints (methods with a @GET, @POST, etc). 

The current implementation of BeanConfig.java looks for the following:
1. Classes that have an @Api annotation
OR
2. Classes that have a @Path annotation
OR
3. Classes that have a @SwaggerDefinition annotation

With our example, the "root" resource is picked up by BeanConfig's annotation scan (which is good and expected) but the "child" resources are picked up and considered "root" resources as well. The swagger ends up with paths that are not actually in the working API (For example the swagger.json in the aforementioned example contains a /findByStatus when the real path is /api/1.0/pets/findByStatus).

While investigating this issue I modified BeanConfig's scanning logic to completely ignore classes with an @Api and only search for classes annotated (at the class level) with @Path. This fixed the problem for me...but I suspect it might cause problems for others if this were merged in. 

Your thoughts and advice? Have you seen this before? I am willing to submit a PR to make BeanConfig smarter.

Thanks!

fehguy:
@frantuma you've been working on similar code--do you have any ideas on this?

frantuma:
@fehguy issue is also related to the way Reader identifies sub resources; it currently scans for @Api annotated return types in method of root resource; therefore it "correctly" identifies subresource if annotated with @Api and creates correct path/operation including parent, but because subresource is annotated with @Api it processes it again wrongly as root resource;

As @Helmsdown mentioned, if @Api annotated classes are not considered while scanning (in BeanConfig), the issue doesn't occur; I would say that it shouldn't cause bad side-effects, because in a jax-rs environment root resources should anyway be annotated with @Path.

[...]

Issue #1106, opened 2015-05-26
Automatically hide subresources


webron:
Right now, subresources are scanned and added to the `swagger.json` as if they're hosted at the root of the application. The workaround for now is to add `@Api#hidden() = true`. The better behavior would be to scan only add it as a resource to the containing resource and not treat it as a first-class resource.

fehguy:
This could be tricky to support in the tree, when scanning to find a parent reference.  Every resource *may* live on it's own, so if you don't want it to be picked up, I don't see how you can avoid marking it as `hidden` at the `@Api` annotation.

webron:
Why would it need the `@Api` annotation in the first place if it's not live on its own?

Helmsdown:
@webron The subresource would want an @Api annotation so you can tag it and have it show up in swagger-ui in its own expander.

webron:
@Helmsdown - the amount of dust I needed to clean in order to remember what this was even about is insane, and I can't still can see clearly. This came from a specific use case, and perhaps that's the key. I believe this relates to sub-resources that are not annotated with `@Path` at the class level.

Helmsdown:
I currently have that use case

On Thursday, January 21, 2016, Ron <notifications@github.com> wrote:

> @Helmsdown <https://github.com/Helmsdown> - the amount of dust I needed
> to clean in order to remember what this was even about is insane, and I
> can't still can see clearly. This came from a specific use case, and
> perhaps that's the key. I believe this relates to sub-resources that are
> not annotated with @Path at the class level.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/swagger-api/swagger-core/issues/1106#issuecomment-173776156>
> .
>


-- 
Sent from Gmail Mobile

Helmsdown:
We have a discussion going in #1614 about how to solve this (if at all).

webron:
Yeah, I actually get aaaaaaaaaall notifications ;)

Helmsdown:
oh

Helmsdown:
really

Helmsdown:
:-)

frantuma:
fixed in #1646

PR #1636, opened 2016-01-27
Refs #1106, #1629, #1617, #1614 - JAX-RS Root and Subresources identification, scanning and reading

frantuma:
@fehguy Refs #1106, #1629, #1617, #1614

As discussed, changes JAX-RS root and subresources identification, scanning and reading to be compliant with JAX-RS specs: Root resources are identified by @Path annotation at class level, and sub resources are identified by @Path annotation at method level, along with no @HttpMethod annotation.

It integrates changes from #1617, changing BeanConfig scanning to only consider @Path annotations, and #1629, used as base for further implementation.

Classes not annotated with @Path are only read in case config.scanAllResources is true, maintaining anyway @Api.hidden processing (@Api.hidden classes are never processed).

Tests have been adapted by setting always scanAllResources = true to be able to handle current test root resources classes which are not always annotated with @Path. A further enhancement would be comprehensively adjusting current test resources/classes to fully adhere to JAX-RS specs, possibly in scope of a different ticket.

Code has been tested via project tests and running swagger-samples projects.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -190,17 +190,13 @@ public class Reader {
         String[] produces = new String[0];
         final Set<Scheme> globalSchemes = EnumSet.noneOf(Scheme.class);
 
-        // only read if allowing hidden apis OR api is not marked as hidden
-        final boolean readable = (api != null && readHidden) || (api != null && !api.hidden());
-        if (readable) {
+        // only generate tag if not a child resource, @Api exists & @Api is not hidden; otherwise use parentTags
+        if (!readHidden && api != null && !api.hidden()) {
             // the value will be used as a tag for 2.0 UNLESS a Tags annotation is present
             Set<String> tagStrings = extractTags(api);
             for (String tagString : tagStrings) {
                 Tag tag = new Tag().name(tagString);
                 tags.put(tagString, tag);
-            }
-            if (parentTags != null) {
-                tags.putAll(parentTags);
             }
             for (String tagName : tags.keySet()) {
                 swagger.tag(tags.get(tagName));
@@ -232,10 +228,13 @@ public class Reader {
                     securities.add(security);
                 }
             }
+        } else if (parentTags != null) {
+            tags.putAll(parentTags);
         }
 
-        // allow reading the JAX-RS APIs without @Api annotation
-        if (readable || (api == null && config.isScanAllResources())) {
+        // allow reading the JAX-RS APIs without @Api annotation if scanAllResources is enabled or
+        // readHidden is set by parent resource.
+        if (readHidden || (api != null && !api.hidden()) || (api == null && config.isScanAllResources())) {
             // merge consumes, produces
 
             // look for method-level annotated properties
@@ -553,12 +552,10 @@ public class Reader {
             return type;
         }
 
-        if (config.isScanAllResources()) {
-            // For sub-resources that are not annotated with  @Api, look for any HttpMethods.
-            for (Method m : type.getMethods()) {
-                if (extractOperationMethod(null, m, null) != null) {
-                    return type;
-                }
+        // For sub-resources that are not annotated with  @Api, look for any HttpMethods.
+        for (Method m : type.getMethods()) {
+            if (extractOperationMethod(null, m, null) != null) {
+                return type;
             }
         }
 
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -175,12 +175,13 @@ public class Reader {
 
         return read(cls, "", null, false, new String[0], new String[0], new HashMap<String, Tag>(), new ArrayList<Parameter>(), new HashSet<Class<?>>());
     }
-    protected Swagger read(Class<?> cls, String parentPath, String parentMethod, boolean readHidden, String[] parentConsumes, String[] parentProduces, Map<String, Tag> parentTags, List<Parameter> parentParameters) {
-        return read(cls, parentPath, parentMethod, readHidden, parentConsumes, parentProduces, parentTags, parentParameters, new HashSet<Class<?>>());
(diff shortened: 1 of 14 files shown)


fehguy:
@Helmsdown can you please comment if this addresses your use case?

Helmsdown:
@fehguy I think it will. When a snapshot is built and deployed let me know so I can give it a spin.

fehguy:
So you want me to merge, push, then undo if you don't like it?

Helmsdown:
Well if you are offering. :-)

frantuma:
@mtraynham @fehguy

I am not sure to get exactly what you mean by *duplicates the function in swagger-ui (under it's own API section and it's parents)*; 
I guess you refer to tags, but here behaviour haven't changed if I am not mistaken; it basically adds parent tags to operation only if existing, can you detail what you mean?

Actually I had some additional thoughts about the overall issue, basically root and child JAX-RS resources identification; the scenarios below should include possible options, in acending order of "Jax-rs compliance", eventually IMO option 5. would be the best, but all but Option 1. involve change in behaviour, which I am not sure if it is ok in same minor version, etc. 

Please let me know what you think

In regards to subresource identification,  in scenario 1. subresources are considered when any method, even not annotated with @Path returned a class with @Api; in all other scenarios subresources are only considered when method annotated with @Path and not annotated with any @HttpMethod, regardless of scanAllResources.

**1. PREVIOUS BEHAVIOUR**

reads also root resources if not annotated with @Path. if not annotated with @Path and not with @Api also it reads if isScanAllResources is true

|@Api|reads|
|@Path|doesn't read|
|scanAllResources|reads|
|@Api hidden|doesn't read|
|@Api @Path hidden|doesn't read|
|@Api hidden scanAllResources|doesn't read|
|@Api scanAllResources|reads|
|@Api @Path|reads|
|@Api hidden scanAllResources|doesn't read|
|@Path scanAllResources|reads|
|subresource|doesn't read|
|@Api subresource|reads|

**2.CURRENT BEHAVIOUR WITHOUT FIX ON SUBRESOURCE** (if (readable || (api == null && config.isScanAllResources()) || isSubresource) {

Current PR https://github.com/swagger-api/swagger-core/pull/1636

reads root resources only if annotated with @Path. if not annotated with @Path and not with @Api also it reads if isScanAllResources is true


|@Api|**doesn't read**|
|@Path|doesn't read|
|scanAllResources|reads|
|@Api hidden|doesn't read|
|@Api @Path hidden|doesn't read|
|@Api hidden scanAllResources|doesn't read|
|@Api scanAllResources|reads|
|@Api @Path|reads|
|@Api hidden scanAllResources|doesn't read|
|@Path scanAllResources|reads|
|subresource|doesn't read|
|@Api subresource|reads|

**3.CURRENT BEHAVIOUR WITH FIX ON SUBRESOURCE** (if (readable || (`api == null && config.isScanAllResources()) || isSubresource)`)

Current PR https://github.com/swagger-api/swagger-core/pull/1636 with change to be applied
Same as 2. but resources are always read.


|@Api|**doesn't read**|
|@Path|doesn't read|
|scanAllResources|reads|
|@Api hidden|doesn't read|
|@Api @Path hidden|doesn't read|
|@Api hidden scanAllResources|doesn't read|
|@Api scanAllResources|reads|
|@Api @Path|reads|
|@Api hidden scanAllResources|doesn't read|
|@Path scanAllResources|reads|
|subresource|**reads**|
|@Api subresource|reads|

**4.MORE JAX-RS COMPLIANT BEHAVIOUR**

Code at https://github.com/frantuma/swagger-core/tree/issue-1614-enhanced/
reads root resources only if annotated with (@Path and @Api) or (@Path and scanAllResources). if not annotated with @Path classes are never read.

|@Api|doesn't read|
|@Path|doesn't read|
|scanAllResources|**doesn't read**|
|@Api hidden|doesn't read|
|@Api @Path hidden|doesn't read|
|@Api hidden scanAllResources|doesn't read|
|@Api scanAllResources|**doesn't read**|
|@Api @Path|reads|
|@Api hidden scanAllResources|doesn't read|
|@Path scanAllResources|reads|
|subresource|reads|
|@Api subresource|reads|


**5.FULLY JAX-RS COMPLIANT BEHAVIOUR**

Code to be updated at https://github.com/frantuma/swagger-core/tree/issue-1614-enhanced/

|@Api|doesn't read|
|@Path|**reads**|
|scanAllResources|**reads**|
|@Api hidden|doesn't read|
|@Api @Path hidden|doesn't read|
|@Api hidden scanAllResources|doesn't read|
|@Api scanAllResources|doesn't read|
|@Api @Path|reads|
|@Api hidden scanAllResources|doesn't read|
|@Path scanAllResources|reads|
|subresource|reads|
|@Api subresource|reads|

mtraynham:
@frantuma 
As long as the subresource `@Api` annotation doesn't actually have a `value` set then it works as you intended because no [tags were extracted](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java#L215) on the subresource.  When I raised the question yesterday, I thought it would generate an empty `Tag`, but that's not the case.

This does leave room for user error though, when we could just skip `Tag` generation for subResources.  For instance, in the case I did set a `value` on the `@Api` annotation for a subresource, the operation would get two `Tag`s.  Albeit, it's a pretty minor issue, I just mistook the code path as *always* generating a `Tag` for the subresource.
```
@Path("/pets")
@Api("Operations on Pets")
public class Pets {
    public Pets() {
    }

    @GET
    public List<String> getAll() {
        return new ArrayList<>();
    }

    @Path("/{id}")
    public Pet get(@PathParam("id") String id) {
        return new Pet(id);
    }

    @Api("Operations on Pet")
    public static class Pet {
        private final String id;

        public Pet(String id) {
            this.id = id;
        }

        @GET
        public String get() {
            return this.id;
        }
    }
}
```
screenshot from 2016-01-28 11 31 26

fehguy:
@frantuma thank you for the detailed response.  I certainly prefer to be fully JaxRS compliant in our solution.  I do believe this will cause some breakage to users though so the impact of this change should be well documented and covered in the release notes.  It has been our intent to follow JAXRS for this feature.

@mtraynham regarding tags, the behavior in the code is to extract them when they're not explicitly set.  So the best way to "avoid user error" is to programmatically configure the tags like such:

```
@Api(value = "/pet", tags = {"pets", "animals"})
```

PR #1639, opened 2016-01-29
adding ApplicationBeanConfig not using reflections

rmannibucau:
This implementation only uses Application class to find endpoints. Avoid the need of reflections (and transitively javassist). Also makes it faster since it doesnt needs to scan (again if in a container) and more reliable in OSGi)


webron:
Wasn't the goal here to remove the reflections dependency? I'd expect it to be removed from the pom as well.

rmannibucau:
Can be excluded from users pom bjt BeanConfig still relies on it.

tandraschko:
Why was it closed? e.g. with CDI you also don't need reflections.
The performance would be better and the WAR file smaller in many cases!

PR #1646, opened 2016-02-03
Refs #1636 Option 5 - full JAX-RS compliance (#1106, #1629, #1617, #1614)

frantuma:
Refs #1636 Option 5 - full JAX-RS compliance (#1106, #1629, #1617, #1614)

@fehguy

It implements Option 5 behaviour in #1636, as discussed it implies quite a change in behaviour regarding JAX-RS classes scanning/reading, and possibly existing clients would experience issues if based on a "non fully JAX-RS" config.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -26,8 +26,6 @@ import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 import io.swagger.annotations.Authorization;
 import io.swagger.annotations.AuthorizationScope;
-import io.swagger.annotations.Extension;
-import io.swagger.annotations.ExtensionProperty;
 import io.swagger.annotations.Info;
 import io.swagger.annotations.ResponseHeader;
 import io.swagger.annotations.SwaggerDefinition;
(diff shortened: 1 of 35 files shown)


fehguy:
Thank you--I will merge this after 1.5.7 is released.

frantuma:
A possible alternative option to maintain compatibility is to enhance the mechanism to behave according to a flag/property set as default for Option 1 or 2 (as now) and settable to behave as options 3-5

ndebeiss:
Hello everybody

I started looking at swagger project a few weeks ago.
That is very interesting, it is the best documentation generator I found.
I use swagger in order to generate web services documentation for a medium project (4500 classes) I am working on. I still have a lot of work to do in order to make it generate exactly what I want, but swagger was easy to integrate in the project and I had good results quickly.

I saw the last commit for 1.5.7 release and I do not agree with the direction you are taking. The problem I am facing is that there are lots of different web services declarations in my application. There are web services declared as : 
- Axis endpoint
- Spring with annotations
- Spring with XML
- jax-rs
So I extended some code of swagger in order to be compatible with a maximum of those declarations.
Actually, is it not very difficult to do, the code is clear and it is just a matter of fallback from custom swagger annotations, to jax-rs, to Spring (but declaring the fields and methods as "protected" would have helped me :-) ).

So I think swagger project should evolve to a maximum compatibility inside the same code instead of being more restrictive like the last commit. I am not sure the modules swagger-jersey-jaxrs, etc... are important.
That would simplify the installation, the user would have to add a single dependency and swagger would be able to read any kind of declarations.

I read the issue https://github.com/swagger-api/swagger-core/issues/1614 and I saw the need for Path annotations looking at class level, in Spring there is an annotation RequestMedia at class level too.

Thanks for reading, tell me what you think frankly.

Regards

webron:
@ndebeiss - swagger-core is intended to provide support for jax-rs based applications and pure servlet solutions. Other frameworks are out of scope for this project, and some have their own implementations to provide support.

PR #1652, opened 2016-02-05
When no ApiOperation, produces or consumes not read

ndebeiss:
When annotation ApiOperation is not present, the code does not fall back to
standard Produces and Consumes annotations.


fehguy:
@frantuma is this ok?

ndebeiss:
Added. Sorry for the code formatting noise.

frantuma:
replaced by #1680

PR #1663, opened 2016-02-15
Refactor ApiListingResource to not use greedy @Path("/") annotation

### LINKED ISSUE #1620: ApiListingResource has a class level @Path("/") which can collide with other JAX-RS resources
The current ApiListingResource has a class level @Path("/") annotation. This is causing us a minor annoyance in that we have dependencies which also do the same thing and collide. We are currently getting around this by sub-classing ApiListingResource and redefining the the path to @Path("/swagger") (so we access the swagger json via /swagger/swagger.json).

Looking at the endpoints offered by ApiListingResource I see why this was done. You wanted to support the following URLs for grabbing the swagger config:

/swagger (with an Accepts of application/yaml) -> returns yaml
/swagger (with an Accepts of application/json) -> returns json
/swagger.json (returns json)
/swagger.yaml (returns yaml)

With that being said a @Path("/") is a bit "greedy" and sets the ApiListingResource to collide with other "greedy" resources.

I see two possible solutions:
1. Support only /swagger.json & /swagger.yaml in ApiListingResource. This would allow you have a class level  @Path("/swagger.{type:json|yaml}").
2. Move the /swagger endpoints to a different class with a class level @Path("/swagger"). If a swagger user depends on calling /swagger then they would need to make sure that this new resource is registered with the container (if they are registering the "io.swagger.listing" package then this would be done for them already). 

WDYT? I am willing and able to submit a pull request to help resolve this.

Thanks!

### LINKED ISSUE #1647: Issue WARNing when resource class has @Path("/")
Recently ran into an issue where if my JAX-RS resource class had `@Path("/")` it could conflict with the same annotation on `ApiListingResource` swagger class.  If annotation scanning order happens to be swagger first, all is well and `/swagger.json` returns 200 OK with expected content.

If annotation scanning order happens to be JAX-RS engine first (Apache Wink 1.4 in my case), then my REST API wins the battle of who gets the listen at path `"/"` and GET requests to `/swagger.json` return whatever is appropriate for the _actual_ listener of `"/"` path (405 in my case because I did not have @GET anywhere in my JAX-RS class).

This git issue is a request to output a clear WARNing from swagger.  Something like:

`Swagger detected @Path("/") on one of your JAX-RS resource classes, which may conflict with the same @Path value on ApiListingResource.  Consider changing your @Path annotation value on your resource class or write your own class that extends ApiListingResource so that you can set your own @Path value.`

Helmsdown:
PR for #1620 

fixes #1620, fixes #1647

### DIFF
--- a/modules/swagger-jaxrs/pom.xml
+++ b/modules/swagger-jaxrs/pom.xml
@@ -139,6 +139,6 @@
     <properties>
         <coverage.complexity.minimum>0.50</coverage.complexity.minimum>
         <coverage.line.minimum>0.00</coverage.line.minimum>
-        <coverage.missed.classes>3</coverage.missed.classes>
+        <coverage.missed.classes>4</coverage.missed.classes>
     </properties>
 </project>
--- a/modules/swagger-jaxrs/pom.xml
+++ b/modules/swagger-jaxrs/pom.xml
@@ -56,6 +56,57 @@
                     </instructions>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <version>${failsafe-plugin-version}</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>integration-test</goal>
+                            <goal>verify</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.eclipse.jetty</groupId>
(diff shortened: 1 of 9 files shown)


fehguy:
@Helmsdown it looks good--can you toss in a couple tests?

Helmsdown:
@fehguy Tests added

fehguy:
@frantuma can you check this out?  I think it solves #1620

frantuma:
@fehguy looks good it fixes #1620 and #1647, rebased locally and tested ok. (Added issue "auto close" message in body..)

Issue #1667, opened 2016-02-17
swagger-models 1.5.7 has an incompatible signature-change in io.swagger.models.Property


vivin:
[This commit](https://github.com/swagger-api/swagger-core/commit/a325b1e62aa87f1d9941cb90a091cbda4962194d) changed the signature of `setExample` in `Properties` from `Properties#setExample(String)` to `Properties#setExample(Object)`. This can cause issues at runtime. I noticed this while building swagger-codegen from source and using it in a project that depends on springfox-swagger2, which brings in 1.5.6 of swagger-models as a dependency. 

If the 1.5.6 version ends up getting loaded first, code-generation fails with the following error:

```
java.lang.NoSuchMethodError: io.swagger.models.properties.Property.getExample()Ljava/lang/Object;
	at io.swagger.codegen.DefaultCodegen.toExampleValue(DefaultCodegen.java:601) ~[swagger-codegen-2.1.6-SNAPSHOT.jar:2.1.6-SNAPSHOT]
	at io.swagger.codegen.DefaultCodegen.fromProperty(DefaultCodegen.java:983) ~[swagger-codegen-2.1.6-SNAPSHOT.jar:2.1.6-SNAPSHOT]
	at io.swagger.codegen.DefaultCodegen.addVars(DefaultCodegen.java:1990) ~[swagger-codegen-2.1.6-SNAPSHOT.jar:2.1.6-SNAPSHOT]
	at io.swagger.codegen.DefaultCodegen.fromModel(DefaultCodegen.java:936) ~[swagger-codegen-2.1.6-SNAPSHOT.jar:2.1.6-SNAPSHOT]
	at io.swagger.codegen.languages.JavaClientCodegen.fromModel(JavaClientCodegen.java:510) ~[swagger-codegen-2.1.6-SNAPSHOT.jar:2.1.6-SNAPSHOT]
	at io.swagger.codegen.DefaultGenerator.processModels(DefaultGenerator.java:744) ~[swagger-codegen-2.1.6-SNAPSHOT.jar:2.1.6-SNAPSHOT]
	at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:196) ~[swagger-codegen-2.1.6-SNAPSHOT.jar:2.1.6-SNAPSHOT]
```

If you override the transitive dependency and force 1.5.7, springfox ends up throwing the following error because it cannot find `Property#setExample(String)`:

```
[...]

PR #1668, opened 2016-02-17
issue #1667

vivin:
Removal of `Property#setExample(String)` can cause linkage errors at runtime, if there are conflicting versions of swagger-models. For example, springfox-swagger2 uses swagger-core:1.5.6, which pulls in swagger-models:1.5.6. If swagger-codegen is also in the classpath, `Property#setExample(String)` from the swagger-models:1.5.6 clashes with `Property#setExample(Object)`; overriding the transitive dependency is not an option due to the incompatible signature-change. This change adds the method back to the `Property` class but marks it as deprecated, thereby preserving backwards-compatibility.


vivin:
@fehguy @wing328 This addresses issue #1667. Hope this approach is ok!

Issue #1671, opened 2016-02-22
getDiscriminator not available on Model object


traviscollins:
The swagger 2.0 spec contains a field named discriminator, that is used to define the field name on a "parent" model that is used to determine the subType model name that is actually serialized. The discriminator property is present on ModelImpl, but not on Model. Because discriminator is a part of the spec, and is in the default Impl, it looks like it was left out of the interface unintentionally. This may have been left out intentionally if the purpose was to ensure backwards compatibility. Clearly changing an interface is a big deal - but this particular missing field seems very important. 

Projects that implement clients, such as the swagger-codegen project, need access to the discriminator field to determine how to choose a class into which a serialized object should be inflated. Currently the swagger-codegen project could simply do a check for "is ModelImpl" and then cast and retrieve this value - but that seems uncivilized for a documented property in the official spec.

traviscollins:
Closing because I now see that ArrayModel also extends Model - meaning that discriminator at the Model interface level would be inappropriate.

PR #1672, opened 2016-02-22
Added Model.getDiscriminator() for #1671

traviscollins:
As discussed in the issue #1671, this pull request simply adds a getDiscriminator() method to the Model interface. The discriminator property is already supported by ModelImpl and detailed in the swagger 2.0 spec.


traviscollins:
I see now that ArrayModel extends Model - which means that Model is really just a base class. Closing this pull request in shame.

Issue #1643, opened 2016-02-01
Swagger doesn't allow custom @PATCH annotations with Jersey 1.x


ed0906:
Hi,

As the annotation PATCH is not available in Jersey 1.x we have created our own annotation to do the job:

```
import javax.ws.rs.HttpMethod;
import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

@Target({ElementType.METHOD})
@Retention(RetentionPolicy.RUNTIME)
@HttpMethod("PATCH")
public @interface PATCH {
}
```

This works with Jersey but doesn't seem to be picked up by the `io.swagger.jaxrs.Reader` when used on a resource like below:

```
public interface OurResource {
    @PATCH
    @Produces(MediaType.APPLICATION_JSON)
    @Path("/{id}")
    @ApiOperation(value= "patch")
    public Response patch();
}
```

_Note possibly unimportant but all our annotations are on resource interfaces (this works with all other HTTP methods)_

Is this an issue or is it intended? I know there's an `io.swagger.jaxrs.PATCH` but we'd rather use our own to make sure we don't couple our business logic to Swagger if possible.

webron:
This may be a bug. Looks like the code just looks for a `PATCH` class and doesn't look for a specific one, but it would compare against the first one it picks up, so it'd be a classloading issue. The alternative would be to try and match a string name and not an actual class.

ed0906:
I think it actually specifically looks for the `io.swagger.jaxrs.PATCH` class (see line Reader:891 in v1.5.6). Perhaps it could be modified to search for method annotations that have the HttpMethod annotation themselves?

webron:
It doesn't point to `io.swagger.jaxrs.PATCH` specifically, just a matter of which one is returned first. If you look after that check, you can see that it looks for `HttpMethod` as well, but since it picks up the `PATCH` first, it doesn't help...

fehguy:
@ed0906 you can make your own extension to scan the PATCH resource.  Take a look at how the https://github.com/swagger-api/swagger-core/tree/master/modules/swagger-hibernate-validations project works.

ed0906:
@webron When I'm stepping through the `io.swagger.jaxrs.Reader.extractOperationMethod` using my configuration above the `method.getAnnotation(PATCH.class)` check returns null (as its looking for `io.swagger.jaxrs.PATCH`)

[...]

PR #1679, opened 2016-02-24
Allow user created @PATCH, @GET ...etc annotations

ed0906:
Solution to defect: [#1643](https://github.com/swagger-api/swagger-core/issues/1643).

Currently Javax does not support the PATCH annotation so some users have created their own in the form below to avoid them coupling their code to Swaggers `io.swagger.jaxrs.PATCH`:

```
@Target({ElementType.METHOD})
@HttpMethod("PATCH")
public @interface PATCH{ 
}
```

Which is picked up by Jersey but not by Swagger. This change makes swagger check through all resource method annotations to find the HttpMethod like above.


webron:
@frantuma - can you review this please?

frantuma:
@webron looks ok, it adds additional option as source of http verb for operation.

PR #1680, opened 2016-02-24
Fix #1546 - consumes and produces reading

### LINKED ISSUE #1546: Not parsing @ApiOperation consumes/produces
Referring to issue https://github.com/swagger-api/swagger-core/issues/1448:

Same issue applies to the produces (and also consumes?) property of the **@ApiOperation** annotation:
I am using a comma-separated list of mime-types, and it gets displayed as comma-separated list in Swagger-UI.
Fix looks like just changing
modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java to split the string for api.produces() and api.consumes(), similar to resolution of issue https://github.com/swagger-api/swagger-core/issues/1448:
```
             if (!api.produces().isEmpty()) {
                 produces = new String[]{api.produces()};
             }
```
Btw., I am just trying to use this property as I am not able to control the sort order of mime types. From @Produces JAX-RS annotation, **sort order** is not preserved  (most probably lost already in JAX-RS implementation...). Is there any other mechanism to define the sort order, or at least the default value of the "Response Content Type", as I don't like duplicating information?

### LINKED ISSUE #1609: Charset of @Produces Annotation are  ignored in Accept header
REST WS with charset in the Produce annotations result in wrong swagger.json. The charset is simple ignored.

@Produces("text/plain;charset=UTF-8") results to "produces" : [ "text/plain" ] in the swagger.json.

Hence, Calling the the REST WS by swagger result in HTTP 406

frantuma:
Fix #1546. Fix #1609. It overlaps and replaces https://github.com/swagger-api/swagger-core/pull/1652

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -218,14 +218,10 @@ public class Reader {
             }
 
             if (!api.produces().isEmpty()) {
-                produces = new String[]{api.produces()};
-            } else if (cls.getAnnotation(Produces.class) != null) {
-                produces = ReaderUtils.splitContentValues(cls.getAnnotation(Produces.class).value());
+                produces = ReaderUtils.splitContentValues(new String[] {api.produces()});
             }
             if (!api.consumes().isEmpty()) {
-                consumes = new String[]{api.consumes()};
-            } else if (cls.getAnnotation(Consumes.class) != null) {
-                consumes = ReaderUtils.splitContentValues(cls.getAnnotation(Consumes.class).value());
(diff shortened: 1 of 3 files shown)


webron:
Can you add a test here?

frantuma:
@webron tests added; PR slightly changes behaviour for comma separated “produces” and “consumes” values in` @ApiOperation, @Api, @Produces and @Consumes` annotations by splitting them into array of items fed into operation.produces and consumes; previously this values would be treated as single consume/produce e.g.

`@Produces (“text/xml,text/html”)`

in previous version would result in:

```
"consumes": [
   “text/xml,text/html”
  ]
```
While after  it will generate:
```
"consumes": [
    “text/xml”,
    “text/html”
  ],
```

Issue #1318, opened 2015-08-03
Feature request: @ApiModelProperty example support number type


ShijunK:
I'd like to request a feature which will allow example field of ApiModelProperty support number type. 

For latest release version (1.5.0), if you use example like this:
````
        @ApiModelProperty(dataType = "double", required = true, example = "1234")
	private Double numberField;
````

The generated swagger spec looks like:
````
 "numberField" : {
          "type" : "number",
          "format" : "double",
          "example" : "1349"
        }
````

Which I believe should be:
````
 "numberField" : {
          "type" : "number",
          "format" : "double",
          "example" : 1349
        }
````

**Comment by OscarF** [comment]
Support for booleans would also be nice.

**Comment by rowanhill** [comment]
I think this is a general problem: I would expect the value you put in the example string to be treated as raw JSON, rather than encoded as a string in the swagger.json. It's not type safe, but it would allow numbers, booleans, arrays, even complex objects; strings would be fine, too, but they'd need quotes adding explicitly within the Java string, e.g.:

```java
@ApiModelProperty(example = "\"foo\"")
```

Is that possible?

fehguy:
unfortunately annotations don't support this type of behavior.  We are a bit stuck with the current JLS, unless we decided to put a "meta syntax" in the example field like @rowanhill is proposing.  That could be quite messy to maintain though.

rowanhill:
For what it's worth, [swagger-maven-plugin](http://kongchen.github.io/swagger-maven-plugin/) had [a similar problem](https://github.com/kongchen/swagger-maven-plugin/issues/233) and ended up [allowing raw JSON](https://github.com/kongchen/swagger-maven-plugin/pull/237) like I suggested above.

I'm not quite sure what else you could do. Maybe example could take an Object, which gets JSON serialized to form the example? I imagine that would be a bit cumbersome to use, though.

etyrrill:
I have this problem as well, and I took a stab at fixing it.  I left @ApiModelProperty.example a string, but I changed Property.example to an Object, and wrote custom setters for number and boolean properties.  This relatively minor change fixed the problem for me, and has the benefit of not changing how the annotation is used.

frantuma:
Closing this issue because @etyrrill solution in last comment has been applied in https://github.com/swagger-api/swagger-core/pull/1640; related issues are handled in #1118 and #1681

vinayshankar:
[...]

Issue #1118, opened 2015-06-01
Example values for a list type in Java-jaxrs not working


mahmed0715:
Hello,
We are using swagger-core, jaxrs. everything works good, @example for string works fine , issues seems to be for defining example of a model property that is a list type.

```java
 @ApiModelProperty(required=true, value="Limit to spend without confirmation", example="1000000") 
private Long limit;
```
produces 
```js
{
example : "100000"
}
```
Thats fine as this is a string.

But for
```java
@ApiModelProperty(required=true, value = "xpubs should be account keys", example = "['1','2']")
private List<String> xpubs;
```
procuces
```js
{
example : "['1','2']"
}
```
Thats a mess. 

This can be fixed as overloading the example method I guess.
Waiting this to be fixed.

webron:
@mahmed0715 - not really sure how we can resolve this one easily. If you have any concrete suggestions, feel free to provide them.

webron:
Closing this one in favor of the more generic feature request - https://github.com/swagger-api/swagger-core/issues/1118.

webron:
That was meant to be https://github.com/swagger-api/swagger-core/issues/1318.

dklotz:
@webron So #1318 is closed, but examples still don't work for array/list types right? Should this be re-opened or should I open a new issue?

webron:
I believe #1699 should address that.

vinayshankar:
@webron , @mahmed0715 What exactly is the fix and what do I need to do to use the fix? 
Declaring the example value as `example = "[\"1\",\"2\"]")` does not help. The back slash is present in the documentation. Should I be declaring the value as example = "['1','2']") ? Will this display as 
`
{
example : [ "1" , "2" ]
}
`

Issue #1107, opened 2015-05-27
Example values in containers


santiagopoli:
I have an object defined as this:

``` java
public class CreationDTO {
   @ApiModelProperty(required = true)
   @JsonProperty("users") 
   private Set<Integer> users;
}
```

That object is shown in the UI as:

``` javascript
{
   "users": [0]
}
```

But I want it to be displayed as:

``` javascript
{
   "users": [1,2]
}
```

Is there a way to provide an example value for a container? I know there is an "example" property but it only accepts strings as values.

Thanks!

webron:
There isn't, no. This sounds more of an issue with the UI though. It *shouldn't* show your suggestion, but it should at least indicate it being a set.

santiagopoli:
If i define

```java
@ApiModelProperty(required = true, example = "Example Text")
private String text;
```

The UI shows:

```
{ 
   "text":"Example Test"
}
```

Which is OK. The problem seems to be regarding values inside Lists.

webron:
I see. Well, perhaps we need to modify the example behavior but this is going to be a challenge as a free-form type.

frantuma:
This seems to be related to enhancing model resolver and PropertyDeserializer as mentioned in #1681

JessieGit:
Is this problem solved? I mean, does example support List or other different objects? This is really annoying and frustrating. Our project get stuck here for a long time.....

indiie:
I too am trying to find out if this is supposed to be fixed, and if so could you please provide an example.

rtomsick:
I don't think that this is fixed yet. It sure as hell doesn't work for Maps...

MichakrawSB:
Closing this issue as part of our [backlog cleanup effort](https://github.com/swagger-api/swagger-core/issues/5074).  
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.  
Thank you for your understanding!

PR #1681, opened 2016-02-26
Refs #1107, #1318, #1118 - property example serialization

frantuma:
Refs #1107, #1318, #1118, https://github.com/kongchen/swagger-maven-plugin/issues/233

initially triggered by #1107 analysis, solves possibly regression related to #1640, causing property field "example" not to be serialized to json.

This fix doesn't solve however neither #1107 (swagger-ui related, but probably solved by updating core/models) nor #1118 and it doesn't address similar discussion in https://github.com/kongchen/swagger-maven-plugin/issues/233.

Either part of this PR, or of further work on #1118, the following steps could be taken to cover the whole "area".

1 - Resolving from ApiModelProperty annotation.

similarly to discussion in https://github.com/kongchen/swagger-maven-plugin/issues/233, code could be enhanced to parse example value according to format and property type.

2 - Array type deserialization in PropertyDeserialization.

Code could be enhanced by deserializing additional properties (which should be part of swagger schema if not mistaken) also for array types, e.g. "example".

@webron please let me know what you think

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/properties/Property.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/properties/Property.java
@@ -1,6 +1,8 @@
 package io.swagger.models.properties;
 
+import com.fasterxml.jackson.annotation.JsonGetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonSetter;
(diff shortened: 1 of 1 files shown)


frantuma:
merged to fix #1640 regression, rest of discussion moved to #1699.

PR #1698, opened 2016-03-04
fix for platform dependent tests

kendzi:
The tests are not passing under Windows. It is because of different line ending.


kendzi:
Looks like it was fixed in #1752

Issue #1506, opened 2015-10-20
Path Parameter annotations in interface are parsed as body param


mohitmutha:
When the @PathParam is annotated only to the interface method it is parsed as a body parameter in the swagger.json.

I have a interface for my endpoint which describes the JAX-RS endpoint. I want to have the JAX-RS annotations in the interface and the swagger annotations in the implementation
```java
@Path("/myep")
@Produces(MediaType.APPLICATION_JSON)
public interface MyEndpointInterface{
 @GET
 @Path("/{id}")
public Response getMyData(@PathParam("id") Long id);
}

@Api(value = "/myep")
public class MyEndpointImpl extends MyEndpointInterface{
@ApiOperation(value = "Blah")
@Override
public Response getMyData(Long id){
 //My Impl
}
}

```
In the case above the id parameter is parsed as a body parameter.

I think the reason for this is the way parameter annotations of a method are read in the Reader.java (line 838). 
```java
Annotation[][] paramAnnotations = method.getParameterAnnotations();

```
Because of this only the annotations on the child method get parsed and the ones defined in the interface are not parsed. 

I have created a workaround in ReflectionUtils.java to parse the parameter annotations in the base classes and interface as well. Please let me know if the same is acceptable
```java
//Change in Reader.java
Annotation[][] paramAnnotations = ReflectionUtils.getParameterAnnotations(method);

//Method in ReflectionUtils.java
public static Annotation[][] getParameterAnnotations(Method method) {
  Annotation[][] methodAnnotations = method.getParameterAnnotations();
  Method overriddenmethod = getOverriddenMethod(method);

  if (overriddenmethod != null) {
   Annotation[][] overriddenAnnotations = overriddenmethod.getParameterAnnotations();
  
   
   for (int i = 0; i < methodAnnotations.length; i++) {
     List<Type> types = new ArrayList();
    for(int j = 0; j < methodAnnotations[i].length; j++){
     types.add(methodAnnotations[i][j].annotationType());
    }
    for(int j = 0; j < overriddenAnnotations[i].length; j++){
     if(!types.contains(overriddenAnnotations[i][j].annotationType())){
      methodAnnotations[i] = ArrayUtils.add(methodAnnotations[i], overriddenAnnotations[i][j]);
     }
    }
    
   }
  }
  return methodAnnotations;
 }

webron:
This is using which version?

mohitmutha:
1.5.4-SNAPSHOT

[...]

PR #1735, opened 2016-04-01
Look for parameter annotations on parent classes/interfaces.

captdestrukto:
This is a fix for issue #1506.
Reader got parameter annotations from a child method, ignoring any
on a parent class or interface. This fix will look for annotation
on the parent class or interface, if it exists. Credit mohitmutha
with the fix. I just implemented it and created the pull request.


janhoeve:
I just applied that fix and those kind of JAX-RS resources are working fine now, thanks!

Issue #1740, opened 2016-04-05
Inherited model definition won't get generated unless used in the API or explicitly mentioned


roni-frantchi:
when attempting to inherit form a model not explicitly referenced by any other model explicitly used in the API, the model will not get generated in the definitions.

So for the following:
```
@ApiModel(parent = Parent.class)
public class Child extends Parent {
...
}
```

will only work for me if ```Parent``` is explicitly referenced somewhere in the exposed API.
to workaround it I would have to:
```
@ApiModel(parent = Parent.class)
public class Child extends Parent {
    private Parent dummyReference;
...
}
```
Or perhaps add a dummy API...

arpit8403:
hi Roni-frantchi ,i have the same use case , where Child model schema does not contain parent model . Even after making a dummy reference of Parent in teh child class it does not work . Which swagger-core version i need to use to make it work ,is it swagger-core-1.5.8 or swagger-core-1.5.7?

roni-frantchi:
@arpit8403 my PR here: #1741 fixed the issue for us.
We've been using [our fork](https://github.com/sapienstech/swagger-core) ever since -
I'm waiting for @webron, @frantuma, or @fehguy to approve it.

MichakrawSB:
Closing this issue as part of our [backlog cleanup effort](https://github.com/swagger-api/swagger-core/issues/5074).  
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.  
Thank you for your understanding!

Issue #1737, opened 2016-04-03
`@ApiModel(parent = Pet.class)` seems to have no affect when `subTypes` isn't set


roni-frantchi:
Say one has `Pet.class` with perhaps `Cat.class`
I understand the approach is then to have:
```
@ApiModel(subTypes = {Cat.class})
public class class Pet {
...
}

@ApiModel(parent = Pet.class)
public class Dog {
...
}
```

This works well for me.
In fact, even this would suffice:
```
@ApiModel(subTypes = {Cat.class})
public class class Pet {
...
}

// Note we do not annotate Dog and specify a parent
public class Dog {
...
}
```

But, the following won't work for me:
```
// Note this time we don't annotate Pet with its subtypes
public class class Pet {
...
}

@ApiModel(parent = Pet.class)
public class Dog {
...
}
```

It seems like `io.swagger.jackson.SwaggerAnnotationIntrospector` only cares about the `subTypes` and not the `parent`, which is a problem because, what am I to do if say `Dog.class` is actually a part of some other module and is not visible to the `Pet` module?  I cannot annotate the `Pet` class with its subtypes... and i'm stuck.

fehguy:
How would you expect this to work?  Looking for input on how we scan objects.

roni-frantchi:
@fehguy see my PR here:
https://github.com/swagger-api/swagger-core/issues/1741

There's also a unit test included to demonstrate the behavior.
Let me know if that's what you've been looking for

morpheus9999:
Any news?

rala72:
> Any news?

I guess there are still no news?

Hakky54:
I also would like to have this feature if it is possible. I am currently in a situation where i needed to have a subclass of a specific class. That class comes from an external library so I can't annotate that class with ApiModel and provide the subclass information...

Only thing what I can do is annotating the subclass and provide the property from which parent it is a subclass, but it won't reflect in the generated swagger-ui model section as @roni-frantchi mentioned

shankar96:
Any updates?

shankar96:
> Say one has `Pet.class` with perhaps `Cat.class`
> I understand the approach is then to have:
> 
> ```
> @ApiModel(subTypes = {Cat.class})
> public class class Pet {
> ...
> }
> 
> @ApiModel(parent = Pet.class)
> public class Dog extends Pet {
> ...
> }
> ```
> 
> This works well for me.
> In fact, even this would suffice:
> 
> ```
[...]

Issue #1549, opened 2015-11-20
Definitions using polymorphism seem incorrect


rdiehl:
I'm not sure whether this belongs here, or in the swagger-parser project, or in the swagger2markup project https://swagger2markup.readme.io/, but I'll start here.  We have a number of models that use polymorphism, some of them extend multiple levels.  We're using Java annotations and the JAX-RS integration to create our Swagger definition at runtime.  I'm running into issues running our spec through the swagger2markup tool, and I'm trying to figure out whether the issue is something we have defined incorrectly, or a bug in one of the 3 projects.  Here's our Java classes with the swagger annotations.

``` java
@ApiModel(description = "The representation of a Chart-like object.", discriminator = "_class",
        subTypes = { TextChartDTO.class, ChartDTO.class, TableChartDTO.class })
public abstract class BaseChartDTO implements Serializable {
    private String id;
    private String name;
    public String getId() { return id;}
    public void setId(final String id) {this.id = id;}
    public String getName() { return name;}
    public void setName(final String name) {this.name = name};
    // Various other properties
}

@ApiModel(description = "A parent class for all charts.", parent = BaseChartDTO.class, discriminator = "_class",
        subTypes = { PieChartDTO.class, BarChartDTO.class, AreaChartDTO.class, LineChartDTO.class,
            MixedChartDTO.class, GaugeChartDTO.class })
public abstract class ChartDTO extends BaseChartDTO implements Serializable {
    private ChartType chartType;
    public ChartType getChartType() {return chartType;}
    public void setChartType(final ChartType type) {this.chartType = type;}
    // various other properties
}

@ApiModel(description = "An area chart.", parent = ChartDTO.class)
public class AreaChartDTO extends ChartDTO implements Serializable {
    // various properties
}
```

When our swagger definition is generated, the definition for AreaChartDTO looks like this:

``` json
"AreaChartDTO": {
"allOf": [
{
"$ref": "#/definitions/ChartDTO"
},
{
"type": "object",
"properties": {
  "id": {
    "type": "string"
  },
  "name": {
    "type": "string"
  },
},
.... more stuff
```

Note the existence of "allOf", referencing the ChartDTO definition, and then individual properties like "id" and "name".  The definition for ChartDTO looks like this:

``` json
"ChartDTO": {
"allOf": [
{
"$ref": "#/definitions/BaseChartDTO"
},
{
"type": "object",
"properties": {
  "id": {
    "type": "string"
  },
  "name": {
[...]

PR #1741, opened 2016-04-05
Fixing polymorphism issues

roni-frantchi:
- Fixed: https://github.com/swagger-api/swagger-core/issues/1737
- Fixed: https://github.com/swagger-api/swagger-core/issues/1740
- Fixed: https://github.com/swagger-api/swagger-core/issues/1549

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/converter/ModelConverters.java
+++ b/modules/swagger-core/src/main/java/io/swagger/converter/ModelConverters.java
@@ -1,26 +1,25 @@
 package io.swagger.converter;
 
 import com.fasterxml.jackson.databind.type.TypeFactory;
+import com.google.common.collect.Sets;
 import io.swagger.jackson.ModelResolver;
+import io.swagger.models.ComposedModel;
 import io.swagger.models.Model;
+import io.swagger.models.ModelImpl;
+import io.swagger.models.RefModel;
 import io.swagger.models.properties.Property;
 import io.swagger.util.Json;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Type;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.ServiceLoader;
-import java.util.Set;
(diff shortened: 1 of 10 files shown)


roni-frantchi:
@fehguy ?... anyone...?
It's been almost 6 months now... 
Any reason not to merge?.. Any changes that I should make?..

webron:
Hi @roni-frantchi - the biggest reason is time :) We can possibly review this in about two weeks (definitely not this week), but I can't guarantee that. It looks like this PR needs to be resynced with the code before it can be merged.

webron:
Oh, and Shana Tova.

roni-frantchi:
@webron will get right on resolving those pesky conflicts and push for reviewal

Shana Tova :)

roni-frantchi:
@webron @frantuma @fehguy any news on this one?...

fehguy:
I'm sorry for this taking so long, but this introduces attributes which are illegal in the spec (`parent`) so we can't merge it.

roni-frantchi:
@fehguy wait, what?
First, wasn't 'parent' an attribute that was already available?
Second, isn't the spec extendible so that one can add further metadata?
Third, how about a suggestion of what would be a valid approach?

fehguy:
There is a difference between having a property in the java pojo and in the JSON/YAML description.  This:

https://github.com/sapienstech/swagger-core/blob/2e5df42c7053660c8fc34e4e6cb128a9d4e3661b/modules/swagger-core/src/test/resources/Cat.json#L15

Is not valid in the specification:

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md

And can't be there.  We can have private fields in the pojos, but they cannot live in the JSON generated.

roni-frantchi:
> There is a difference between having a property in the java pojo and in the JSON/YAML description. 
@fehguy that's understood. Looking for recommendation of what fixes I may apply here that would make the approach valid.

roni-frantchi:
@fehguy the spec says:
> While the Swagger Specification tries to accommodate most use cases, additional data can be added to extend the specification at certain points.
>The extensions properties are always prefixed by "x-" and can have any valid JSON format value.

So, if I add that prefix, meaning `parent` would become `x-parent`, that would become a valid approach, right?
Would you like me to open another PR with that change in place?

PR #1758, opened 2016-04-14
Ref #1757: Retain refs from Array/Map Property types when using "removingUnreferencedDefinitions"

### LINKED ISSUE #1757: SwaggerSpecFilter with removingUnreferencedDefinitions, removes definitions within containers (List/Map)
The example condition is at (but there are other lines just like it):
https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/core/filter/SpecFilter.java#L140
```
if (response.getSchema() != null && response.getSchema() instanceof RefProperty) {
```

As an example:
```
public List<Foobar> getAll() {
    ...
}
```
Will filter out the `FooBar` definition, as the operation's response is an `ArrayProperty` which does not extend `RefProperty`.
```
public Foobar getAll() {
    ...
}
```
Retains this property correctly...

mtraynham:
Should fix #1757.  I was unsure of how to get the ref (or refs) from ComposedModel objects, so I left them out.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/core/filter/SpecFilter.java
+++ b/modules/swagger-core/src/main/java/io/swagger/core/filter/SpecFilter.java
@@ -1,6 +1,7 @@
 package io.swagger.core.filter;
 
 import io.swagger.model.ApiDescription;
+import io.swagger.models.ArrayModel;
 import io.swagger.models.Model;
 import io.swagger.models.Operation;
 import io.swagger.models.Path;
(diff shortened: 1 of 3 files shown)


mtraynham:
@fehguy Hey is there something else I need to do to get this accepted?

fehguy:
I haven't been in this piece of code for a bit.  @frantuma can you please check this out and merge if it looks OK?

PR #1759, opened 2016-04-14
Consider java.time.OffsetDateTime a Swagger DateTime primitive

acourtneybrown:


Hades32:
@fehguy why hasn't this been merged yet?

fehguy:
Because there's too much to do :)  I don't just merge without testing, and I should get to this shortly.  Perhaps @frantuma can review faster than me.

Hades32:
Good to know, I already feared there might be valid reasons to refuse it. 

It's really strange how the swagger online editor and cli generate Java8 time types, but the swagger UI doesn't properly support it...

webron:
@Hades32 doesn't support it how?

Hades32:
@webron it's not shown as a timestamp but instead a large json object with many fields. so pretty useless...

fehguy:
@webron I don't see why we wouldn't do this--but we do need to add some tests before merging, or it'll get lost.

webron:
All I was asking is how this is seen in swagger-ui as some people expect it to be something other than a string. It definitely shouldn't appear as a multi-field object.

danny-zegel-zocdoc:
@fehguy @webron any ETA on merging this?

Helmsdown:
+1 for this PR

Helmsdown:
@webron This is how it renders...
image

Helmsdown:
@webron @fehguy if it means anything....I vouch for this change it is almost identical to the change I made to support ZonedDateTime. Unfortunately, you can't add tests because the project compiles for java 6 (assuming that has not changed).

https://github.com/swagger-api/swagger-core/pull/1715

xunnanxu:
+1 for this

dwhu:
+1

PR #1764, opened 2016-04-18
Add extension method to intercept Operation parsed from Jersey resource method

HaloFour:
Adds method to `SwaggerExtension` interface to allow an extension to intercept when a Jersey method is parsed into a Swagger `Operation` model so that the extension can perform custom operations.  The specific use case intended to be addressed is being able to scan custom annotations on the Java method via reflection and then use those annotations to append vendor extensions to the operation.


HaloFour:
Covered by PR #1513

Issue #1201, opened 2015-06-25
swagger-models - ArrayModel and ArrayProperty are missing support for minItems and maxItems


benken-parasoft:
Support for minItems and maxItems appears to be mentioned here in the Swagger 2.0 spec:
https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#schemaObject

However, ArrayModel and ArrayProperty from the swagger-models API appear to be missing support.  I am using swagger-models to marshall Swagger back and forth.  If the JSON Schema in the literal Swagger document defines an array with minItems and maxItems then this information gets lost parsed.  ArrayModel and ArrayProperty have no fields or methods for getting these values.  Marshalling those objects back to JSON (round-trip) results in those minItems and maxItems being lost.

This deficiency causes a problem for any application that wishes to accurately consume Swagger documents and their embedded JSON Schemas.

Here is a JUnit that demonstrates the problem with ArrayModel and ArrayProperty:

<pre><code>import static org.junit.Assert.*;

import com.fasterxml.jackson.core.*;
import com.fasterxml.jackson.databind.*;

import io.swagger.models.*;
import io.swagger.models.properties.*;
import io.swagger.util.*;

import java.io.*;
import java.util.*;

import org.junit.*;

public class SwaggerTest {
    @Test
    public void testMinMaxItems() throws JsonProcessingException, IOException {
        ObjectMapper mapper = Yaml.mapper();
        JsonNode rootNode = mapper.readTree(SAMPLE_SWAGGER);
        assertNotNull(rootNode);
        Swagger doc = mapper.convertValue(rootNode, Swagger.class);
        Map<String, Model> definitions = doc.getDefinitions();
        assertEquals(2, definitions.size());

        // ArrayModel is missing minItems and maxItems?!?
        Model model = definitions.get("SampleArrayModel");
        assertEquals(ArrayModel.class, model.getClass());
        //ArrayModel arrayModel = (ArrayModel) model;
        //assertEquals(2, arrayModel.minItems());
        //assertEquals(5, arrayModel.maxItems());
        assertEquals("{\"type\":\"array\",\"items\":{\"type\":\"string\"}}",
                mapper.valueToTree(model).toString());

        // ArrayProperty is missing minItems and maxItems?!?
        model = definitions.get("SampleArrayProperty");
        assertEquals(ModelImpl.class, model.getClass());
        ModelImpl modelImpl = (ModelImpl) model;
[...]

PR #1767, opened 2016-04-21
Fixing 1201 - array property constraint deserialisation

conorclifford:
Small fix for #1201 (missing deserialisation of ArrayProperty constraints (`minItems`, `maxItems` and `uniqueItems`).

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
+++ b/modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
@@ -1,5 +1,6 @@
 package io.swagger.util;
 
+import com.google.common.collect.Maps;
 import io.swagger.models.Xml;
 import io.swagger.models.properties.ArrayProperty;
 import io.swagger.models.properties.MapProperty;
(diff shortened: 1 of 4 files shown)


fehguy:
@frantuma look OK to you?

cipresso:
Hi @fehguy. I have the same issue described in #1201 on both 1.5.8 and 1.5.9.  I need the fix committed by @conorclifford as I'm deserializing JSON schema that has minItems and maxItems and embedding the resultant Model.

cipresso:
Hi.  I tried out the fix and it works.  There's an unnecessary import statement "import com.google.common.collect.Maps;" you might want to remove.

webron:
@cipresso thanks for validating it.

brando76:
I understand that this pull request is the first step to resolving the issue of maxItems within definitions of not being read by swagger-parser (https://github.com/swagger-api/swagger-parser/issues/239).  Just wanted to mention that this issue is currently causing our internal code generator (which uses swagger-parser/swagger-core) to not handle this case.  Whatever could be done to resolve this we would be most grateful for.

PR #1771, opened 2016-04-25
Fixes #1770 - Set vendor extensions on RefProperty deserialization

### LINKED ISSUE #1770: PropertyDeserializer omits vendor extensions for RefModel properties
Example:
```
definitions:
  pet:
    properties:
      id:
        type: string
  taggedPet:
    properties:
      pet:
        $ref: '#/definitions/pet'
        x-vendor-ext: example
```
x-vendor-ext will be omitted from the resulting `Property` object produced from the deserializer.

nbruno:
Call `AbstractProperty#setVendorExtensionMap` when constructing a `RefProperty`. Added to the existing unit test to validate setting a vendor extension on a `$ref` property.


webron:
So following the comment on the original issue, I don't think we can merge it because it's not really supported.

PR #1772, opened 2016-04-26
Use XmlElementWrapper name when wrapping repeated elements with JAXB XmlElementWrapper annotation

ernaneso:
The following model:

```java
@XmlRootElement(name = "monster")
public class Monster {
  public String name;
  @XmlElementWrapper(name = "children")
  @XmlElement(name = "child")
  public java.util.List<String> child;
}
```

produces the following swagger.json spec:

```json
{
  "properties" : {
    "name" : {
      "type" : "string"
    },
    "children" : {
      "type" : "array",
      "xml" : {
        "name" : "child",
        "wrapped" : true
      },
      "items" : {
        "type" : "string"
      }
    }
  },
  "xml" : {
    "name" : "monster"
  }
}
```

This pullrequest will fix the swagger.json spec to:

```json
{
  "properties" : {
    "name" : {
      "type" : "string"
    },
    "children" : {
      "type" : "array",
      "xml" : {
        "name" : "children",
        "wrapped" : true
      },
      "items" : {
        "type" : "string"
      }
    }
  },
  "xml" : {
    "name" : "monster"
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Swagger Core library
 
+** PS: This version is a fork from https://github.com/swagger-api/swagger-core and fixes the JAXB XmlElementWrapper annotation issue. **
+
 [![Build Status](https://travis-ci.org/swagger-api/swagger-core.svg?branch=master)](https://travis-ci.org/swagger-api/swagger-core)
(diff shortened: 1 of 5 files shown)


fehguy:
OK but the current behavior _is_ the desired behavior.  We don't want the inner name to be the same as the wrapper.

PR #1783, opened 2016-05-08
Fix #1258 - Serialize/deserialize security in "security" instead of "…

### LINKED ISSUE #1258: SecurityRequirements: Mismatch between model & spec

boillodmanuel:
Fix #1258 - Serialize/deserialize security in `"security"` instead of `"securityRequierement"` to be spec compliant.

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/util/JsonDeserializationTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/util/JsonDeserializationTest.java
@@ -51,7 +51,7 @@ public class JsonDeserializationTest {
     }
 
     @Test
-    public void testDeserializeSecurityRequirement() throws Exception {
+    public void testDeserializeSecurity() throws Exception {
         final Swagger swagger = TestUtils.deserializeJsonFileFromClasspath("specFiles/securityDefinitions.json", Swagger.class);
 
         final List<SecurityRequirement> security = swagger.getSecurity();
--- a/modules/swagger-core/src/test/java/io/swagger/util/JsonSerializationTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/util/JsonSerializationTest.java
@@ -64,6 +64,6 @@ public class JsonSerializationTest {
                 .security(new SecurityRequirement().requirement("api_key").requirement("basic_auth"))
                 .security(new SecurityRequirement().requirement("oauth2", Arrays.asList("hello", "world")));
         json = Json.mapper().writeValueAsString(swagger);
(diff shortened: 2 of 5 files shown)


fehguy:
So, I appreciate the PR but this has been discussed before.  You _must_ use our jackson serializers and when you do, the `securityRequirement` is correctly handled.  If you do not, well, you'll have other problems, this being the least of them.

We cannot accept this PR because (a) the serialization strategy works fine as-is and (b) this is a breaking change to a model signature, and this alone is not worthy of bumping to a 2.0 release.

During the 2.0 development we will fix this, so I'll keep your issue #1258

boillodmanuel:
Unless I'm doing something wrong, I'm using your jackson serializer as you
can see in the test case I added in a comment of the original issue.

If I'm not wrong the current seralizer generates invalid swagger 2, doesn't
it ?

fehguy:
I can guarantee you that our serializer is not producing invalid swagger, or all swaggers would be broken.  How are you enabling the serializer?

boillodmanuel:
I'm using `io.swagger.util.Json` like this:

```java
Swagger swagger = new Swagger();
swagger.setSecurity(Arrays.asList(new SecurityRequirement().scope("user").scope("movie")));
System.out.println(Json.mapper().writeValueAsString(swagger));
```
which produces:
```json
{"swagger":"2.0","securityRequirement":[{"scopes":["user","movie"]}]}
```

*(note that I updated the example in this [comment](https://github.com/swagger-api/swagger-core/issues/1258#issuecomment-211440049) because there was a mistake)*

naab:
I want to bump up this PR, because the latest swagger-ui can't handle "securityRequirement" field generated by this generator on the authentication feature. (it expects a "security" field, like it should be !)

as per swagger-codegen, its security definitions only work for the "securityRequirement" field, not "security" as it is based on this code (removing the @JsonIgnore on the setters could fix this tho...)

This is a major issue that should be integrated before everyone forks it to fix it...

fehguy:
I'm a terrible person for not merging this.

PR #1807, opened 2016-05-26
refs #1765 - numeric enums in parameter

### LINKED ISSUE #1765: enums print as strings
The code below shows that the numeric enums are converted to strings when being written as YAML:

```java
        String yaml =
            "swagger: '2.0'\n" +
            "paths:\n" +
            "  /test:\n" +
            "    get:\n" +
            "      parameters:\n" +
            "      - name: \"days\"\n" +
            "        in: \"path\"\n" +
            "        required: true\n" +
            "        type: \"integer\"\n" +
            "        format: \"int32\"\n" +
            "        enum:\n" +
            "        - 1\n" +
            "        - 2\n" +
            "        - 3\n" +
            "        - 4\n" +
            "        - 5\n" +
            "    responses:\n" +
            "      default:\n" +
            "        description: great";

        Swagger swagger = Yaml.mapper().readValue(yaml, Swagger.class);
        Yaml.prettyPrint(swagger);
```

frantuma:
Fixes #1765, changed Parameter enum property to List<Object> to correctly handle numeric values

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
@@ -14,6 +14,7 @@ import io.swagger.models.properties.DecimalProperty;
 import io.swagger.models.properties.Property;
 import io.swagger.models.properties.StringProperty;
 
+import java.util.ArrayList;
 import java.util.List;
 
 @JsonPropertyOrder({ "name", "in", "description", "required", "type", "items", "collectionFormat", "default", "maximum",
@@ -25,7 +26,7 @@ public abstract class AbstractSerializableParameter<T extends AbstractSerializab
     protected String format;
     protected String collectionFormat;
     protected Property items;
-    protected List<String> _enum;
+    protected List<Object> _enum;
     protected Boolean exclusiveMaximum;
     protected Double maximum;
     protected Boolean exclusiveMinimum;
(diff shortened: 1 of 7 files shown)


modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
```diff
@@ -25,7 +26,7 @@
     protected String format;
     protected String collectionFormat;
     protected Property items;
-    protected List<String> _enum;
+    protected List<Object> _enum;
```
tomtit: This change (including getter and setter) is not backward compatible and may affect clients of the `AbstractSerializableParameter` class. I would suggest to apply the same approach that we did for the `default` property (`getDefaultValue` vs `getDefault`, `setDefaultValue` vs `setDefault`)

modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
```diff
@@ -169,7 +170,7 @@ public void setProperty(Property property) {
         this.format = property.getFormat();
         if (property instanceof StringProperty) {
             final StringProperty string = (StringProperty) property;
-            setEnum(string.getEnum());
+            if (string.getEnum() != null) setEnum(new ArrayList<Object>(string.getEnum()));
```
tomtit: The `enum` property from other primitive types has to be taken into account as well (`IntegerProperty`, `LongProperty`, `FloatProperty`, `DoubleProperty`, etc.)

frantuma:
replaced by #1829

PR #1827, opened 2016-06-20
Use null Instead of false When No Default Value is Provided

qtstc:
Code for parseBoolean:
```
   public static boolean parseBoolean(String s) {
        return ((s != null) && s.equalsIgnoreCase("true"));
    }
```
So when the BooleanProperty has no default value (e.g. s == null), we are setting the default value as false.

One of the problems caused by this issue:
https://github.com/kongchen/swagger-maven-plugin/issues/284

### DIFF
--- a/modules/swagger-core/src/test/resources/Animal.json
+++ b/modules/swagger-core/src/test/resources/Animal.json
@@ -54,8 +54,7 @@
                     },
                     "isDomestic": {
                         "type": "boolean",
-                        "position": 3,
-                        "default": false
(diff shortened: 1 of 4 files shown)


fehguy:
@frantuma is this good to merge?

thebeeber:
I could really use this as well. Any update on the merge?

webron:
@thebeeber unfortunately this didn't make the cut for 1.5.10, but we'll try to push it to 1.5.11 (no ETA yet).

Issue #1682, opened 2016-02-26
Unhandled Jackson @JsonUnwrapped annotation


scoro:
Hi,

I have found a functional gap in generating model using Jackson model specification. 
It does not respect @JsonUnwrapped annotation.

It seems that Swagger reads Jackson model as it as. 
Unfortunately Jackson does not expose flatten properties but handles it quite late (during serialization/deserialization), so generated Swagger model is also a tree one despite of @JsonUnwrapped.

Similiar issue has been fixed in other project:
https://github.com/springfox/springfox/issues/192

It would be nice to have this feature working.

mapoulin:
+1

frantuma:
implemented in #1833

PR #1833, opened 2016-06-22
Jackson JsonUnwrapped support

smil2k:
Issue: https://github.com/swagger-api/swagger-core/issues/1682


fehguy:
@frantuma can you see if this is mergable?

PR #1839, opened 2016-06-29
Infer bean param type also from getters

gm2211:
At the moment if you use frameworks like FreeBuilder (https://github.com/google/FreeBuilder) to create POJOs (with builders), you end up with interfaces or abstract classes that only have getters (with builders and concrete classes being automatically generated at compilation time).

Previously that meant that those objects would be ignored by Swagger if used as @BeanParam on jersey resources; this change allows type params and annotations (like @QueryParam) to be picked up from getters.


webron:
@frantuma - can you review this please?

frantuma:
@gm2211 this seems ok, would you mind adding a test demonstrating the use case?

PR #1867, opened 2016-07-26
Make ApiImplicitParams available for annotation types

yaronyam:
When using ApiImplicitParams on a custom annotation you can avoid duplicating repeated parameter definitions.
##### Example:

``` java
@Api(value = "lots of API endpoints here...")
public class LotsOfMethodsController{

  @ApiOperation(value = "Method1")
  @MySwaggerImplicitSecurityHeader
  public void ApiMethod1(){}

  @ApiOperation(value = "Method2")
  @MySwaggerImplicitSecurityHeader
  public void ApiMethod2(){}
  //...with more of these ApiMethods down the line

  @Retention(value = RetentionPolicy.RUNTIME)
  @ApiImplicitParams({@ApiImplicitParam(name = "x-my-custom-auth-header-that-keeps-repeating", required = true, dataType = "string", paramType = "header",defaultValue = "SOME_TEST_TOKEN_VALUE")})
  public @interface MySwaggerImplicitSecurityHeader { }

}
```

### DIFF
--- a/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiImplicitParams.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiImplicitParams.java
@@ -26,7 +26,7 @@ import java.lang.annotation.Target;
  *
  * @see ApiImplicitParam
  */
-@Target(ElementType.METHOD)
+@Target({ElementType.METHOD,ElementType.ANNOTATION_TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface ApiImplicitParams {
     /**


webron:
@frantuma - can you review?

JamieMagee:
Is it possible to get this merged in for `v1.5.11`?

It is a very small change, and would allow for reusable `@ApiImplicitParams` annotations. As far as I can see there isn't currently a way to achieve this.

fehguy:
@frantuma, look OK to you?

JamieMagee:
Actually, testing this locally I don't think this change is enough to allow for reusable annotations. There need to be changes made to either `Reader.java` or `ReflectionUtils.java` in order to correctly find meta-annotations.

espiegel:
I would also like this feature. Having to repeat the same annotation on many methods is annoying.

clementdevos:
This would be a much appreciated feature for us as well, would be very useful for ranges, limits and such.

Issue #1894, opened 2016-08-24
None deterministic behaviour when using service interface inheritance


tonyhart759456:
When generating swagger file from JAX-RS with service inferface inheritance, swagger non deterministically generates different swagger.json on sequential builds. 
For Example:
I have Car a interfaces with a generic type of `<T extends Brand>`.

```
public interface Car <T extends Brand>{
   String isBroken(T car);
}
```

This is extended by an AudiService interface adding the Swagger annotations and adding a brand Audi (which extends Brand).

```
public interface AudiService extends Car<Audi>{

//swagger annotations go here  
  @Override
  String isBroken(Audi car);
}
```

This is then implemented by a AudiServiceImpl.

This means that the swagger spec contains a parameter schema ref for isBroken that is sometimes captured as 
`"$ref" : "#/definitions/Brand"`
and sometimes as 
`"$ref" : "#/definitions/Audi"`

I believe the problem to be in Reader.java at 

```
final javax.ws.rs.Path apiPath = ReflectionUtils.getAnnotation(cls, javax.ws.rs.Path.class);
Method methods[] = cls.getMethods();
for (Method method : methods) {
   if (ReflectionUtils.isOverriddenMethod(method, cls)) {
       continue;
     }
       ....
```

ReflectionUtils.isOverriddenMethod cannot tell if an interface method is overridden, so thus the eventual model is decided by the cls.getMethods(), which has no guaranteed order.

tonyhart759456:
Solution in pull request #1895

iblocks-shaun:
+1

fehguy:
#1895 was merged

PR #1895, opened 2016-08-24
#1894 Fixed Reflection.isOverriddenMethod to detect interface inheritance

tonyhart759456:
Issue 1894 identified an issue where Reflection.isOverriddenMethod could not tell if the method was over ridden in another interface.
This now looks for interfaces & any superclass and then loops through them all, performing the same logic as before


iblocks-shaun:
@fehguy I have merged the master into the branch involved in the pull request. Would it be possible to merge it into master in time for the 1.5.11 release?

PR #1910, opened 2016-09-04
Use In.forValue instead of In.valueOf in SwaggerDefinition reader

kdubb:
Fixes failure when reading an `ApiKeyAuthDefinition` inside a `SwaggerDefinition`


webron:
We don't have a schedule for the next release. You can use the SNAPSHOT in the meantime.

PR #1925, opened 2016-09-12
Add support for meta-annotation processing

### LINKED ISSUE #1899: @ApiImplicitParams should target ANNOTATION_TYPE too
Currently `@ApiImplicitParams` annotation is supported only in `METHOD` context, so that you are able to describe implicit parameters only on top of method declaration. 

When having implicit parameters (such as request headers) common to more than one resource it's really inconvenient to describe the same params for each and every resource along other annotations.

`@ApiImplicitParams` could target also `ANNOTATION_TYPE` so that it will be possible to create combined annotations and reuse them elsewhere.

For example when using Spring MVC

```
@RequestMapping(
            value = "/test"
            produces = MimeTypeUtils.APPLICATION_JSON_VALUE + ";charset=UTF-8",
            method = RequestMethod.GET)
@ResponseBody
@ApiImplicitParams({@ApiImplicitParam(name = "name", value = "desc", required = false, dataType = "string", paramType = "header")})
```

I'd like to have instead custom combined annotation in the following way

```
@RequestMapping(
        produces = MimeTypeUtils.APPLICATION_JSON_VALUE + ";charset=UTF-8",
        method = RequestMethod.GET)
@ResponseBody
@ResponseStatus(HttpStatus.OK)
@ApiImplicitParams({@ApiImplicitParam(name = "name", value = "desc", required = false, dataType = "string", paramType = "header")})
@Target({ElementType.METHOD, ElementType.TYPE})
public @interface Get {

    @AliasFor(annotation = RequestMapping.class, attribute = "value")
    String[] value() default {};

}
```

 So that `@Get` could be used instead of describing `@ApiImplicitParams` explicitly everywhere.

JamieMagee:
A more complete implementation of #1867. Also fixes #1899.

### DIFF
--- a/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiImplicitParams.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiImplicitParams.java
@@ -26,7 +26,7 @@ import java.lang.annotation.Target;
  *
  * @see ApiImplicitParam
  */
-@Target(ElementType.METHOD)
+@Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface ApiImplicitParams {
     /**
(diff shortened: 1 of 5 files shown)


JamieMagee:
Is it possible to have this reviewed for merge?

Issue #1929, opened 2016-09-14
SpecFilter throws NPE on invalid definition reference


fehguy:
A definition like such:

``` json
{
  "swagger": "2.0",
  "info": {
    "description": "This is a sample server Petstore server.  You can find out more about Swagger at <a href=\"http://swagger.io\">http://swagger.io</a> or on irc.freenode.net, #swagger.  For this sample, you can use the api key \"special-key\" to test the authorization filters",
    "version": "1.0.0",
    "title": "Swagger Petstore",
    "termsOfService": "http://swagger.io/terms/",
    "contact": {
      "email": "apiteam@swagger.io"
    }
  },
  "paths": {
    "/pet": {
      "get": {
        "parameters": [],
        "responses": {
          "default": {
            "description": "successful operation",
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/ResponseModel"
              }
            }
          }
        }
      }
    }
  },
  "definitions": {
    "Foo": {
      "type": "object"
    }
  }
}
```

Will cause a NPE in the SpecFilter because there's no value to resolve to.

PR #1931, opened 2016-09-15
SpecFilter throws NPE on invalid definition reference #1929

surajchhetry:
Getting NPE when there is no value to resolve and definitions value is null. So in order to resolve it, I have to put null check condition at SpecFilter at line 174 and allow to locateNestedReferencedDefinitions if it is not null. 

To test it I have created noValueForGetRequest.json at test/java/resources/specFiles and added a test case  shouldNotThroughNPEIfThereIsNoValueToResolve()  to test the condition at SpecFilterTest


dastoikov:
Pull request #1949 should fix your problem as well. Also, it deals with the case where the broken ref is nested within another model.

Issue #1696, opened 2016-03-04
Path.getOperationMap should preserve the insertion order


RobWin:
The method `Path.getOperationMap` should preserve the insertion order.
The `HashMap` should be replaced by a `LinkedHashMap`.

https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/models/Path.java#L170

That way we can render Path Operations in Swagger2Markup as defined in the Swagger file and not randomly.

RobWin:
If you agree to this change, I can create a PR.

webron:
@RobWin can't think of an issue with that. Feel free to submit a PR.

RobWin:
Ok.thx. we have an issue in Swagger2Markup when we want to preserve the order of the user (contract-first design via the editor).

webron:
Sure thing, but in general I advise users to not rely on the order as it is *technically* not preserved. It's represented as a JSON object which by definition does not preserve order. I know some parsers try to preserve it, but in general users shouldn't assume that it's just going to work.

fehguy:
We've changed to use `LinkedHashMap`s

PR #1934, opened 2016-09-17
Preserve order

paukiatwee:
PR for #1696

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/Swagger.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/Swagger.java
@@ -272,6 +272,11 @@ public class Swagger {
         return sorted;
     }
 
+    @JsonIgnore
+    public Map<String, Path> getPathsMap() {
+        return paths;
+    }
+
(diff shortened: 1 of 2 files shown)


modules/swagger-models/src/main/java/io/swagger/models/Swagger.java
```diff
@@ -272,6 +272,11 @@ public void addProduces(String produces) {
         return sorted;
     }
 
+    @JsonIgnore
+    public Map<String, Path> getPathsMap() {
```
fehguy: We already have the `getPaths` to do this

paukiatwee: `getPaths` return `Map` sorted by `key` and this one return original `Map` with insertion order. I can remove it if it is not ok.

fehguy: Yes, I would suggest removing it and opening a new PR to _remove_ the sorting on the existing method.  I see no reason why we need to monkey with the sort order when fetching the paths.

Issue #1595, opened 2016-01-07
Field with @JsonIgnore is shown in model schema


jbrazdil:
I have this class, with ignored field ``ga``, but the field is still shown in the model schema.
```java
@AllArgsConstructor
@EqualsAndHashCode
public class GAV {

    @JsonIgnore
    @NonNull
    private final GA ga;

    @Getter
    @NonNull
    private final String version;

    @JsonCreator
    public GAV(@JsonProperty("groupId") String groupId,
            @JsonProperty("artifactId") String artifactId, @JsonProperty("version") String version) {
        this.ga = new GA(groupId, artifactId);
        this.version = version;
    }

    @JsonIgnore
    public GA getGA() {
        return ga;
    }

    public String getGroupId() {
        return ga.getGroupId();
    }

    public String getArtifactId() {
        return ga.getArtifactId();
    }

    @Override
    public String toString() {
        return getGroupId() + ":" + getArtifactId() + ":" + getVersion();
    }
}
```
```json
{
  "ga": {
    "groupId": "string",
    "artifactId": "string"
  },
  "version": "string",
  "groupId": "string",
  "artifactId": "string"
}
```

webron:
For now you can use `@ApiModelProperty(hidden=true)` as a workaround.

jbrazdil:
I've added ``@ApiModelProperty(hidden=true)`` but the ``ga`` is still shown in the model.

webron:
Odd. Which version of swagger-core do you use?

jbrazdil:
io.swagger:swagger-core:jar:1.5.6:compile

jbrazdil:
I've did some debugging and found that in [ModelResolver#L305](https://github.com/swagger-api/swagger-core/blob/v1.5.6/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java#L305)
```
ApiModelProperty mp = member.getAnnotation(ApiModelProperty.class);
```
mp is null, even when ``_annotations`` in the member is:
```
_annotations = (com.fasterxml.jackson.databind.introspect.AnnotationMap) {
interface org.codehaus.jackson.annotate.JsonIgnore=@org.codehaus.jackson.annotate.JsonIgnore(value=true),
interface io.swagger.annotations.ApiModelProperty=@io.swagger.annotations.ApiModelProperty(reference=, allowableValues=, notes=, access=, hidden=true, dataType=, name=, readOnly=false, position=0, value=, required=false, example=)
[...]

PR #1935, opened 2016-09-18
#1595 added test to verify issue

HugoMario:
This test has been created in order to reproduce issue #1595 but unsuccessfully, the number of model  properties according to the issue should be 7 but the test just get 6, letting us know that "@JsonIgnore" annotation is doing its work properly.


fehguy:
Thanks @HugoMario.  @jbrazdil, I know this is an old issue, but please see what is different between your setup and this test.

Issue #1941, opened 2016-09-22
build fails in tests (dynamically loading classes) due to missing dependencies in swagger-models


DavidBiesack:
clone swagger-core repo and build

mvn -N
mvn install

I get a build failure:

```
Tests run: 288, Failures: 1, Errors: 0, Skipped: 55, Time elapsed: 4.066 sec <<< FAILURE! - in TestSuite
setup(io.swagger.PojosTest)  Time elapsed: 1.189 sec  <<< FAILURE!
java.lang.NoClassDefFoundError: Lorg/joda/time/DateTime;
    at io.swagger.PojosTest.setup(PojosTest.java:62)
Caused by: java.lang.ClassNotFoundException: org.joda.time.DateTime
    at io.swagger.PojosTest.setup(PojosTest.java:62)


Results :

Failed tests: 
io.swagger.PojosTest.setup(io.swagger.PojosTest)
  Run 1: PojosTest.setup:62 ? NoClassDefFound Lorg/joda/time/DateTime;
  Run 2: PASS
  Run 3: PASS


Tests run: 259, Failures: 1, Errors: 0, Skipped: 26

[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] swagger-project ................................... SUCCESS [  2.382 s]
[INFO] swagger-annotations ............................... SUCCESS [  7.357 s]
[INFO] swagger-models .................................... FAILURE [  5.921 s]
```

the swagger-models.

I can add joda-time to modules/swagger-models/pom.xml but then I get another failure:

```
Tests run: 288, Failures: 1, Errors: 0, Skipped: 55, Time elapsed: 4.422 sec <<< FAILURE! - in TestSuite
setup(io.swagger.PojosTest)  Time elapsed: 1.2 sec  <<< FAILURE!
java.lang.NoClassDefFoundError: Lorg/apache/commons/lang3/tuple/Pair;
    at io.swagger.PojosTest.setup(PojosTest.java:62)
Caused by: java.lang.ClassNotFoundException: org.apache.commons.lang3.tuple.Pair
    at io.swagger.PojosTest.setup(PojosTest.java:62)
```

adding apache-commons to swagger-models/pom.xml 

after adding that, the same dynamic class load ( io.swagger.PojosTest.setup ) fails because jackson.databind is missing. adding that dependency resolves the build and I get SUCCESS

I'll submit a PR for this.

webron:
[...]

PR #1942, opened 2016-09-22
add missing swagger-models dependencies: issue #1941

DavidBiesack:


DavidBiesack:
See @webron 's comments on issue #1941 -- the team is unable to reproduce the error, and in fact after a day passed, I'm unable to reproduce the problem either.

PR #1957, opened 2016-09-27
Extracting generic return type's actual runtime type

martingy:
With this change, extracting the actual return type of the generic super class's methods generic return type became possible. Please consider this PR.
If there is a better solution for this, please let me know.


martingy:
Sorry, I missed the java version...

fehguy:
We are using java 1.7, not 1.8.  I think this makes this PR not possible, please reopen if that's not the case and change the java version to 1.7

martingy:
With my last commit, I changed the version back to 1.6, and travis succeeded as well.

PR #1962, opened 2016-10-05
Handle complex types in the schema.

ewanmellor:
A type declaration may be an array of strings, as opposed to a single string.
An object matches this schema if it matches any one of the named types.

Handle this by mapping the type schema to an appropriate simple type for
the property to use, e.g. by mapping ["string", "null"] to "string" and
["string", "bool"] to "object".

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
+++ b/modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
@@ -168,7 +168,7 @@ public class PropertyDeserializer extends JsonDeserializer<Property> {
     }
 
     Property propertyFromNode(JsonNode node) {
-        final String type = getString(node, PropertyBuilder.PropertyId.TYPE);
+        final String type = getType(node);
         final String title = getString(node, PropertyBuilder.PropertyId.TITLE);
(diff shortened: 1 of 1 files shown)


webron:
Actually, the spec doesn't allow defining multiple types (that is, only a single type and not an array of types), so I'm afraid we can't merge this. We appreciate the effort though.

ewanmellor:
@webron Here's the specification wording that I found:

http://swagger.io/specification/#schemaObject:

> Further information about the properties can be found in JSON Schema Core and JSON Schema Validation. Unless stated otherwise, the property definitions follow the JSON Schema specification as referenced here.
>
> The following properties are taken directly from the JSON Schema definition and follow the same specifications:
>
> [Bulleted list of properties]
> * type

http://json-schema.org/latest/json-schema-validation.html:

> 5.5.2.  type
> 5.5.2.1.  Valid values
> The value of this keyword MUST be either a string or an array. If it is an array, elements of the array MUST be strings and MUST be unique.

That seems pretty clear to me.  A schema object contains a property called type, which is by definition the same as the JSON Schema one, and the JSON Schema validation allows it to be an array.  Am I missing something?

webron:
If it helps, I wrote it, and made a mistake in the description of it. It really is not supported.

PR #1980, opened 2016-10-28
Use varargs instead of arrays when possible

ebourg:
Hi,

I'd like to suggest this minor modification to use varargs instead of arrays. This change is binary compatible with the previous releases and makes the API slightly easier to use (`setSchemes("http", "https")` instead of `setSchemes(new String[] {"http", "https"})`).

Thank you

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/util/BaseReaderUtils.java
+++ b/modules/swagger-core/src/main/java/io/swagger/util/BaseReaderUtils.java
@@ -23,7 +23,7 @@ public final class BaseReaderUtils {
      * @param extensions is an array of extensions
      * @return the map with extensions
      */
-    public static Map<String, Object> parseExtensions(Extension[] extensions) {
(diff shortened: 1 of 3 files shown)


fehguy:
Hi, this is more of a style change and doesn't affect the functionality.  I'm closing as not required

ebourg:
Well, it affects the usability of the API though. This change is rather harmless, it's backward compatible and saves a few characters for the developers. Only pull requests fixing actual bugs are accepted?

fehguy:
Hi, it's a style issue.  If we merged every variation in style of code, we'd have things flip-flopping every day and nothing would get released.  I agree that it may be easier for some to read, but there are probably the same number who prefer the array style.

ebourg:
Sorry I don't get it, the change proposed also works with arrays, you can use both syntaxes:

    setSchemes(new String[] {"http", "https"})

and

    setSchemes("http", "https")

You can only make more developers happy with this.

PR #1981, opened 2016-10-28
#1319 Error log when dataType is "file" in @ApiImplicitParam (swagger-jaxrs)

msiemczyk:
added file as a primitive type


msiemczyk:
Can this be merged and released in 1.5.11 please? We have some continuous integration tests that are now failing because this bug is logging an error in our logs.

dklotz:
maybe ping @fehguy will help? this is intended as a fix for #1319, which was erroneously closed

msiemczyk:
@dklotz Thanks, I emailed him.

fehguy:
@msiemczyk @dklotz sorry for the lag.  Yes, this should be fine.

PR #1984, opened 2016-11-01
#1983 - allow non-strings in x-examples field inside body paramers

### LINKED ISSUE #1983: Non-String x-examples
We have some swagger documents that define x-examples fields as objects and not strings in body parameters.  Seems that the code merged to resolve #1500 adds specific handling for x-examples field and assumes the value is a string.  We cannot parse our swagger files (which are valid according to the spec since x-* values should be allowed to be any type).  They blow up with a deserialization error like below:

```
java.lang.IllegalArgumentException: Can not deserialize instance of java.lang.String out of START_OBJECT token
 at [Source: N/A; line: -1, column: -1] (through reference chain: io.swagger.models.parameters.BodyParameter["x-examples"]->java.util.LinkedHashMap["application/json"])
```

Would be useful to allow x-examples to be any type.

gilbode:
Fixes #1983 and adds support for non-string values for x-examples field inside body parameters.

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/deserialization/JsonDeserializationTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/deserialization/JsonDeserializationTest.java
@@ -4,6 +4,8 @@ import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertTrue;
 
 import io.swagger.models.Swagger;
+import io.swagger.models.parameters.BodyParameter;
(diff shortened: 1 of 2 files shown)


fehguy:
While I agree that objects in the extensions makes sense, jamming the `example` object into them isn't right.

gilbode:
@fehguy - any suggestions on a better approach?

fehguy:
seems like having an `examples` map with object is a simple alternative.

PR #2004, opened 2016-11-17
Add parameter 'replace' to annotation @ApiImplicitParam

mikkoi:
To make it possible to completely redefine a normal query/body/path/header parameter by replacing an existing parameter (when creating swagger.json/yaml). Example:

```java
@ApiImplicitParams({
        @ApiImplicitParam(name = "query", value = "A query in the form of JSON serialized Query.class",
                required = true, dataType = "Query", paramType = "query",
                example = "{\"max_decimals\":-1,\"page\":1,\"pageSize\":20,\"version\":false,\"store\":\"query-resource-test-store1\"}",
                replace = "query"
        )
})
public List<Map<String,Object>> queryGetV2(
        @ApiParam(value = "A Query object serialized as JSON.", name = "query", required = true, hidden = false)
        @QueryParam(value = "query")
        final Query query
) { }
```

Signed-off-by: Mikko Johannes Koivunalho <mikko.koivunalho@iki.fi>

### DIFF
--- a/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiImplicitParam.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiImplicitParam.java
@@ -111,6 +111,11 @@ public @interface ApiImplicitParam {
     String paramType() default "";
 
     /**
+     * The name of another parameter to replace, such as a parameter specified with JAX-RX annotation @QueryParam.
+     */
+    String replace() default "";
+
+    /**
      * a single example for non-body type parameters
      *
      * @since 1.5.4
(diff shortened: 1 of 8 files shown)


webron:
The example is a bit confusing. I imagine it replaces `query` as the parameter name and not the parameter type?

Also, this would be equivalent to creating the `@ApiImplicitParam` and setting the `@QueryParam` as `hidden=true`?

mikkoi:
Yes, the example does that: replaces the parameter "query" with another definition (including datatype). The parameter type stays same (query, not body/header/etc.).

I think you are right. We get the same result by setting `hidden=true`. I had not considered that. I wonder if `replace=*` would be any clearer...

mikkoi:
The suggested functionality can be achieved by other ways already implemented.

PR #2015, opened 2016-11-28
Handle non ref property types when removing unused defs

Mikosz91:


Mikosz91:
This is a simple null check I wrote because I was running into NPEs when using `removeBrokenReferenceDefinitions()` because a few of the body parameters for resources I have documented are a list of strings `@ApiOperation(response = String.class, responseContainer = "List")` which is of type StringProperty.

These are resolved in `getPropertyRef()` as null which eventually causes the filter to try and add null values to a TreeSet of referenced definitions.

PR #2027, opened 2016-12-08
fixed so example in array property response schema is correctly dese…

olensmar:
…rialized

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
+++ b/modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
@@ -1,25 +1,5 @@
 package io.swagger.util;
 
-import io.swagger.models.Xml;
-import io.swagger.models.properties.ArrayProperty;
-import io.swagger.models.properties.MapProperty;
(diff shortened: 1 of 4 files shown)


modules/swagger-core/src/main/java/io/swagger/util/PropertyDeserializer.java
```diff
@@ -250,6 +248,11 @@ Property propertyFromNode(JsonNode node) {
                 arrayProperty.setMinItems(getInteger(node, PropertyBuilder.PropertyId.MIN_ITEMS));
                 arrayProperty.setMaxItems(getInteger(node, PropertyBuilder.PropertyId.MAX_ITEMS));
                 arrayProperty.setUniqueItems(getBoolean(node, PropertyBuilder.PropertyId.UNIQUE_ITEMS));
+                JsonNode example = getDetailNode( node, PropertyBuilder.PropertyId.EXAMPLE);
+                if( example != null ) {
+                    arrayProperty.setExample( Json.mapper().convertValue(example, Object.class));
```
olensmar: not this is the right type of value to set as the example; we could also simply save the JsonNode!?

PR #2034, opened 2016-12-19
added format override

### LINKED ISSUE #469: Requirement: hidden password for APiParam
Not sure currently swagger can support to hidden the password in plaintext.
Currently, we 're using swagger-annotations_2.10. For ApiParam definition, we don't see any field to support hidden param value in plaintext.

fehguy:
Fixes #469

### DIFF
--- a/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiParam.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/annotations/ApiParam.java
@@ -110,4 +110,13 @@ public @interface ApiParam {
      * @return
      */
     Example examples() default @Example(value = @ExampleProperty(mediaType = "", value = ""));
+
+    /**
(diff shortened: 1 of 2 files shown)


andreas-fey:
So, how does this fix showing a password in swagger as **** ?

fehguy:
It's part of the solution to let the format be `password`.  Now the front-end can choose to hide the input value

fehguy:
Simply set `type="string", format="password"` and swagger-ui will render the password correctly.

andreas-fey:
Seems that the "format" attribute did it not make into the https://github.com/swagger-api/swagger-core/wiki/Annotations-1.5.X#apimodelproperty , did it?

fehguy:
No, the original issue was around params, not model properties.  Please open a new issue if you're looking at model property support.

PR #2039, opened 2016-12-20
remove unused jackson dependencies from swagger-jaxrs

pjfanning:
swagger-jaxrs has a number of jackson dependencies in its pom.xml but it appears it only really needs jackson-databind.
https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2016-3720 is open against jackson-dataformat-xml and this fix removes that dependency from swagger-jaxrs

### DIFF
--- a/modules/swagger-jaxrs/pom.xml
+++ b/modules/swagger-jaxrs/pom.xml
@@ -103,13 +103,8 @@
     </build>
     <dependencies>
         <dependency>
-            <groupId>com.fasterxml.jackson.dataformat</groupId>
-            <artifactId>jackson-dataformat-yaml</artifactId>
-            <version>${jackson-version}</version>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.dataformat</groupId>
-            <artifactId>jackson-dataformat-xml</artifactId>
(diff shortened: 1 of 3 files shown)


modules/swagger-jaxrs/pom.xml
```diff
@@ -103,13 +103,8 @@
     </build>
     <dependencies>
         <dependency>
-            <groupId>com.fasterxml.jackson.dataformat</groupId>
-            <artifactId>jackson-dataformat-yaml</artifactId>
```
fehguy: I believe this is needed by the yaml listing provider: https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/listing/BaseApiListingResource.java#L135

pjfanning: swagger-core module still has this dependency so this will be inherited transitively by swagger-jaxrs. If you prefer, I can add back the depency explicitly here too. It's more useful to remove the other 2 dependencies (jackson-jaxrs-json-provider and jackson-dataformat-xml)

fehguy: Got it

Issue #2047, opened 2016-12-21
Issue with ModelResolver. Not producing the correct json representation for attributes annotated with XmlElementWrapper annotation


ritesh-garg:
Unless I am totally missing something, there is a bug in the model resolution logic in ModelResolver, JAXBAnnotationsHelper java files.

According to [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#xml-object); to represent an xml structure like below:

```
<animals>
    <animal>value</animal>
    <animal>value</animal>
    <animal>value</animal>
</animals>
```

for a java entity as:
```
{
.
.
        @XmlElementWrapper(name = "animals")
        @XmlElement(name = "animal")
        public List<String> animals;
.
.
 }
```

the following definition can be used:

```
{
   .
   .
      "animals":{
         "type":"array",
         "xml":{
            "wrapped":true
         },
         "items":{
            "type":"string"
            "xml":{
              "name":"animal"
            }
         }
      }
  .
  .
    }
```
    
However, the definition generated by Swagger is:

```
    {
      .
      .
      "animals":{
         "type":"array",
         "xml":{
            "name":"animal",
            "wrapped":true
         },
         "items":{
            "type":"string" 
         }
      }
      .
      .
    }
```

This definition is going to result in the following xml:

```
    <animal>
        <animal>value</animal>
        <animal>value</animal>
        <animal>value</animal>
    </animal>
```
 
Key points:

1.  Xml name for the wrapper element should be set to "animals"
2. Xml name for the items should be set to "animal" in order to generate the xml tag for the array elements as "animal"

I checked the code in the test class XMLInfoTest.java to cross-check this and it seems like the test is also expecting the element at wrong place. Please let me know if more details are needed.

fehguy:
PR merged

PR #2050, opened 2016-12-23
Issue 2047

ritesh-garg:
Creating pull request for fix for [Swagger Core Issue 2047](https://github.com/swagger-api/swagger-core/issues/2047)

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/jackson/XMLInfoTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/jackson/XMLInfoTest.java
@@ -7,6 +7,7 @@ import io.swagger.converter.ModelConverterContextImpl;
 import io.swagger.models.Model;
 import io.swagger.models.ModelImpl;
 import io.swagger.models.Xml;
+import io.swagger.models.properties.ArrayProperty;
 import io.swagger.models.properties.Property;
 import java.util.List;
 import javax.xml.bind.annotation.XmlAccessType;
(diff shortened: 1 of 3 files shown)


modules/swagger-core/src/test/java/io/swagger/jackson/XMLInfoTest.java
```diff
@@ -34,14 +35,19 @@ public void testSimple() throws Exception {
...
         final Xml propertyXml = property.getXml();
-
         assertNotNull(propertyXml);
-        assertEquals(propertyXml.getName(), "item");
+        // Check the name of array property
+        assertEquals(propertyXml.getName(), "items");
         assertTrue(propertyXml.getWrapped());
-
+        // Get the xml for items for the array property
+        final Xml itemsXml = property.getItems().getXml();
```
fehguy: @webron is this correct?  That the _wrapped_ value would take the name `items` and the XML property itself would be `item`?  I believe it's the opposite, but I've been staying far away from XML-land.

webron: In `XmlDecoratedBean` the `items` should be represented eventually as:

```yaml
items:
  type: array
  items:
    type: string
    xml:
      name: item
  xml:
    name: items
    wrapped: true
```

Two notes:
1. the `name: items` is not required in this case, because it's already the name of the property in any case. We could add a check if the value of the `@XmlElementWrapper` is the same as the field name, then not add the `name` to the `xml` object.
2. The example itself is confusing to a reader because it uses the name `items` for an array, which in JSON Schema uses the property `items` - just making it difficult to follow.

ritesh-garg: I agree with both the points. I did not change them earlier because they were part of the existing code. Have made changes now. Thanks!

Issue #1564, opened 2015-12-09
ApiImplicitParam should support class as parameter


DerekYangYC:
Currently we know it's important to set `name`, `dataType` and `paramType` for `ApiImplicitParam` just like the example:

```java
@ApiImplicitParam(name = "name", value = "User's name", required = true, dataType = "string", paramType = "query")
```

However, if this implicit parameter is needed for several REST api, it has to be copied and pasted several times. Worse case is when you need to change it, you have to change it several times as well. Instead, it could be simplified by supporting class as parameter like this:

```java
@ApiImplicitParam(class=User.class)
```

And User.class could provide necessary information like `name`, `value` etc.

fehguy:
It would be convenient, to do something like this, yes

pjfanning:
https://github.com/swagger-api/swagger-core/issues/2068 has been merged so this can probbaly be closed

MichakrawSB:
Closing this issue as part of our [backlog cleanup effort](https://github.com/swagger-api/swagger-core/issues/5074).  
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.  
Thank you for your understanding!

PR #2068, opened 2016-12-29
support @ApiImplicitParam dataTypeClass

pjfanning:
https://github.com/swagger-api/swagger-core/issues/1564

I'm sure there is a bit more needed here but it would be nice to support classes directly like in other swagger annotations. In my dev team, the most common error with swagger is to not properly fully reference the class names in @ApiImplicitParam dataType.


fehguy:
I really hate the `dataTypeClass` annotation property but I do understand that it is necessary to avoid breaking the API.

pjfanning:
@fehguy I'm not wedded to the annotation name, I'm not that happy with it either. If you can think of a better name, then feel free to change it.

fehguy:
Yea, I don't have a better solution right now, @pjfanning.  Not without breaking the signature

PR #2077, opened 2017-01-10
Fixes 1107 - allow non-strings in x-examples field inside body parameters

gilbode:
Fixes 1107, originally a fix for 1983.  This simply makes x-examples Map<String,Object> instead of Map<String, String>.  Unfortunately this does break API compatibility in BodyParam.   The original pull request for 1983, preserved API compatibility of BodyParam (i.e. examples remained Map<String,String>) however that approach was rejected.  Let me know if these API changes are OK, unclear of the potential downstream impact.


fehguy:
At this stage, we are not going to change the model implementation signatures but this will be supported in the 3.0.0 support, which is in the feature/3.0.0-rc1 branch.

PR #2098, opened 2017-01-24
Extracting generic return type's actual runtime type

martingy:
With this change, extracting the actual return type of the generic super class's methods generic return type became possible.
The java version is 1.6.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -18,21 +18,8 @@ package io.swagger.jaxrs;
 
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.type.TypeFactory;
-import io.swagger.annotations.Api;
-import io.swagger.annotations.ApiImplicitParam;
-import io.swagger.annotations.ApiImplicitParams;
-import io.swagger.annotations.ApiKeyAuthDefinition;
-import io.swagger.annotations.ApiOperation;
-import io.swagger.annotations.ApiResponse;
-import io.swagger.annotations.ApiResponses;
-import io.swagger.annotations.Authorization;
-import io.swagger.annotations.AuthorizationScope;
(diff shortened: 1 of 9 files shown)


fehguy:
@martingy thank you.  What affect does bringing `classmate` into the picture have on backwards compatibility with Jackson?

martingy:
@fehguy no affect on that as far as I can work out. classmate is another library from fasterxml, it is independent of Jackson.

fehguy:
@martingy could you please rebase?

modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
```diff
@@ -823,7 +817,22 @@ private Operation parseMethod(Class<?> cls, Method method, List<Parameter> globa
...
+                    && resolvedMethod != null && resolvedMethod.getReturnType() != null
+                    && resolvedMethod.getReturnType() instanceof ResolvedObjectType) {
+                responseType = ((ResolvedObjectType) resolvedMethod.getReturnType()).getErasedType();
+            } else if (method.getGenericReturnType() != null && method.getGenericReturnType() instanceof GenericArrayType
+                    && resolvedMethod != null && resolvedMethod.getReturnType() != null
+                    && resolvedMethod.getReturnType() instanceof ResolvedArrayType) {
+                responseType = ((ResolvedArrayType) resolvedMethod.getReturnType()).getErasedType();
+            }
+
+            LOGGER.debug("response class of method " + method + " is " + responseType);
```
pjfanning: Could you use`LOGGER.debug("response class of method {} is {}", method, responseType`? https://www.slf4j.org/faq.html#logging_performance

pom.xml
```diff
@@ -512,5 +512,7 @@
         <commons-io-version>2.4</commons-io-version>
         <enforcer-plugin-version>1.4</enforcer-plugin-version>
         <failsafe-plugin-version>2.19.1</failsafe-plugin-version>
+
+        <java.version>1.8</java.version>
```
pjfanning: I'm not sure that removing support for JRE 1.6 and JRE 1.7 should be done in a minor release.

PR #2105, opened 2017-02-06
Do not serialize null values in path

tobilarscheid:
Hi everyone,

when trying to set up swagger with dropwizard, I stumbled over [this line](https://github.com/swagger-api/swagger-samples/blob/master/java/java-dropwizard/src/main/java/io/swagger/sample/SwaggerSampleApplication.java#L31). I do understand that it is required by the [spec](https://github.com/swagger-api/swagger-js/issues/268) to not serialize null values to the swagger.json file. I do however not really want to globally disable null serialization for my whole application.

I therefore feel that this should directly be annotated at the `io.swagger.models.Path` class instead of requiring users to change their whole application's config. Please tell me if there is more classes  where this change might make sense, I am happy to change them as well.

I ran all tests locally and they passed.

Regards,

Tobi

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/Path.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/Path.java
@@ -9,6 +9,7 @@ import java.util.List;
 import java.util.Map;
 
 @JsonPropertyOrder({"get", "head", "post", "put", "delete", "options", "patch"})
+@JsonInclude(JsonInclude.Include.NON_NULL)
 public class Path {
 
     private Map<String, Object> vendorExtensions = new LinkedHashMap<String, Object>();


tobilarscheid:
This is strange, the build seems to be stalled or something, can someone please check what's wrong?

fehguy:
Travis gets mad sometimes, I've restarted it.

fehguy:
@tobilarscheid understood--and I agree that what is used for Swagger should not need to apply to the entire application.

I believe the code to serialize the Swagger POJO into JSON [lives here](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/listing/BaseApiListingResource.java#L159).  The issue is, the swagger specification does not allow _null_ anywhere in it--that means not serializing `null` in the Path is only part of the problem.  There are many other locations that may have a null value, which will make any consumer of the spec very unhappy if null appears...

Probably the best solution would be to modify the `BaseApiListingResource` to call the `Json.mapper()` explicitly--and ensure that that class (which also lives in swagger-core) has the NON_NULL setting in it.  Then, you don't care what the global mapper says, we only write with the swagger-specific Json mapper.

Make sense?

tobilarscheid:
Hi @fehguy, I like your idea. The problem is that the serialization does not really happen in the class you linked to - this only generates the jax-rs response which is then serialized into JSON by jackson. I really like that this is the absolute java standard way and I feel we should not change this.
I know that annotating everything with @JsonInclude (like I did for the path) is a lot of work, it is however consistent with how @Json-_something_ annotations are used in the model - they are all over the place already. Also, when looking closer at the 'swagger-models' module and all classes under 'io.swagger.models' it is probably about 10 classes we would actually need to touch - everything else is then dealt with by inheritance.

tobilarscheid:
sorry, accidentally closed

fehguy:
But @tobilarscheid we can "take over" the serialization process here and return the string rather than letting JAXRS choose what it thinks is best.  The problem will become that there are other settings that will break the process--for example you may have an underscore naming rule.  Well that's great, but it'll bust the swagger definition, because it is designed as camelCase.

tobilarscheid:
Well, you are right. It may not be most beautiful from the JAX-RS standard point of view, but it sure is the safest solution. If you don't mind I will provide another PR with a customized object mapper used in the `BaseApiListingResource`. Thanks for your feedback!

fehguy:
Sounds good.  Consider using the `io.swagger.util.Json` class in that PR!

PR #2106, opened 2017-02-06
use custom object mapper to serialize json in order to avoid null values

tobilarscheid:
As discussed in #2105 

I decided to let the `JsonProcessingException` just go all the way up, as this is closest to what would happen if a JsonProcessingException was thrown when Jackson tries to serialize the object.


fehguy:
LGTM

tobilarscheid:
Thank you! I will open a corresponding PR in the swagger-samples repo to update the example.

tobilarscheid:
https://github.com/swagger-api/swagger-samples/pull/71 there it is

PR #2126, opened 2017-02-28
Break for loop once a matching package is found

yinonavraham:


fehguy:
I'm happy to merge this, but did it create an issue?  If so, could you please provide a test case that shows it?  Will help the project to document such cases.

yinonavraham:
There is no issue or test to write - the code is correct. 
This is a simple slight performance improvement - there is no need to continue checking acceptable packages if we already know it is acceptable (since we found at least one matching).

fehguy:
Thanks

Issue #2127, opened 2017-03-02
BeanConfig.java line#302: startsWith check is maybe a bit too loose


KeldOelykke:
I am trying to debug why I get the same swagger.json from 3 different servlets and stumbled on this tiny issue.

In BeanConfig (https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/config/BeanConfig.java) method classes() there is a check only to consider classes part of the bean configured list of packages:

```
                for (String pkg : acceptablePackages) {
                    if (cls.getPackage().getName().startsWith(pkg)) {
                        output.add(cls);
                    }
                }
```
Unfortunately, this will add classes from packages that are siblings to an acceptable package.
Example:
If we have class "a.b.c.S1" and class "a.b.cc.S2" - each with their respective BeanConfig's - both S1 and S2 will be considered for the BeanConfig of S1, since "a.b.cc" startsWith "a.b.c".

One possible solution to this would be to add a dot in the check e.g.
```
                for (String pkg : acceptablePackages) {
                    if (cls.getPackage().getName().startsWith(pkg + ".")) {
                        output.add(cls);
                    }
                }
```
This is ofc assuming that BeanConfig is configured with whole package parts e.g. "com.ajax" cannot be accepted with "com.aja" anymore.

Kind Regards,
Keld Ølykke

fehguy:
Hi Keld, that does seem like an issue.  Could you please send a pull request?  I don't think `startsWith` is doing what was intended.

KeldOelykke:
Sure, i'll try to make 1 in the weekend.

KeldOelykke:
According to test shouldDeepScanPakagesPer1011() in https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs/src/test/java/io/swagger/BeanConfigTest.java all classes within the list of packages should be allowed.
I guess I will just stick with the startsWith check, but ensure it is viable packages that are tested against each others.

KeldOelykke:
Btw. after this fix here is how I ended up getting the 3 different servlets presented correctly with 3 different swagger.json. From searching the web I am not the only one unable to get multiple servlets to work with swagger in an embedded jetty :)

In jetty setup:
```
...
        ArrayList<Handler> tempHandlerList = new ArrayList<Handler>();

        boolean enableFeBe = true;
        if(enableFeBe)
        {
[...]

PR #2128, opened 2017-03-05
BeanConfig classes() now shouldn't return classes outside specified l…

KeldOelykke:
…ist of packages (https://github.com/swagger-api/swagger-core/issues/2127).


KeldOelykke:
Maybe. The issue is #2127. How do I "add" the issue... except from linking to it as above?

Yes. I can make a test case that will break without the fix.

fehguy:
@KeldOelykke do you have a test that we can add to this PR?

KeldOelykke:
@fehguy Yes. I have added a new test resource 'org.myles' and a new test case in BeanConfigTest that tests that 'org.my' now doesn't include 'org.myles'. Without the fix that wouldn't be possible. Cheers!

PR #2152, opened 2017-03-27
Bump versions

msmerc:
Where I work we're strict about maven version conflicts. Would there be any problem with bumping some of the swagger dependencies.

Full disclosure: These versions have been chosen to line up with dropwizard 1.1.0. The reflections one is the main one to update as the current version pulls in a defunct google annotations jar, the others are just nice-to-have.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/util/Json.java
+++ b/modules/swagger-core/src/main/java/io/swagger/util/Json.java
@@ -14,7 +14,7 @@ public class Json {
         }
         return mapper;
     }
-    public static ObjectWriter pretty() {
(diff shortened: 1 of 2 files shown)


fehguy:
@webron any concerns about updating the reflections library?

webron:
Yes, but only because there's no change log for the library at all, so can't really tell what changed and if there are any breaking changes  (yeah, yeah, semver).

It looks like our tests pass though... so let's hope for the best and go with it.

fehguy:
@webron, `semver` is an aspiration endpoint, nothing more.

PR #2179, opened 2017-04-25
2061 Possible Solution And Successful Tests

ralphdoe:
A Bugfix was added, and also a new test for the solution

The solution was implemented in the SwaggerAnnotationIntrospector adding a new validation in the hasIgnoreMarker method, when there's a Hidden field and the constructor is treated with a JsonCreated it mark the Constructor as ignored. So, when it try to remove the property it doesn't have any trouble to do it.

### DIFF
--- a/modules/swagger-jersey2-jaxrs/pom.xml
+++ b/modules/swagger-jersey2-jaxrs/pom.xml
@@ -85,11 +85,19 @@
             <artifactId>testng</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>${lombok.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <properties>
         <!-- TODO increase coverage and remove these overrides-->
         <coverage.complexity.minimum>0.50</coverage.complexity.minimum>
         <coverage.line.minimum>0.00</coverage.line.minimum>
         <coverage.missed.classes>2</coverage.missed.classes>
+        <lombok.version>1.16.16</lombok.version>
+
     </properties>
 </project>


modules/swagger-jersey2-jaxrs/src/test/java/io/swagger/models/HiddenFieldTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Map;
+
+import static org.testng.Assert.*;
+
+public class HiddenFieldTest {
+
+    private final static int NUMBER_OF_PROPERTIES = 2;
+
+    @Test(description = "it shouldn't ignore a hidden field")
```
fehguy: I think you mean "it should ignore a hidden field"?

modules/swagger-jersey2-jaxrs/src/test/java/io/swagger/models/HiddenFieldTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertTrue(nameValue instanceof StringProperty);
+
+        final Property passwordValue = model.getProperties().get("password");
+        assertNull(passwordValue);
+
+        final Property emailValue = model.getProperties().get("email");
+        assertNull(emailValue);
+    }
+
+    @Getter
```
pjfanning: I like lombok but it requires IDE setup and would make life a bit more difficult for 1st time contributors - so for this small usage, I think it would be better to avoid lombok and add  the getters and setters explicitly - your IDE will probably generate them for you

fehguy:
@pjfanning that's just for a specific flavor of tests so I think it's OK?

pjfanning:
@fehguy https://projectlombok.org/download.html describes what you need to install on various IDE environments. Potential contributors would need to know about lombok and how to use it in their IDE. Unless the test is specifically about supporting lombok, I would suggest it is better not to introduce the dependency.

fehguy:
@pjfanning yes that is ugly.  So if you do not do this, what happens in your IDE?

pjfanning:
@fehguy the IDE won't compile the particular class with the lombok annotations with the IDE modifications

fehguy:
@ralphdoe, per comments from @pjfanning could you please keep your _fix_ but remove the dependency on Lombok and the test that depends on it?

modules/swagger-jersey2-jaxrs/src/test/java/io/swagger/models/HiddenFieldTest.java
ralphdoe: I think you're right, it's better to use getters and setters considering is so easy to generate them, but we have to consider that someone may want to use Lombok, so it's necessary to give solution for this problem.

fehguy:
Replaced by #2251

Issue #2178, opened 2017-04-24
Properties of type "number" of "integer" without format lose their "enum" and "defaultValue" at import


boillodmanuel:
**Affect** 
swagger-core 1.5.13

**How to reproduce**
Example of swagger definition that loses `enum` and `defaultValue` at import:
```yaml
---
swagger: "2.0"
info:
  version: 1.0.0
  title: Examples
paths: {}
definitions:
  AnObject:
    type: object
    properties:
      numberProperty:
        type: number
        default: 2
        enum: [ 2, 42, 666 ]
        minimum: 2
        maximum: 666
      integerProperty:
        type: integer
        default: 2
        enum: [ 2, 42, 666 ]
        minimum: 2
        maximum: 666
        example: 2
        multipleOf: 2 # Not supported by RWADef 3.0
```

**Cause:** 
Swagger parser (in`PropertyBuilder` class):
-  parses `numberProperty` into `DecimalProperty` that doesn't have those two fields contrary to `DoubleProperty`.
- parses `integerProperty` into `BaseIntegerProperty` that doesn't have those two fields contrary to `LongProperty`.

**Resolution proposition**
Either add the missing attributes in `DecimalProperty` and `BaseIntegerProperty`

or deserialize in the largest implementation:`DoubleProperty` and `LongProperty`

fehguy:
Please supply a test case showing the issue

boillodmanuel:
I added this PR with a failing unit test: #2180

boillodmanuel:
Hello, any news about this issue?

saguinav:
Same issue here. I agree with @boillodmanuel 's resolution proposal: `add the missing attributes in DecimalProperty and BaseIntegerProperty`.

MichakrawSB:
Closing this issue as part of our [backlog cleanup effort](https://github.com/swagger-api/swagger-core/issues/5074).  
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.  
Thank you for your understanding!

PR #2180, opened 2017-04-25
add unit test showing deserialization issue of number type without fo…

boillodmanuel:
add unit test showing deserialization issue of number type without format (#2178)

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/ObjectPropertyTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/ObjectPropertyTest.java
@@ -6,6 +6,7 @@ import static org.testng.Assert.assertTrue;
 
 import java.io.IOException;
 
+import io.swagger.models.properties.DoubleProperty;
 import org.testng.annotations.Test;
 import io.swagger.models.ModelImpl;
 import io.swagger.models.properties.ObjectProperty;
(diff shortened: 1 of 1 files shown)


modules/swagger-core/src/test/java/io/swagger/ObjectPropertyTest.java
```diff
@@ -47,4 +48,46 @@ public void readModelWithObjectProperty() throws IOException {
...
+                "properties:\n" +
+                "  id:\n" +
+                "    type: \"number\"\n" +
+                "    enum: [1.0, 2.0]\n" +
+                "    default: 1.0";
+
+        final ModelImpl modelWithoutFormat = io.swagger.util.Yaml.mapper().readValue(yamlNumberWithoutFormat, ModelImpl.class);
+
+        // NOTE: Expect to be a DoubleProperty because DecimalProperty didn't have enum or default attributes
+        final DoubleProperty propertyWithoutFormat = (DoubleProperty)modelWithoutFormat.getProperties().get("id");
```
boillodmanuel: The test is currently failing here (see NOTE above)

fehguy:
Since we're focusing on the 3.0 version, I prefer to leave the current behavior as-is, unless there's a strong need to change.

PR #2187, opened 2017-04-27
Fix Empty model output

smrutimo:
Currently while creating models, if the model object has no properties and is just a dummy response, the generated swagger json does not include the empty object model. 

This would fix such scenarios and always generate the model even if its an empty object


fehguy:
Hi, we're not actively updating this (very old) version.  Can you consider updating to the latest?

smrutimo:
@fehguy Sure. So which branch should I be fixing ?

fehguy:
Well I would first look to see if it's even an issue in master (currently 1.5.13).  If so, then that would be the right place.

PR #2196, opened 2017-05-08
renamed packages

fehguy:


modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
```diff
@@ -70,7 +70,7 @@ public ObjectMapper objectMapper() {
     protected boolean shouldIgnoreClass(Type type) {
         if (type instanceof Class) {
             Class<?> cls = (Class<?>) type;
-            if (cls.getName().equals("javax.ws.rs.Response")) {
+            if (cls.getName().equals("javax.ws.rs.ApiResponse")) {
```
mrglavas: Guessing this got changed in the package renaming. Shouldn't this be "javax.ws.rs.core.Response"?

fehguy: thanks, fixed that.

PR #2204, opened 2017-05-12
Add Property.isPrimitiveType()

garydgregory:
For advanced Swagger model processing, I need a clean way to distinguish primitive types from others. This patch add the method Property.isPrimitiveType() and implements it in primitive property classes.

### DIFF
--- a/modules/swagger-core/.gitignore
+++ b/modules/swagger-core/.gitignore
@@ -1,0 +1,1 @@
+/test-output/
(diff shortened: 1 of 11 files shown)


garydgregory:
I see that Travis failed the build: https://travis-ci.org/swagger-api/swagger-core/builds/231706519?utm_source=github_status&utm_medium=notification

I am not sure what this failure means or what the Swagger Model way is here. Suggestions?

fehguy:
I'm guessing that the build won't work on your local machine either--there are numerous test failures due to changes in signatures, probably making the JSON schema invalid.  I'd suggest looking at them locally to see what the cause is.

garydgregory:
I am going to revisit this convenience method later. I will create a new PR with a narrower footprint to achieve the same goal with a new method PrimitiveType.isSwaggerPrimitive(Property).

fehguy:
Closing per #2206

PR #2206, opened 2017-05-15
A method PrimitiveType.isSwaggerPrimitive()

garydgregory:
A method PrimitiveType.isSwaggerPrimitive().

I am calling the new method isSwaggerPrimitive() to distinguish the fact that PrimitiveType includes OBJECT as an enum which is NOT a Swagger Primitive Data Type per the 2.0 spec.

This is a smaller footprint than https://github.com/swagger-api/swagger-core/pull/2204, which I might revisit.

### DIFF
--- a/modules/swagger-core/.gitignore
+++ b/modules/swagger-core/.gitignore
@@ -1,0 +1,1 @@
+/test-output/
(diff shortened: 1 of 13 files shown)


fehguy:
I understand the change, but what is the motivation for doing so?

fehguy:
Similar logic is in the 3.0 support branch here:

https://github.com/swagger-api/swagger-core/tree/feature/3.0.0-rc1

PR #2214, opened 2017-05-22
Add not null check for innerModel properties

### LINKED ISSUE #2213: NPE in ModelResolver
I have a very specific case that causes Swagger to throw an NPE.  I have an example project which recreates the problem [here](https://github.com/pofallon/swagger-issue-example).  

In the example project I have a Doodad object, which implements an Identifiable<T> interface.  In this interface, the getId() method has the Jackson `@JsonIgnore` annotation.  In my resource, I return a Wrapper<Doodad> where Wrapper uses Jackson's `@JsonUnwrapped` to add an extra field to the Doodad JSON.  The combination of `@JsonIgnore` in the interface and `@JsonUnwrapped` in the wrapper (plus Swagger's `@Api` in the resource) causes Swagger to throw an NPE at `io.swagger.jackson.ModelResolver.handleUnwrapped(ModelResolver.java:608)` (see the project README above for the full stack trace).  Remove any of these three annotations and the NPE goes away (along with the accompanying functionality).

Any insight is greatly appreciated!

pofallon:
Fixes #2213

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
@@ -615,7 +615,10 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
 
     private void handleUnwrapped(List<Property> props, Model innerModel, String prefix, String suffix) {
         if (StringUtils.isBlank(suffix) && StringUtils.isBlank(prefix)) {
-            props.addAll(innerModel.getProperties().values());
+            Map<String, Property> innerProps = innerModel.getProperties();
+            if (innerProps != null) {
+                props.addAll(innerProps.values());
+            }
         } else {
             if (prefix == null) {
                 prefix = "";


dantran:
I would like to expand few more NPE issues at the same line

1. When the model has no property ( ie empty class) which  should ever happen
    innerModel.getProperties() is empty, but InnerModel.getProperties().values() is null

2. When @ApiOperation/ApiResponse has response = generic class ( ie Spring Hateoas  Resource.class which as an #JsonUnwrapped field)
    @Path("/{id}")
    @ApiOperation(value = "Get ClusterStatus by id", httpMethod = "POST", produces = "application/json", response = Resource.class)
    @ApiResponses(value = { @ApiResponse(code = 201, message = "Successfully created an AssetRuleSet task.") })
    public void create(@PathVariable String id) {
        //return null;
    }
3. When actual API return Spring Resource<Pojo> and @ApiOperation/ApiResponse has no response field ( must set to 'Pojo.class' to get it working)

    @Path("/{id}")
    @ApiOperation(value = "Get ClusterStatus by id", httpMethod = "POST", produces = "application/json")
    @ApiResponses(value = { @ApiResponse(code = 201, message = "Successfully created an AssetRuleSet task.") })
    public Resource<Pojo>create(@PathVariable String id) {
        return null;
    }

PR #2242, opened 2017-06-08
Allow @Api annotations be part of meta-annotations

kubahorak:
I'd like to use `@ApiModelProperty` and other `@Api*` annotations on Spring [meta-annotations](
https://docs.spring.io/spring/docs/current/spring-framework-reference/html/beans.html#beans-meta-annotations).

It's a simple code change, that improves Spring integration.


fehguy:
this will be part of the 3.0 support, and is already added here:

https://github.com/swagger-api/swagger-core/tree/feature/3.0.0-rc1/modules/swagger-annotations

Issue #2150, opened 2017-03-27
Support processing package-level annotations, specifically, to get namespaces


rwalkerands:
If you generate Java code from an XML Schema with JAXB's `xjc` tool, you normally get a `package-info.java` file generated that contains a package-level annotation. Here's an example:

```
//
// This file was generated by the JavaTM Architecture for XML Binding(JAXB) Reference Implementation, v2.2.11
// See <a href="http://java.sun.com/xml/jaxb">http://java.sun.com/xml/jaxb</a>
// Any modifications to this file will be lost upon recompilation of the source schema.
// Generated on: 2017.03.27 at 11:06:28 AM AEDT
//

@javax.xml.bind.annotation.XmlSchema(namespace = "http://vocabs.ands.org.au/registry/schema/2017/01/vocabulary", elementFormDefault = javax.xml.bind.annotation.XmlNsForm.QUALIFIED)
package au.org.ands.vocabs.registry.schema.vocabulary201701;
```

Note the `namespace` property, with the value as defined in my XML Schema. And in this case, the `namespace` property is _not_ then also included in the `@XmlType` annotations at the _class_ level.

If you use the generated classes in your API, so that swagger-core includes their definitions in the generated `swagger.json`, swagger-core currently doesn't process these package-level annotations, so the `namespace` properties get lost when generating the model which is then used to generate `swagger.json`.

There is a workaround: add the `-npa` option to the `xjc` command line. Then, `package-info.java` is not generated, and the `namespace` property is included at the class level.

But, it would be nice to have support for the package-level annotation `@XmlSchema` ...

packleader:
I have added a fix for this issue and created a pull request - https://github.com/swagger-api/swagger-core/pull/2244

packleader:
I have added a fix for this issue and created a pull request - https://github.com/swagger-api/swagger-core/pull/2246

rwalkerands:
Thank you for this feature.

PR #2244, opened 2017-06-09
Get namespace from package-info #2150

packleader:
This PR addresses [Issue 2150](https://github.com/swagger-api/swagger-core/issues/2150).  When resolving a `JavaType` to a Swagger `Model`, if the type is annotated with `XmlRootElement` and no `namespace` is specified, the resolver will look in `package-info` for an package-level annotation.


packleader:
I just identified some issues with this PR.  Will resolve and re-submit.

PR #2246, opened 2017-06-09
Get namespace from package-info #2150

packleader:
This PR addresses Issue [2150](https://github.com/swagger-api/swagger-core/issues/2150). When resolving a `JavaType` to a Swagger `Model`, if the type is annotated with `XmlRootElement` and no namespace is specified, the resolver will look in `package-info` for an package-level annotation.


packleader:
@fehguy Thanks for accepting this PR.  My team has a defect in our product, and the resolution depends on these changes.  Do you have a timeline for when this PR will be released?  That would be very helpful to us as we try to plan our work.

fehguy:
Today!

packleader:
Thanks! We've adopted the new version, and it enabled us to resolve our bug.  Should we close https://github.com/swagger-api/swagger-core/issues/2150 ?

fehguy:
Thanks--it's closed now.

Issue #2207, opened 2017-05-15
Equals and hashCode methods missing extensions field in OAS models


arturdzm:
OAS model classes are missing vendor extension field from equals and hashcode methods

arturdzm:
Also some other fields that are recently added also are missing as well in some cases

arthurdm:
Fixed

PR #2250, opened 2017-06-13
Fix for issue #2207

a-saf:
Updates for equals and hashCode methods for missing fields in OAS models.


a-saf:
Pull request for issue #2207

PR #2282, opened 2017-06-27
Feature/jaxrs test migration

ralphdoe:
Annotations Test Migration. From annotations to jaxrs2 module. (Pet Resource Sample added)
 - Operations
 - Links
 - Callbacks
 - Info

Reader Improvements
 - Empty and default values deleted.
 - Merging operations with the same "path" but different HTTPMethod.
 - Schema improvements
 - Parameters improvements
 - Links Improvements.
 - SwaggerContextService added.

TODO Work
 - Explode and style management.
 - Subresources and subclasses to the Reader.
 - Reuse the models and adding them to the components object.


modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/Reader.java
```diff
@@ -46,389 +46,401 @@
...
+            operation.setOperationId(getOperationId(apiOperation.operationId()));
+        }
+        if (apiOperation.deprecated()) {
+            operation.setDeprecated(apiOperation.deprecated());
+        }
+        ReaderUtils.getStringListFromStringArray(apiOperation.tags()).ifPresent(operation::setTags);
+        OperationParser.getTags(apiOperation.tags()).ifPresent(tag -> openApiTags.addAll(tag));
+        OperationParser.getExternalDocumentation(apiOperation.externalDocs()).ifPresent(operation::setExternalDocs);
+        OperationParser.getRequestBody(apiOperation.requestBody()).ifPresent(operation::setRequestBody);
+        OperationParser.getApiResponses(apiOperation.responses()).ifPresent(operation::setResponses);
```
fehguy: Do a check to see if `responses.size()` is > 0 or not.  The `apiOperation.responses()` is never `Empty` because it's an empty map.  That will avoid this:

```
        "responses": {
          "default": {}
        }
```
Which we don't want

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
```diff
@@ -32,256 +35,497 @@
...
+        if (isEmpty) {
+            return Optional.empty();
+        }
+
+        getContents(parameter.content()).ifPresent(parameterObject::setContent);
+
+        return Optional.of(parameterObject);
+    }
+
+    public static Optional<Map<String, Schema>> getSchema(io.swagger.oas.annotations.media.Schema schema) {
```
fehguy: Suggest you don't return `Optional` on maps.  Return an empty map instead

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
```diff
@@ -32,256 +35,497 @@
...
+        Class<?> schemaImplementation = annotationContent.schema().implementation();
+        Map<String, Schema> schemaMap;
+        if (schemaImplementation != Void.class) {
+            schemaMap = ModelConverters.getInstance().readAll(schemaImplementation);
+            schemaMap.forEach((k, v) -> mediaType.setSchema(v));
+        } else {
+            getSchemaFromAnnotation(annotationContent.schema()).ifPresent(mediaType::setSchema);
+        }
+        if (StringUtils.isNotBlank(annotationContent.mediaType())) {
+            content.addMediaType(annotationContent.mediaType(), mediaType);
```
fehguy: For some reason, the schemas are being added in-line.  They need to be added to the `components` object and set as a reference.  This:

```
"content": {
  "application/json": {
    "schema": {
      "type": "object",
      "properties": {
        "id": {
          "type": "integer",
          "format": "int64"
        },
        "username": {
          "type": "string"
        },
        "firstName": {
          "type": "string"
        },
        "lastName": {
          "type": "string"
        },
        "email": {
          "type": "string"
        },
        "password": {
          "type": "string"
        },
        "phone": {
          "type": "string"
        },
        "userStatus": {
          "type": "integer",
          "description": "User Status",
          "format": "int32",
          "enum": [
            null
          ]
        }
      }
    }
  }
}
```

Should be this:

```
"content": {
  "application/json": {
    "schema": {
      "$ref": "#/components/schemas/User"
    }
  }
}
```

With the corresponding components:

```
"components": {
  "schemas": {
    "User": {
      "type": "object",
      "properties": {
        "id": {
          "type": "integer",
          "format": "int64"
        },
        "username": {
          "type": "string"
        },
        "firstName": {
          "type": "string"
        },
        "lastName": {
          "type": "string"
        },
        "email": {
          "type": "string"
        },
        "password": {
          "type": "string"
        },
        "phone": {
          "type": "string"
        },
        "userStatus": {
          "type": "integer",
          "description": "User Status",
          "format": "int32",
          "enum": [
            null
          ]
        }
      }
    }
  }
}
```

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
```diff
@@ -32,256 +35,497 @@
...
+        parameterObject.setStyle(StringUtils.isNoneBlank(parameter.style()) ? Parameter.StyleEnum.valueOf(parameter.style()) : null);
+        if (parameter.allowEmptyValue()) {
+            parameterObject.setAllowEmptyValue(parameter.allowEmptyValue());
+            isEmpty = false;
+        }
+        if (parameter.allowReserved()) {
+            parameterObject.setAllowReserved(parameter.allowReserved());
+            isEmpty = false;
+        }
+        if (parameter.explode()) {
```
fehguy: Do an empty string check to avoid this:

```
"explode": false
```

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/annotations/operations/AnnotatedOperationMethodTests.java
```diff
@@ -0,0 +1,321 @@
...
+                "          description: \"Pet not found\"\n" +
+                "        405:\n" +
+                "          description: \"Validation exception\"\n" +
+                "    post:\n" +
+                "      summary: \"Add a new pet to the store\"\n" +
+                "      parameters:\n" +
+                "      - description: \"Pet object that needs to be added to the store\"\n" +
+                "        required: true\n" +
+                "        explode: false\n" +
+                "        schema:\n" +
```
fehguy: `parameter` object should _never_ be missing an `in` attribute.  In this case, the parameter body must be a `requestBody` rather than a `parameter`.

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/annotations/operations/AnnotatedOperationMethodTests.java
```diff
@@ -0,0 +1,321 @@
...
+                "                type: \"object\"\n" +
+                "                properties:\n" +
+                "                  id:\n" +
+                "                    type: \"string\"\n" +
+                "                    description: \"the user id\"\n" +
+                "              examples:\n" +
+                "                basic:\n" +
+                "                  summary: \"shows a basic example\"\n" +
+                "                  description: \"basic\"\n" +
+                "                  value: \"{\\\"id\\\": 19877734}\"\n" +
```
fehguy: This is not right, the value should be parsed into an object not shown as an escaped string

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/Reader.java
ralphdoe: Done! I have added new validations.

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
ralphdoe: Done. I made a refactor of that methods.

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
ralphdoe: Validation added in the ParameterProcessor, there was functioning correctly here.

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/annotations/operations/AnnotatedOperationMethodTests.java
ralphdoe: The type of value property was changed, now is an object, and JSonNode parsing added.

fehguy:
@ralphdoe almost there!  I can't locate the spot in the code right now but I'll point out some issues I'm seeing.

* The `requestBody` is still missing conversion.  They're showing up as parameters without an `in` attribute
* `operationId` values should be used when explicitly provided, or extracted from method names.  I don't see them working in either case
* All operations _must_ have a `responses` object with at least _one_ response.  If there is literally nothing to extract, I would use `default` as the response code, and a description of "no description provided" when extracting

fehguy:
Great!  I see some issues still but let's get this merged.  Thank you @ralphdoe !

Issue #2291, opened 2017-07-03
[3.0] Reader - Explode and style improvement


ralphdoe:
1. There's some validations to add in the explode an style properties of the parameter object.
- [x] The style parameter depends on the "in" attribute
- [x] The explode depends on the parameter type ("in") and the schema type.

@fehguy @webron

frantuma:
fixed in https://github.com/swagger-api/swagger-core/pull/2294

PR #2294, opened 2017-07-04
Bugfix/issue 2291 explode style refactor

ralphdoe:
The explode and style now have default values depending on the parameter type
Issue: #2291


modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
```diff
@@ -144,6 +149,44 @@ public static Parameter applyAnnotations(OpenAPI openAPI, Parameter parameter, T
...
+                if (!schema.type().equals("object") && !schema.type().equals("array")) {
+                    explode = false;
+                }
+            }
+        }
+        return explode;
+    }
+
+    public static void setParameterStyle(Parameter parameter, io.swagger.oas.annotations.Parameter p) {
+        if (isExplodable(p)) {
```
webron: Parameter style doesn't depend on it being explodable or not.

modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
```diff
@@ -144,6 +149,44 @@ public static Parameter applyAnnotations(OpenAPI openAPI, Parameter parameter, T
...
+            if (StringUtils.isNotBlank(p.style())) {
+                parameter.setStyle(Parameter.StyleEnum.valueOf(p.style()));
+            } else {
+                if (HEADER.equals(p.in()) || PATH.equals(p.in())) {
```
webron: Just like with `explode`, if no value is provided by the user, we can assume the default value, and in that case it doesn't need to be serialized.

modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
```diff
@@ -144,6 +149,44 @@ public static Parameter applyAnnotations(OpenAPI openAPI, Parameter parameter, T
...
+                }
+            }
+        }
+        return explode;
+    }
+
+    public static void setParameterStyle(Parameter parameter, io.swagger.oas.annotations.Parameter p) {
+        if (isExplodable(p)) {
+            if (StringUtils.isNotBlank(p.style())) {
+                parameter.setStyle(Parameter.StyleEnum.valueOf(p.style()));
```
webron: The `style` value needs to be in lowercase - based on the tests, it is not.

modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
ralphdoe: Done!

modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
ralphdoe: Done!

modules/swagger-core/src/main/java/io/swagger/util/ParameterProcessor.java
ralphdoe: Done! The Enums are now in lower case: mapper.configure(SerializationFeature.WRITE_ENUMS_USING_TO_STRING, true);

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
```diff
@@ -83,7 +85,7 @@
             parameterObject.setRequired(parameter.required());
             isEmpty = false;
         }
-        parameterObject.setStyle(StringUtils.isNoneBlank(parameter.style()) ? Parameter.StyleEnum.valueOf(parameter.style()) : null);
+        parameterObject.setStyle(StringUtils.isNoneBlank(parameter.style()) ? Parameter.StyleEnum.valueOf(parameter.style().toUpperCase()) : null);
```
ralphdoe: Sure, there's a method that handles the style values. This line was deleted.

Issue #2191, opened 2017-05-02
Parameterized type as body parameter results in incorrect generated JSON 


sebeard:
Hi,

I'm using swagger annotations with JAX-RS and have an 'Pet' resource. The resource has the usual;

- Get Pet Collection 
- Get Pet
- Create Pet
- Update Pet
- Remove Pet

These resource endpoints revolve around 3 core classes and their children classes; GetPet<T extends Pet>, UpsertPet<T extends Pet>, and Pet. This allows me to use the same resource for Cats, Dogs, Platypuses, etc. while segregating out creation/modification from retrieval.

A typical endpoint looks like this (with Swagger annotations);
```@PUT
    @Path("/{id}")
    @Produces(MediaType.APPLICATION_JSON)
    @ApiOperation(value = "update a pet", response = GetPet.class)
    public GetPet update(
            @ApiParam(value = "pet ID", required = true)  @PathParam("id") final String petId,
            @ApiParam(type = "com.example.api.model.UpsertPet", value = "Pet to update", required = true) UpsertPet<Pet> upsertPet
    ) {
        return petService.updatePet(id, upsertPet);
    }
```

In the Swagger class ParameterProcessor on method applyAnnotations, I can see that BodyParameters are handled differently to say QueryParams or PathParams. The type, although defined in the annotation for the body param (see example above), is ignored by this method and not set at all (lines 219-275).

This results in swagger parsing the `UpsertPet<Pet>` parameter as UpsertPetPet. Using `<? extends Pet>` results in the same behavior and there doesn't seem to be a workaround other than manually editing the resulting JSON. As you will see in the JSON below, both UpsertPet and UpsertPetPet are produced. UpsertPet by looking at the heirarchy (UpsertCat, UpsertDog) and UpsertPetPet by looking at the resource.

I think this is missing from Swagger core and needs to be added, unless anyone can suggest how to get this to work while respecting the annotation. Removing the parameterized type `<Pet>` or `<? extends Pet>` is not really an option since it introduces unchecked casting which I'm looking to avoid.

Below is the JSON produced.
```
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.2"
  },
  "host": "localhost:8180",
  "basePath": "/api/v1",
  "tags": [
    {
      "name": "hello"
    }
  ],
  "schemes": [
    "http"
  ],
  "paths": {
    "/api/v1/{id}": {
      "get": {
        "tags": [
          "hello"
        ],
[...]

PR #2295, opened 2017-07-04
#2191 - Added typing to BodyParameter 

sebeard:
To support specifying the correct class name when class in Parameterized. Swagger Spec should now use just the class name (e.g. A class UpsertPet<T extends Pet> will now appear as UpsertPet when specified using the full package reference where previously this was UpsertPetPet which did not match any other generated classes in the Swagger Spec.


kopernic-pl:
1.5 tree is not maintained anymore. Closing.

Issue #2292, opened 2017-07-03
YAML Serialization Improvements


webron:
Not sure what the YAML serializer supports, but the following should be handled:
- [x] String values should not be quoted unless required. For example:

```yaml
      summary: "Create user"
```
should be
```yaml
      summary: Create user
```

- [x] Values that must be quotes, should use single quotes. For example:

```yaml
              $ref: "#/components/schemas/User"
```
should be
```yaml
              $ref: '#/components/schemas/User'
```

- [ ] Numeric keys must be quoted. For example:
```yaml
      400:
          description: "User not found"
```
should be
```yaml
      '400':
          description: "User not found"
```

ralphdoe:
It seems the Response Code isn't an integer, is a natural String.

ralphdoe:
Closed because it was treated in other PR. Yaml options #2303

fehguy:
Hey @ralphdoe that PR is only partially fixing this issue.  Namely the unchecked item above is not done.

@webron I've dug into this more and I don't know how we can support quoting numeric keys without hacking Jackson.  It looks like the logic to determine if its "a number" or not is a regex and when we say "minimize quotes", it will automatically remove quotes in keys, if it passes [the regex](https://github.com/FasterXML/jackson-dataformat-yaml/blob/master/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java#L157).

Here's the code that determines this:

https://github.com/FasterXML/jackson-dataformat-yaml/blob/master/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java#L504-L526

MichakrawSB:
Closing this issue as part of our https://github.com/swagger-api/swagger-core/issues/5074.
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.
Thank you for your understanding!

PR #2303, opened 2017-07-07
Yaml options

fehguy:
Partial fix for #2292


fehguy:
@webron this doesn't cover the quoted numeric keys but should get the others.  If good w/you, let's merge and keep #2292 open to track the numeric key issue (that existed in the previous major version too)

modules/swagger-core/src/main/java/io/swagger/util/ObjectMapperFactory.java
```diff
@@ -24,7 +25,13 @@ protected static ObjectMapper createYaml() {
     }
 
     protected static ObjectMapper createYaml(boolean includePathDeserializer, boolean includeResponseDeserializer) {
-        return create(new YAMLFactory(), includePathDeserializer, includeResponseDeserializer);
+        YAMLFactory factory = new YAMLFactory();
+        factory.disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER);
+        factory.enable(YAMLGenerator.Feature.MINIMIZE_QUOTES);
+        factory.enable(YAMLGenerator.Feature.SPLIT_LINES);
+        factory.enable(YAMLGenerator.Feature.ALWAYS_QUOTE_NUMBERS_AS_STRINGS);
```
webron: Will this quote actual numeric values? Like `maxLength`?

fehguy: see the tests, only if they start that way

fehguy:
Looks like the test failures are because we're expecting quoted strings in places where we're now getting unquoted ones.

Issue #2193, opened 2017-05-04
v3 programmatic interface


arthurdm:
This issue is to propose a programmatic interface that enables complementary services to the use of annotations (JAX-RS 2.0, OpenAPI v3, or both).

We'll have a single interface (for simplicity) that encapsulates a few different scenarios.  Here's a mock of what we're thinking about:

interface OpenAPIController
- bootstrap(OpenAPI) 
- beforeScan(OpenAPI, Config)
- afterScan (OpenAPI)

Where:
- bootstrap would be used for populating the entire OpenAPI model, and avoiding annotating scanning
- beforeScan would be used for partially populating a OpenAPI model before an annotation scan starts.  The optional Config obj would encapsulate things like a filter, and scanning options such as which packages / classes to scan, etc.
- afterScan would be used for manipulating the OpenAPI model after it was populated via annotations.

leochr:
After more discussion, we think it's better to further simplify the interface.

interface OpenAPIController

- OpenAPIConfig bootstrap();

where bootstrap() can be used to return an OpenAPIConfig object which encapsulates all configurations (for Reader, Scanner, OpenAPI, etc) along with a predefined OpenAPI definition (optional) to start-off (similar to beforeScan in v2). 

OpenAPIConfig class will have getters and setters for configurable items:

getOpenAPI()
setOpenAPI(OpenAPI)
getClasses()
setClasses(Set<Class<?>>)
isScan()
setScan(boolean)
getFilterClass()
setFilterClass(String)
isPrettyPrint()
setPrettyPrint(boolean)
getIgnoredRoutes()
setIgnoredRoutes(Collection<String>)
.....
.....

All setters will return the OpenAPIConfig back to allow users to set more configurations easily.

OpenAPIConfig should ideally only use exposed classes / interfaces and avoid using internal / heavy-lift classes (more dependency),  so it can be exposed to users easily.

Any other methods that needs to be invoked by Reader/Scanner can be added to OpenAPIController. Dropped 'afterScan' method, since it seems users can instead use the OpenAPIConfig to control the generated OpenAPI definition (by setting which packages / classes to scan / ignore) instead of overwriting the generated definition. @fehguy your thoughts?

leochr:
A sample implementation of OpenAPIController : 

```
public class OpenAPIControllerImpl implements OpenAPIController {

	@Override
	public OpenAPIConfig bootstrap() {
		return new OpenAPIConfig()
[...]

PR #2307, opened 2017-07-12
Initial commits for the v3 programmatic interface

navidsh:
PR for #2193: Interfaces and a simple implementation for the OpenAPIConfig interfaces a new module called swagger-web.


arthurdm:
Thanks a lot for the feedback @frantuma 

Here are my thoughts:

1.  That's an interesting point - can you provide an example scenario where we would use the OpenAPIController and OpenAPIConfig in a non-web environment?   Given that these interfaces are controlling the building of documentation for REST APIs, I am having a hard time imagining a non-web environment.  By the way, the choice of "web" was on purpose to keep it general from jaxrs or servlet, so to mean any web module that is exposing REST APIs.

2.  Good point.  I like OpenApiConfigBuilder + build method.  

3.  We definitely don't want more than 1 Filter version in the repo, so if you could please share the current work that would be great.  We could always pass that in through the general user-defined map too.  

4.  That's true.  We'll revisit that.  Perhaps we could have both the classes and the package variations, so the user can have choices (they are optional)

5.  That's a tricky area.  I can see a point in providing access to the Reader to control the reading behaviour - but then again, if users have a nice Filter that controls what goes in, would someone really want to implement a full Reader, since transforming a set of Classes into OpenAPI models is what most people would use the library for.     For the Scanner, I really think that taken care of from OpenAPIConfig (ie: using the getClasses method, and user defined Map<String,Object>).  

6.  We were hoping to push these in as the "spec", so we can work in parallel in the plumbing from different implementations (ie: OSS, WebSphere, or vendors, etc).  We will definitely contribute a JAXRS sample for OpenAPI v3, but using the OpenAPIConfig to configure the Reader (beyond its current default) is something that could be done in a separate issue/PR.

navidsh:
Committed changes to address comments 2 and 4 above.

frantuma:
@arthurdm some comments about points 1 and 5 above.

1. you're probably right, I was thinking (not much..) about e.g. some scenario where user would want to expose an existing app service layer, with the spec used to build a dynamic javafx or swing UI based on the schema or the like; Don't have a strong opinion here though.

5. here I still think it would be valuable to keep the scanning logic separate from config building, as it could e.g. potentially be time consuming, and also be "dynamic" and so on. I understand we can pass whatever in a `Map<String, Object>`, but this sounds too generic imo. Also for the reader I would think that adding it to the config would add value, e.g. to allow a user to use a custom (or library) jaxrs-free reader, or a scala reader, keeping an easy entry point.
I understand we have diverging opinions here, not sure if @webron or @fehguy have any comment to add, we can give it maybe a little more thought and then we decide.

arthurdm:
Thanks @frantuma 

I can see your points about Reader and Scanner.  How about this revised set of interfaces:

-------------------------------------------------------
public interface OASConfigBuilder {
     OASConfig build(Map<String, Object> environment);
} 
-------------------------------------------------------
public interface OASConfig {
     OpenAPI getOpenAPI();
   
     boolean isScanDisabled();

     Class<?> getFilterClass();

     Set<String> getResourcePackages();
      
     Set<String> getIgnoredClasses();

     Map<String, Object> getOptions();
 }
-------------------------------------------------------
 public interface OASScanner {
    Set<Class<?>> classes();
    Map<String, Object> resources()
 }
-------------------------------------------------------
 public interface OASReader {
     OpenAPI read(OASConfig configuration, OASScanner scanner)
 }
-------------------------------------------------------

So basically, the application developer would choose which services they want to provide to the framework (OSS Swagger, WebSphere, etc):

custom config:   META-INF/services/io.swagger.oas.web.OASConfigBuilder 
custom scanner:  META-INF/services/io.swagger.oas.web.OASScanner
custom reader:  META-INF/services/io.swagger.oas.web.OASReader

For each of those, the framework would have default services impl if they aren't specified by the app dev.

Thoughts?

frantuma:
@arthurdm Thanks for comments and revised PR! getting there I'd say.. I add here a couple of  comments.

We have a working "integration implementation" draft, which makes use of `swagger-web` module for a part of the functionality. Possibly it can be further refactored to "integrate more" with `swagger-web` design concepts, however at the moment it references and implements scanner and reader, and uses config and builder classes which implement `swagger-web` related interfaces.

This effort is visible in branches for [core](https://github.com/frantuma/swagger-core/tree/bootstrapping) and for [samples](https://github.com/frantuma/swagger-samples/tree/bootstrapping) (jersey2 and java-dropwizard modules).

At the moment it uses a slightly modified [version](https://github.com/frantuma/swagger-core/blob/bootstrapping/modules/swagger-web/src/main/java/io/swagger/oas/web/OpenAPIConfig.java) of `OpenApiConfig` interface, which basically includes also reader and scanner and adds couple of props/methods which were previously scattered around 2.0 config classes. 

As you know, my idea is that scanners and readers are part of the game, and their implementation classes would be part of the config; if you have a strong opinion against it however, we can keep them out from the interface (we'll keep them in the specific oss "integration implementation"), in this case [this](https://github.com/frantuma/swagger-core/blob/bootstrapping/modules/swagger-web/src/main/java/io/swagger/oas/web/MinimalOpenAPIConfig.java) would be the interface more similar to what you had in mind (with additional props added).

Another possible point is about returning `Class` from e.g. `getFilterClass`, in my idea I'd rather have a simple and serializeable  bean so returning strings, this would facilitate loading of config from files, or any declarative config, with no custom deserialization mechanism, and would also delegate class loading to the caller. This alternative is implemented [here](https://github.com/frantuma/swagger-core/blob/bootstrapping/modules/swagger-web/src/main/java/io/swagger/oas/web/AnotherOpenAPIConfig.java) (for "full config), or [here](https://github.com/frantuma/swagger-core/blob/bootstrapping/modules/swagger-web/src/main/java/io/swagger/oas/web/AnotherMinimalOpenAPIConfig.java) (for "minimal config").

Basically my preferences would be in this order:
1. [Full / Strings](https://github.com/frantuma/swagger-core/blob/bootstrapping/modules/swagger-web/src/main/java/io/swagger/oas/web/AnotherOpenAPIConfig.java)
2. [Full / Classes](https://github.com/frantuma/swagger-core/blob/bootstrapping/modules/swagger-web/src/main/java/io/swagger/oas/web/OpenAPIConfig.java)
3. [Basic / Strings](https://github.com/frantuma/swagger-core/blob/bootstrapping/modules/swagger-web/src/main/java/io/swagger/oas/web/AnotherMinimalOpenAPIConfig.java)
4. [Basic / Classes](https://github.com/frantuma/swagger-core/blob/bootstrapping/modules/swagger-web/src/main/java/io/swagger/oas/web/MinimalOpenAPIConfig.java)

Please let me know what you think, and what's your preference, and we can use that (merging with your latest effort) in the delivered version.

An additional point is about naming: I would feel that sticking to `OpenApi` instead of `oas` would be possibly be more consistent with the rest of the code and the spec, let me know what you think.

frantuma:
@arthurdm forgot a further point about builder: in [my version](https://github.com/frantuma/swagger-core/blob/bootstrapping/modules/swagger-web/src/main/java/io/swagger/oas/web/OpenAPIConfigBuilder.java#L4) I removed the parameters (as you coded in the first place..), as that would be passed to the builder via its implementation

arthurdm:
I had a chat  with @frantuma today and the latest proposal that we're discussing is:

Introduce these 4 interfaces:

`public interface OpenAPIConfigBuilder {
OpenAPIConfig build(Map<String, Object> environment);
}

public interface OpenAPIConfig {
  OpenAPI getOpenAPI();

    String getFilterClass();

    String getReaderClass();

    String getScannerClass();

    Set<String> getIgnoredClasses();

    Map<String, Object> getOptions();

    Set<String> getResourceClasses();

    Set<String> getResourcePackages();

    boolean isScanDisabled();
}

public interface OpenAPIScanner {
Set<Class<?>> classes();
Map<String, Object> resources()
}

public interface OpenAPIReader {
OpenAPI read(OpenAPIConfig configuration, OpenAPIScanner scanner)
}
`
So the flow is that an application developer provides (optionally) an implementation of OpenAPIConfigBuilder, which would then allow then to give the framework a copy of OpenAPIConfig that may contain things like a custom Reader, Scanner, options, etc.  

Not part of this PR, but for completeness:  the idea is that the "specification" part of MicroProfile's OpenAPI Java proposal will say something like this:

The application developer can provide an OpenAPIConfig to implementations of the OpenAPI Java spec in 2 ways:
(1) By specifying an openapi-config.yaml file in the application's classpath, which contains the configuration corresponding to an OpenAPIConfig object (link to schema)

(2) By specifying the qualified name of their OpenAPIConfigBuilder service inside META-INF/services, following JDK practice (https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html)

frantuma:
@arthurdm @webron @fehguy 

looks quite ok to me, just one note:

Would add to config interface `isScanAllResources` and `isPrettyPrint`, and remove (if there is no particular reason to keep it) `isScanDisabled`, so like:
```
public interface OpenAPIConfig {

    Set<String> getResourcePackages();

    Set<String> getResourceClasses();

    String getReaderClass();

    String getScannerClass();

    String getFilterClass();

    Collection<String> getIgnoredRoutes();

    OpenAPI getOpenAPI();

    Map<String, Object> getUserDefinedOptions();

    boolean isScanAllResources();

    boolean isPrettyPrint();
}
```

frantuma:
@arthurdm an additional note about loading via service loader; this doesn't cover the case where uses needs different configs in same JVM (e.g. a config per servlet); a possible solution could be adding a `getName()` to the interface to allow the framework to e.g. match it to servlet name, even if I don't like it so much.

arthurdm:
Hey @frantuma - thanks for the comments.

We chose `isScanDisabled` instead of `isScanAllResources` to be more descriptive of the purpose of this boolean:  is it used to disable scanning.     We felt that `isScanAllResources` was not being descriptive enough (for example: I am turning off the scan for ALL resources, but maybe SOME resources will still be scanned?)

Regarding the multiple configs per JVM, I see this as an implementation add-on, rather than a specification item.  Therefore, this is where the `Map<String,Object>` that is passed into `OpenAPIConfigBuilder#build` can be used, where the implementation (ie: Swagger OSS) could pass in an option such {key = SERVLET_NAME, value = MyServlet} and the proper OpenAPIConfig would be return.

frantuma:
with regard to `isScanDisabled` and `isScanAllResources`, `isScanAllResources` is bad naming actually (better `isReadAllResources`), and is meant (used like that in 2.0) to allow to read/resolve all classes passed to reader independently of the presence of specific annotation.

I am not sure what is the usage/meaning `isScanDisabled`, what does it mean `to disable scanning`?

About the multiple configs per JVM, in this way if I am not mistaken, in order to have different configs e.g. per servlet (a quite common scenario), a final user would need to implement a builder without knowing what this `Map<String,Object>` contains, for example this `SERVLET_NAME`, in the best case it would be specified by the implementation and therefore different depending on implementation.

arthurdm:
hey @frantuma - I see now what you mean with `isReadAllResources`.   The option`isScanDisabled` is for scenarios where the user will provide a full OpenAPI model via the Config, so it turns off annotation scanning. (no need to scan the app).

About the multiple JVMs, I still believe that is implementation specific - for example, in Liberty our JVM (the app server) supports multiple applications (WAR module), and each application will have their own OpenAPIConfig inside their WAR module.  So even without extra Map options we can already achieve multiple apps in a JVM scenario.  

Perhaps you are referring to multiple APIs in a single WAR module (single app), which then that can be handled differently depending on the implementation (some implementations may not support it).

Lastly, I think `isPrettyPrint` is outside of the spec - it deals with the serialization format of the vendor/implementation as it handles the OpenAPI model that it got from the OAS3 Java programming model (ie: from OpenAPIConfig or from OpenAPIReader).  How that model is handled will vary per implementation.

frantuma:
hey @arthurdm `isScanDisabled` is now clear, I feel that this also could be an implementation specific thing, but no strong opinion; the same goes for `prettyPrint` and `isReadAllResources` at this point.

With regard to the multiple config per JVM, it's indeed per classloader (e.g. per web application), a common scenario would be e.g. exposing v1 and v2 of APIs or the like; I still feel this would fit in scope of the "spec", but maybe it can be part of a later version.

arthurdm:
hey @frantuma - I think we can then agree to include `isReadAllResources` and `isScanDisabled` as spec, and other things like `isPrettyPrint` as implementation-dependent.  So we will add `isReadAllResources` into the current PR.

Regarding the multiple API per application, my thought was that the javadoc of the `OpenAPIConfigBuilder#build` method would specify that if implementations/vendors want to support different configurations for different APIs in the same application, then they must define an appropriate `key`  in the `Map` that gets passed into `build` - the reason I am trying to keep this away from spec is because "servlet name" wouldn't be application to all implementations, such as Spring.

arthurdm:
Using https://github.com/swagger-api/swagger-core/pull/2398 instead

PR #2313, opened 2017-07-15
 Inner properties may have XML declared

mobreza:
Swagger specification allows for "xml" element defined on the "items" of an array:

> When defined within the Items Object (items), it will affect the name of the individual XML elements within the list. When defined alongside type being array (outside the items), it will affect the wrapping element and only if wrapped is true. If wrapped is false, it will be ignored.

For `ArrayProperty`, the `detailNode = node.get("items")` is read, but the "xml" declaration on that element is not set. This PR moves to `property.setXml(getXml(node));` to the `propertyFromNode(JsonNode node)` method.


mobreza:
See #2316

zregvart:
There is now a pull request rebased against the master in #2674

frantuma:
included and replaced by #2688

Issue #2319, opened 2017-07-18
Ability to add example response to an API


pnrakesh:
Actual swagger spec has ability to add multiple response examples to an API method. In swagger core there is no annotation to add example responses for an API method. Below is the actual swagger spec

```
paths:
  /api/v1/languages:
    get:
      description: Get list of supported languages.
      responses:
        '200':
          description: OK
          schema:
            $ref: '#/definitions/ResponesBody'
          examples:
            example_1: {
			  "responseBody": {
				"error": null,
				"status": "success",
				"data": [
				  "Chinese (Simplified)",
				  "Chinese (Traditional)",
				  "English",
				  "French",
				  "German",
				  "Japanese",
				  "Korean",
				  "Portuguese",
				  "Russian",
				  "Spanish"
				]
			  }
			}

            example_2: {
			  "responseBody": {
				"error": {"code" : 401, "message" : "Authentication failed. username could not be identified from cookies"},
				"status": "error",
				"data": null
			  }
			}
            example_3: {
			  "responseBody": {
				"error": {"code" : 500, "message" : "There was an internal server error"},
				"status": "error",
				"data": null
			  }
			}
}
```

The swagger-core library implements annotations for the entire swagger spec or just a limited set of features?

frantuma:
#2801 introduces example support to response annotation, limited to String types.
Also related to #1107 and linked tickets, a full solution would be similar to the one applied in 2.0 version, probably in scope of #1107 and related.

PR #2323, opened 2017-07-20
#2319 parse @Example from @APiResponse annotations

pnrakesh:
Parse @Example nested inside @ApiResponse annotation

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -22,21 +22,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.introspect.AnnotatedMethod;
 import com.fasterxml.jackson.databind.introspect.AnnotatedParameter;
 import com.fasterxml.jackson.databind.type.TypeFactory;
-import io.swagger.annotations.Api;
-import io.swagger.annotations.ApiImplicitParam;
-import io.swagger.annotations.ApiImplicitParams;
-import io.swagger.annotations.ApiKeyAuthDefinition;
-import io.swagger.annotations.ApiOperation;
-import io.swagger.annotations.ApiResponse;
-import io.swagger.annotations.ApiResponses;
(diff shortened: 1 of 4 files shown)


modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
```diff
@@ -1001,9 +987,10 @@ private void processOperationDecorator(Operation operation, Method method) {
 
     private void addResponse(Operation operation, ApiResponse apiResponse) {
         Map<String, Property> responseHeaders = parseResponseHeaders(apiResponse.responseHeaders());
+        Map<String, Object> examples = parseExamples(apiResponse.examples());
 
         Response response = new Response()
-        .description(apiResponse.message()).headers(responseHeaders);
+        .description(apiResponse.message()).headers(responseHeaders).setExamples(examples);
```
frantuma: would call `response.setExamples()` on new line, with no need of modifying setter behaviour in response POJO

modules/swagger-models/src/main/java/io/swagger/models/Response.java
```diff
@@ -67,8 +67,9 @@ public void setSchema(Property schema) {
         return this.examples;
     }
 
-    public void setExamples(Map<String, Object> examples) {
+    public Response setExamples(Map<String, Object> examples) {
         this.examples = examples;
+        return this;
```
frantuma: we don't want to modify setter behaviour, see also comment above

modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
pnrakesh: Sure, I will change that.

modules/swagger-models/src/main/java/io/swagger/models/Response.java
pnrakesh: Sure.

pnrakesh:
There is already an unit test for this feature at `swagger-core\modules\swagger-core\src\test\java\io\swagger\ResponseExamplesTest.java`

frantuma:
included and replaced by #2801

Issue #2326, opened 2017-07-20
Update equals, toString and hashCode methods for models in feature/3.0.0-rc2 branch


a-saf:
- requestBody is missing from “equals”, “hashCode” and “toString” in Link.java
 - Encoding.java does not implement equals and hashCode

This is for branch feature/3.0.0-rc2.

arthurdm:
Closing as fixed

PR #2327, opened 2017-07-20
Update for issue 2326 - swagger models v3

a-saf:
PR for issue #2326


a-saf:
Also: 
- added missing $ref from Header's equals and hashCode
- discriminator to equals and hashCode in Schema
- $ref to equals and hashCode in Parameter

frantuma:
included in #2429 and merged

Issue #2306, opened 2017-07-10
3.0 Annotations need javadoc


arthurdm:
Some of the new 3.0 annotations have javadocs (class level and method level), but a lot of them are missing information, which makes it hard for new users to understand.

janamanoharan:
It would also be helpful if each of the annotations had examples on how to use them so users can have a better understanding of their purpose

MichakrawSB:
Closing this issue as part of our https://github.com/swagger-api/swagger-core/issues/5074.
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.
Thank you for your understanding!

Issue #2272, opened 2017-06-23
Array fields inside annotations should have empty array as their default value


navidsh:
Default value of array fields such as `servers` in `@Operation` are arrays with empty objects which is a bit confusing. An empty array `{}` might be a better default value. 

@Operation: `Server[] servers() default @Server();`

Also fields like `io.swagger.oas.annotations.Operation.externalDocs` also don't need to have a default value.

frantuma:
implemented in #2394

PR #2333, opened 2017-07-25
Annotation enhancements

arthurdm:
PR for issue https://github.com/swagger-api/swagger-core/issues/2306

-Proper documentation annotations (including some fixes to make Java8's compiler happy) and fixing a few errors in them (in terms of the spec)
- Fixing related tests and processing


arthurdm:
I thought there was no need to have the annotation "ArraySchema", since we could easily do things like @Schema(type="array", implementation=POJO.class) without an extra annotation.

arthurdm:
I am currently looking into the CI build failure logs

arthurdm:
@frantuma  / @fehguy  - I fixed the build issues, ready for review now.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
```diff
@@ -16,92 +16,103 @@
...
+   * Allows this parameter to be marked as hidden
+   * @return whether or not this parameter is hidden
    */
   boolean hidden() default false;
+  
+  /**
+   * Provides an array examples of the schema.  When associated with a specific media type, the example string shall be parsed by the consumer to be treated as an object or an array.
+   * @return the list of examples for this parameter
+   **/
+  String[] examples() default {};
```
fehguy: Not sure if we want this here...

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
```diff
@@ -16,92 +16,103 @@
...
+
+  /**
+   * Provides an example of the schema.  When associated with a specific media type, the example string shall be parsed by the consumer to be treated as an object or an array.
+   * @return an example of the parameter
+   **/
+  String example() default "";
```
fehguy: This too...

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/callbacks/Callback.java
```diff
@@ -39,18 +35,21 @@
...
+ */
   String callbackUrlExpression() default "";
 
   /**
-   * the operation that will be executed on callback
+   * The array of operations that will be called out-of band 
+   * @return the callback operations
    **/
-  Operation[] operation() default @Operation();
+  Operation[] operation() default {};
```
fehguy: This is very different from defaulting to `@Operation()`.  We need to see the impact of this

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/links/Link.java
```diff
@@ -38,28 +36,45 @@
...
    **/
   String operationId() default "";
 
   /**
-   * 
+   * Array of parameters to pass to an operation as specified with operationId or identified via operationRef.
+   * @return the list of parameters for this link
    **/
-  LinkParameters parameters() default @LinkParameters();
+  LinkParameter[] parameters() default {};
```
fehguy: Need to see the effect of this, very different than using a default of `@LinkParameters`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/links/Link.java
```diff
@@ -38,28 +36,45 @@
...
+   * A description of the link. CommonMark syntax may be used for rich text representation.
+   * @return the link's description
    **/
   String description() default "";
+  
+  /**
+   * A literal value or {expression} to use as a request body when calling the target operation.
+   * @return the request body of this link
+   **/
+  String requestBody() default "";
```
fehguy: Not sure if this makes any sense.  If a proper `RequestBody`, it should be the annotation object rather than a string.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Content.java
```diff
@@ -23,28 +23,27 @@
...
    **/
   String mediaType() default "";
 
   /**
-   * an array of examples used to show the use of the associated schema
+   * An array of examples used to show the use of the associated schema.
+   * @return the list of examples
    **/
-  ExampleObject[] examples() default @ExampleObject();
+  ExampleObject[] examples() default {};
```
fehguy: This too is a big change.  Need @webron @ralphdoe @gracekarina to look at this

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/DiscriminatorMapping.java
```diff
@@ -16,13 +16,19 @@
...
 /**
  * 
  *
  * 
  **/
-
-
+@Target({  })
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
```
fehguy: @frantuma can you confirm this is OK?

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,230 @@
...
    **/
   Class<?> not() default Void.class;
 
   /**
-   * provides an array of java class implementations which can be used to describe multiple acceptable schemas.  If more than one match the derived schemas, a validation error will occur
+   * Provides an array of java class implementations which can be used to describe multiple acceptable schemas.  If more than one match the derived schemas, a validation error will occur.
+   * @return the list of possible classes for a single match
    **/
-  Class<?>[] oneOf() default Void.class;
+  Class<?>[] oneOf() default {};
```
fehguy: @frantuma is this OK?  Not sure why we would make this change but I trust your input

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,230 @@
...
 
   /**
-   * provides an array of java class implementations which can be used to describe multiple acceptable schemas.  If any match, the schema will be considered valid
+   * Provides an array of java class implementations which can be used to describe multiple acceptable schemas.  If any match, the schema will be considered valid.
+   * @return the list of possible class matches
    **/
-  Class<?>[] anyOf() default Void.class;
+  Class<?>[] anyOf() default {};
```
fehguy: See comment on `oneOf`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,230 @@
...
-//
-//  /**
-//   * determines whether an array of items will be unique
-//   **/
-//  boolean uniqueItems() default false;
+  /**
+   * Sets the maximum number of items in an array.  Ignored if value is Integer.MIN_VALUE.  Used only if type is array.
+   * @return the maximum number of items in this schema
+   **/
+  int maxItems() default Integer.MIN_VALUE;
```
fehguy: @gracekarina @ralphdoe why were these disabled?  Concerns about enabling them?

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,230 @@
...
+
+  /**
+   * Sets the minimum number of items in an array.  Ignored if value is Integer.MAX_VALUE.  Used only if type is array.
+   * @return the minimum number of items in this schema
+   **/
+  int minItems() default Integer.MAX_VALUE;
```
fehguy: Same as above

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,230 @@
...
 
   /**
-   * constrains the number of arbitrary properties when additionalProperties is defined.  Ignored if value is 0.
+   * Determines whether an array of items will be unique.  Used only if type is array.
+   * @return whether or not the items in this schema must be unique
+   **/
+  boolean uniqueItems() default false;
```
fehguy: This lives in the `ArraySchema` and shouldn't go here, unless I'm mistaken

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,230 @@
...
    **/
   int minProperties() default 0;
 
   /**
-   * Allows multiple properties in an object to be marked as required
+   * Allows multiple properties in an object to be marked as required.
+   * @return the list of required schema properties
    **/
-  String[] requiredProperties() default "";
+  String[] requiredProperties() default {};
```
fehguy: Why change the default here @arthurdm ?

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,230 @@
...
+  String example() default "";
 
   /**
-   * provides an example of the schema.  When associated with a specific media type, the example string shall be parsed by the consumer to be treated as an object or an array.
+   * Provides an array examples of the schema.  When associated with a specific media type, the example string shall be parsed by the consumer to be treated as an object or an array.
+   * @return a list of examples of this schema
    **/
-  String example() default "";
-
+  String[] examples() default {};
```
fehguy: @gracekarina @ralphdoe is this change OK?

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,230 @@
...
    **/
   String type() default "";
 
   /**
-   * provides a list of allowable values.
+   * Provides a list of allowable values.  This field map to the enum property in the OAS schema.
+   * @return a list of allowed schema values
    */
-  String[] _enum() default {""};
+  String[] allowableValues() default {};
```
fehguy: @arthurdm why deviate from the naming convention previously used?  It doesn't follow the schema property names.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,230 @@
...
 
   /**
-   * provides a default value.
+   * Provides a default value.
+   * @return the default value of this schema
    */
-  String _default() default "";
+  String defaultValue() default "";
```
fehguy: Same as above.  Why make this change?  It's confusing.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/parameters/Parameters.java
```diff
@@ -1,19 +0,0 @@
-package io.swagger.oas.annotations.parameters;
```
fehguy: @arthurdm why delete this?

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/responses/ApiResponse.java
```diff
@@ -26,35 +26,31 @@
...
+     *A short description of the response. 
      **/
     String description() default "";
 
     /**
-     * the HTTP response code, or default, for the supplied response
+     * The HTTP response code, or 'default', for the supplied response.  May only have 1 default entry.  
      **/
-    String responseCode() default "";
+    String responseCode() default "default";
```
fehguy: Why???

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/parameters/RequestBody.java
```diff
@@ -25,28 +25,24 @@
...
+   * A brief description of the request body.
    **/
   String description() default "";
 
   /**
-   * 
+   * The content of the request body.
    **/
-  Content[] content() default @Content();
+  Content[] content() default {};
```
fehguy: Why do this???

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/security/OAuthFlow.java
```diff
@@ -23,34 +23,30 @@
...
+   * The URL to be used for obtaining refresh tokens. This must be in the form of a URL.  Applies to oauth2 type.
    **/
   String refreshUrl() default "";
 
   /**
-   * 
+   * Required. The available scopes for the OAuth2 security scheme.  Applies to oauth2 type.
    **/
-  Scopes scopes() default @Scopes;
+  OAuthScope[] scopes() default {};
```
fehguy: Is this just a naming change plus change in default?  If so, I get it, but this deviates from the naming in the spec, which will create confusion

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/security/SecurityScheme.java
```diff
@@ -46,33 +42,33 @@
   String description() default "";
 
   /**
-   * Required. The name of the header or query parameter to be used.
+   * Required. The name of the header or query parameter to be used. Applies to apiKey type.
    **/
-  String name();
+  String name() default "";
```
fehguy: I don't get it.  @arthurdm your PR is moving away from defaults but adding here? Why?

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/servers/ServerVariable.java
```diff
@@ -22,33 +22,29 @@
...
 @Target({  })
 @Retention(RetentionPolicy.RUNTIME)
 @Inherited
 public @interface ServerVariable {
   /**
-   * 
+   * Required.  The name of this variable.
    **/
-  String name() default "";
+  String name();
```
fehguy: Same as above.  Now no default?  What does that mean for usability?

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/servers/ServerVariable.java
```diff
@@ -22,33 +22,29 @@
...
+   * An array of allowable values for this variable.  This field map to the enum property in the OAS schema.
    **/
   String[] allowableValues() default "";
 
   /**
-   * the value attribute maps to the default property in the OAS schema
+   * Required.  The default value of this variable.
    **/
-  String value() default "";
+  String defaultValue();
```
fehguy: Don't like the name change

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/responses/ApiResponse.java
arthurdm: in this particular case, the default responseCode is named "default" as per the spec, so I thought it would be good to have that as the default.

arthurdm:
thanks for the review @fehguy !  I will go through them individually.  As a summary of the thought process here though:

- having the default of an array as a single object didn't feel right.  For example:  String[] field default "" doesn't feel right, it really should be String[] field {}, no?     Unless I missed something, we want the annotation processor to know that this is an empty array, not a singleton instance with default values (which is something different).

- I can see your point about deviating from the spec by having "allowableValues" instead of "_enum", and "defaultValue" instead of "_default".  My intention is to make this more natural for a Java programmer to use (we don't see any JSR annotations leading with underscore) while still being close enough to the spec that it's clear what the purpose of the field is.  

- Why do we need ArraySchema annotation, but don't need all the other types of annotations?  I tried to show in the PR that we can get away with a simple single Schema.  For you can do things like @Schema(type="array", implementation= MyPojo.class).

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/callbacks/Callback.java
arthurdm: See issue https://github.com/swagger-api/swagger-core/issues/2272.    I am curious as to why we would want the default of an array of operations to be a single empty operation?  Shouldn't be an empty array?

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/links/Link.java
arthurdm: Same as comment above

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Content.java
arthurdm: Same as the other array comments

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
arthurdm: Same array comment

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
arthurdm: See comment for `Operation[]`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
arthurdm: See comment for `maxItems`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
arthurdm: See comment for `maxItems`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
arthurdm: See comment for `Operation[]`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
arthurdm: See comment for `Operation[]`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
arthurdm: To make it more natural for Java developers.  Starting with underscore brings a connotation of something being private.   I believe the name `allowableValues` is pretty clear on its intention, and I also updated the javadoc to say `This field map to the enum property in the OAS schema.`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
arthurdm: Same reasoning as `_enum`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/parameters/Parameters.java
arthurdm: it was redundant.  We just needed to use `Parameter[]`.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/parameters/RequestBody.java
arthurdm: Same reasoning as `Operation[]` above

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/security/OAuthFlow.java
arthurdm: All the other security annotations for OAuth started with OAuth, so I thought this was more consistent, and clear that it is only for OAuth - not random ACL scopes.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/security/SecurityScheme.java
arthurdm: In this case we needed a default value because `name` only applies for apiKeys.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/servers/ServerVariable.java
arthurdm: This is required in all cases, so no default needed.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/servers/ServerVariable.java
arthurdm: This maps to the Server Variable `default` field: https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#serverVariableObject

So in that sense, I felt that `defaultValue` better represents that this is a default value than just using `value`.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
arthurdm: The Parameter object in the spec has an `examples` field: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameterObject

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
arthurdm: The Parameter object in the spec has an `example` field: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameterObject

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/links/Link.java
arthurdm: Although this has the same name, I believe it is different than the spec's definition of a `Request Body` object.  For a link, the `requestBody` field's description is  `A literal value or {expression} to use as a request body when calling the target operation.` which implies just a simple string, and not a full blow `Request Body` (which has a description, a required boolean and an array of content objects)

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/DiscriminatorMapping.java
frantuma: it's ok, possibly removing `@Inherited` which doesn't apply to anything but classes

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
frantuma: this is ok

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
frantuma: this is ok

arthurdm:
@fehguy - please let me know if you agree with my replies to the review comments.

fehguy:
@arthurdm, I've tagged a few people to give the feedback on.  I'll add them to the reviewers list--since they've done the bulk of the implementation, I'm guessing they will have feedback on the items, or at least should know that things are being changed.

As far as moving away `ArraySchema` and merging the fields into the `Schema` class, we did this because annotations do not allow circular references, hence you cannot have this:

```java
@Schema(type = "array", items = @Schema(...))
```

That was a deal breaker, because using `@Schema` as an array type meant it would only work with primitive types, not complex ones using a `@Schema` annotation object.  I don't think it would be smart to remove that capability.

arthurdm:
good point @fehguy - I have added ArraySchema back into the PR.

arthurdm:
@frantuma @gracekarina @ralphdoe  - please let me know if you have any review comments about this PR.

arthurdm:
@webron - hey Ron, could you please help move this one forward?

webron:
@arthurdm - One more thing - I reviewed mostly the annotations. Do the other changes simply cover the naming changes in the annotation or do they change any additional logic?

arthurdm:
Thanks @webron !  I will go through the comments today.

Yes, only annotation changes and the corresponding changes for names, with the exception of io/swagger/jaxrs2/OperationParser.java which contains a change to handle the Content[] from ApiResponse (was previously just Content).

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
```diff
@@ -16,93 +16,113 @@
...
+   * @return the parameter's name
    **/
   String name() default "";
 
   /**
-   * the location of the parameter.  Ignored when empty string
+   * The location of the parameter.  Possible values are "query", "header", "path" or "cookie".  Ignored when empty string.
+   * @return the parameter's location
    **/
   String in() default "";
```
arthurdm: I agree @webron  I opened this new issue for this: https://github.com/swagger-api/swagger-core/issues/2371

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
```diff
@@ -16,93 +16,113 @@
...
    * Additional description data to provide on the purpose of the parameter
+   * @return the parameter's description
    **/
   String description() default "";
 
   /**
-   * Specifies that the parameter is not optional and must be present
+   * Specifies that the parameter is not optional and must be present.
+   * @return whether or not the parameter is required
    **/
```
arthurdm: agreed - I will clarify this in the javadoc

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
```diff
@@ -16,93 +16,113 @@
...
+   * @return whether or not the parameter allows empty values
    **/
   boolean allowEmptyValue() default false;
 
   /**
-   * 
+   * Describes how the parameter value will be serialized depending on the type of the parameter value. Default values (based on value of in): for query - form; for path - simple; for header - simple; for cookie - form.
+   * @return the style of the parameter
    **/
   String style() default "";
```
arthurdm: Cool, I opened https://github.com/swagger-api/swagger-core/issues/2372 for the style enum investigation.   

Fixing the javadoc to clarify the exclusion with `content`

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
```diff
@@ -16,93 +16,113 @@
...
   Schema schema() default @Schema();
-
-
+  
+  /**
+   * The schema of the array that defines this parameter.
+   * 
+   * @return the schema of the array
+   */
   ArraySchema array() default @ArraySchema();
```
arthurdm: Sounds good - I am clarifying this in a new commit.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
```diff
@@ -16,93 +16,113 @@
...
 
-  // TODO #2312 single object, according to spec
   /**
-   * 
+   * The representation of this parameter, for different media types.
+   * @return the content of the parameter
    **/
-  Content[] content() default @Content();
+  Content[] content() default {};
```
arthurdm: Clarifying it in new commit

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/callbacks/Callback.java
```diff
@@ -25,12 +25,8 @@
 import io.swagger.oas.annotations.Operation;
 
 /**
- * 
- *
- * 
+ * This object represents a webhook that will be invoked.
```
arthurdm: changing it

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/links/Link.java
```diff
@@ -38,28 +36,45 @@
...
+   * @return an operation reference
    **/
   String operationRef() default "";
 
   /**
-   * 
+   * The name of an existing, resolvable OAS operation, as defined with a unique operationId. This field is mutually exclusive of the operationRef field.
+   * @return an operation ID
    **/
   String operationId() default "";
```
arthurdm: changed

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/ExampleObject.java
```diff
@@ -24,33 +24,33 @@
...
+   @return the value of the example
    **/
   String value() default "";
 
   /**
-   * a URL to point to an external document to be used as an example.  This is mutually exclusive with the value property
+   * A URL to point to an external document to be used as an example.  This is mutually exclusive with the value property
+   * @return an external URL of the example
    **/
   String externalValue() default "";
```
arthurdm: changing it

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,212 @@
 @Inherited
 public @interface Schema {
   /**
-   * provides a java class as implementation for this schema.  When provided, additional information in the Schema annotation will augment the java class after introspection
+   * Provides a java class as implementation for this schema.  When provided, additional information in the Schema annotation will augment the java class after introspection.
+   * @return a class that implements this schema
```
arthurdm: clarifying

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,212 @@
...
    **/
-  Class<?>[] anyOf() default Void.class;
+  Class<?>[] anyOf() default {};
 
   /**
-   * provides an array of java class implementations which can be used to describe multiple acceptable schemas.  If all match, the schema will be considered valid
+   * Provides an array of java class implementations which can be used to describe multiple acceptable schemas.  If all match, the schema will be considered valid
+   * @return the list of classes to match   
    **/
   Class<?>[] allOf() default Void.class;
```
arthurdm: yup, good catch.  I missed that one.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,212 @@
...
+   * @return the title of the schema
    **/
   String title() default "";
 
   /**
-   * constrains a value such that when divided by the multipleOf, the remainder must be an integer.
+   * Constrains a value such that when divided by the multipleOf, the remainder must be an integer.  Ignored if the value is 0.
+   * @return the multiplier constraint of the schema
    **/
   int multipleOf() default 0;
```
arthurdm: good point, let's go with double.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,212 @@
...
   /**
-   * if true, makes the maximum value exclusive, or a less-than criteria
+   * if true, makes the maximum value exclusive, or a less-than criteria.
+   * @return the exclusive maximum value for this schema
    **/
   boolean exclusiveMaximum() default false;
 
   /**
-   * if true, makes the maximum value exclusive, or a greater-than criteria
+   * Sets the minimum numeric value for a property.  Ignored if the value is an empty string.
```
arthurdm: clarifying it in the javadoc

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,212 @@
...
+   * @return the exclusive minimum value for this schema
    **/
   boolean exclusiveMinimum() default false;
 
   /**
-   * sets the maximum length of a string value.  Ignored if the value is Integer.MIN_VALUE.
+   * Sets the maximum length of a string value.  Ignored if the value is Integer.MIN_VALUE.
+   * @return the maximum length of this schema
    **/
   int maxLength() default Integer.MIN_VALUE;
```
arthurdm: agreed, changing.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/media/Schema.java
```diff
@@ -39,192 +35,212 @@
...
 
   /**
-   * sets the minimum length of a string value.  Ignored if the value is Integer.MAX_VALUE.
+   * Sets the minimum length of a string value.  Ignored if the value is Integer.MAX_VALUE.
+   * @return the minimum length of this schema
    **/
```
arthurdm: agreed, changing.

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/security/OAuthFlow.java
```diff
@@ -23,34 +23,30 @@
...
  **/
-
-
 @Target({  })
 @Retention(RetentionPolicy.RUNTIME)
 @Inherited
 public @interface OAuthFlow {
   /**
-   * 
+   * Required. The authorization URL to be used for this flow. This must be in the form of a URL.  Applies to oauth2 ("implicit", "authorizationCode") type.
```
arthurdm: agreed

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/security/SecurityRequirement.java
```diff
@@ -23,18 +23,20 @@
...
-
-
 @Target({ ElementType.METHOD,
           ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @Inherited
 public @interface SecurityRequirement {
-    String name() default "";
+	/**
+	 * Required. This name must correspond to a declared SecurityRequirement.
```
arthurdm: agreed

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/security/SecurityScheme.java
```diff
@@ -46,33 +42,33 @@
...
 
   /**
-   * Required. The location of the API key. Valid values are "query" or "header".
+   * Required. The location of the API key. Valid values are "query" or "header".  Applies to apiKey type.
    **/
   String in() default "";
```
arthurdm: Opened issue https://github.com/swagger-api/swagger-core/issues/2373 for those

arthurdm:
Addressed review comments

arthurdm:
@webron - ready for final review and merge.

arthurdm:
I created another PR to target the new branch `2.0`, https://github.com/swagger-api/swagger-core/pull/2394, so closing this PR

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/links/Link.java
```diff
@@ -38,28 +36,45 @@
...
-   * 
+   * A relative or absolute reference to an OAS operation. This field is mutually exclusive of the operationId field, and must point to an Operation Object. Relative operationRef values may be used to locate an existing Operation Object in the OpenAPI definition.
+   * @return an operation reference
    **/
   String operationRef() default "";
 
   /**
-   * 
+   * The name of an existing, resolvable OAS operation, as defined with a unique operationId. This field is mutually exclusive of the operationRef field.
+   * @return an operation ID
```
ralphdoe: In the current behavior we are setting both fields individually, there's no exclusivity restriction. Issue created to treat this: #2395

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
ralphdoe: This isn't implemented yet, also added to the issue: #2395

PR #2349, opened 2017-08-03
Revert pointless commit to fix #2320

### LINKED ISSUE #2320: ApiListingResource doesn't respect pretty print support
I configured Swagger using `BeanConfig` on my `Application` start up to use pretty print, but when I retrieve the API from `swagger.json` endpoint it's all on a single line.

I've tracked the issue down to the [ApiListingResource](https://github.com/swagger-api/swagger-core/blob/222ef109c83c56a287e59207eb7f6dc794e6ae9d/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/listing/ApiListingResource.java#L1). When this resource processes the API, it doesn't return a Swagger object on the Response (that would be then converted by [SwaggerSerializers](https://github.com/swagger-api/swagger-core/blob/222ef109c83c56a287e59207eb7f6dc794e6ae9d/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/listing/SwaggerSerializers.java)), but returns a text string instead.

If `ApiListingResource` responds with the Swagger object instead, the serialization works fine and respect the pretty print request. Could the `getListing()` code be changed from 

``` java
        if (StringUtils.isNotBlank(type) && type.trim().equalsIgnoreCase("yaml")) {
            return getListingYamlResponse(app, context, sc, headers, uriInfo);
        } else {
            return getListingJsonResponse(app, context, sc, headers, uriInfo);
        }
```

to 

``` java
		Swagger swagger = process(app, context, sc, headers, uriInfo);
		if(swagger == null) {
			return Response.status(Status.NOT_FOUND).build();
		} else {
			return Response.ok(swagger, "json".equals(type) ? MediaType.APPLICATION_JSON_TYPE : new MediaType("application", "yaml")).build();
		}
```
or would it affect some functionality I'm not aware of?

ChristianCiach:
Revert "use custom object mapper to serialize json in order to avoid
null values"

This reverts commit 3ad759596f321ffb195256e045ec5803c00ee53c.

This fixes issue #2320. The original commit is pointless and dangerous,
as it surprisingly overrides the behavior of SwaggerSerializers.java,
but does essentially the same (but without supporting pretty printing).

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/listing/AcceptHeaderApiListingResource.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/listing/AcceptHeaderApiListingResource.java
@@ -1,6 +1,5 @@
 package io.swagger.jaxrs.listing;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import io.swagger.annotations.ApiOperation;
 
 import javax.servlet.ServletConfig;
@@ -26,7 +25,7 @@ public class AcceptHeaderApiListingResource extends BaseApiListingResource {
             @Context Application app,
             @Context ServletConfig sc,
             @Context HttpHeaders headers,
-            @Context UriInfo uriInfo) throws JsonProcessingException {
+            @Context UriInfo uriInfo) {
         return getListingJsonResponse(app, context, sc, headers, uriInfo);
     }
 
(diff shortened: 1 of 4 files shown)


ChristianCiach:
I've added a second commit to remove a similar issue for YAML documents, where SwaggerSerializers.java has been ignored, too. Also, I've removed some split-and-then-rejoin-code that has been pointless since ages.

ChristianCiach:
Is there any reason not to merge this?

tobilarscheid:
I see the issue with #2320, and I absolutely understand how this commit fixes it. However, your fix now introduces a regression to #2105 
/cc @fehguy

Issue #2369, opened 2017-08-14
BeanParam and MatrixParam Support


ralphdoe:
There's no support for that kind of parameters.
 
Add some tests to ensure this changes are working fine.

frantuma:
closing as BeanParam implemented, Matrix param part of further ticket

PR #2370, opened 2017-08-15
Issue 2369 - BeanParam and MatixParam added

ralphdoe:
Solution to the issue #2369 

Now we consider the BeanParam and MatrixParam Annotations.


frantuma:
closing as BeanParam is implemented, and proposed solution for MatrixParam not applicable

Issue #1813, opened 2016-06-01
Swagger-servlet includes logback-test.xml in classpath


tirsoh:
Looks like the swagger-servlet 1.5.9 dependency includes a logback-test.xml file on the classpath. This is not a good practice because logback will pick this logback file over the consumers logback.xml configuration. Can someone please remove this from the final JAR so that it doesn't override consumer's logback configuration?

junegod:
+1

phomlish:
+1

fehguy:
There is an exclusion for logback now

ajay-kmr:
Hi,
I am still facing this issue while using below dependency:-
compile group: 'io.swagger', name: 'swagger-models', version: "1.5.15"
compile group: 'io.swagger', name: 'swagger-servlet', version: "1.5.15"

datagitlies:
I'm also facing this issue. I see in the pom file for swagger-servlet the `<exclude>logback.xml</exclude>` but swagger-servlet contains a `logback-test.xml` file in `/src/main/resources`

phomlish:
try this:
`		<dependency>
			<groupId>io.swagger</groupId>
			<artifactId>swagger-servlet</artifactId>
			<version>${swagger.version}</version>
			<exclusions>
				<exclusion>
					<groupId>ch.qos.logback</groupId>
					<artifactId>logback-api</artifactId>
				</exclusion>
				<exclusion>
					<groupId>ch.qos.logback</groupId>
					<artifactId>logback-core</artifactId>
				</exclusion>
				<exclusion>
					<groupId>ch.qos.logback</groupId>
					<artifactId>logback-access</artifactId>
				</exclusion>
			</exclusions>
		</dependency>
`

datagitlies:
@phomlish that doesn't work as it's not the logback dependency causing the problem, it's the physical file `logback-text.xml` that exists in `/src/main/resources` which is then being packaged with `swagger-servlet-1.5.16.jar`. If I have time I'll fork `swagger-servlet` and open a pull request that simply moves the file from `/src/main/resources` to `/test/main/resources` which will fix the problem.

datagitlies:
@fehguy since you were last to close this issue... I'm wondering, given the information here and the pull request, whether you still believe the issue should be considered "closed"

binlecode:
@fehguy is there any update on this issue... it would be great to have a fix so that the hosting application no longer needs to override logback config file location... Thanks.

Issue #709, opened 2014-10-01
swagger modules should not contain logback configuration


francisdb:
Logback.xml config should not be provided with library jars.
This file appears in logback core somewhere between 1.3.6 and 1.3.9

If you want to disable logging for tests you should place a logback-test.xml under the src/test/resources folder as explained in:
http://logback.qos.ch/manual/configuration.html

Also why does the project depend on logback in provided scope? Libraries should only depend on slf4j and not implementations.

PR #2402, opened 2017-08-25
move logback-test.xml to completely resolve issues 1813 & 709

datagitlies:
see #1813 & #709


Spectrality:
Any reason this hasn't been approved/pulled? This file is causing major issues for anyone using Swagger that also wants to use their own logback file. Seems to me the logback-test.xml file is currently in the wrong place, since it would seem it should be in the test resources.

Spectrality:
@frantuma

Spectrality:
@frantuma any chance you could update the swagger-project/swagger-servlet version to 1.5.17 to push the change to Maven repo? Or do you have a timeframe for when that might happen?

Issue #2403, opened 2017-08-25
@RequestBody should be populated with @Consumes from method or class when mediaType is not specified


arturdzm:
Since the RequestBody, ApiResponse and Parameter all have content
I think we can have 1 method called 
getContent(Content[] content, String[] mediaTypes) 
so each type can pass in value() of @Consumes or @Produces
or null if no mediaTypes  specified. 
if mediaTypes is null set it to `*/*`

ralphdoe:
You're totally right, @arturdzm. I have solved this here: https://github.com/swagger-api/swagger-core/pull/2404

frantuma:
closing as implemented in related PRs

Issue #2395, opened 2017-08-23
bugfix/reader-annotations-improvement


ralphdoe:
- [x] In Links (operationId - operationRef): 
 - operationId and operationRef are mutually exclusive.
 - in case the operationId and operationRef are both filled, the operationId has the priority.

- [x] Schema annotation: 
 - The maximum is not being treated. 
 - The minimum is not being treated. 
 - The value must be ignored if it's empty or it's not a number, considering the field is a String. 

- [x] Parameter annotation:
 - The schema, content and arraySchema must be mutually exclusive.
 - In case they are all filled, the priority must be the next: content > array > schema.

- [x] ArraSchema in Parameter annotation:
 - The ArraySchema Annotation hasn't been implemented yet.
 - Adding some tests about this. 

- [x] Contents updates:
 - Method getContents deleted. 
 - Method getContent improved, solving this issue: #2403. 
-  applyProduces replaced by applyMediaType (now receives String arrays)

Everything solved in this PR: https://github.com/swagger-api/swagger-core/pull/2404

frantuma:
closing as implemented in related PRs

PR #2404, opened 2017-08-27
Bugfx/2395 reader annotations improvment

ralphdoe:
Bugfix: #2395 
- [x] In Links (operationId - operationRef): 
 - operationId and operationRef are mutually exclusive.
 - in case the operationId and operationRef are both filled, the operationId has the priority.

- [x] Schema annotation: 
 - The maximum is not being treated. 
 - The minimum is not being treated. 
 - The value must be ignored if it's empty or it's not a number, considering the field is a String. 

- [x] Parameter annotation:
 - The schema, content and arraySchema must be mutually exclusive.
 - In case they are all filled, the priority must be the next: content > array > schema.

- [x] ArraSchema in Parameter annotation:
 - The ArraySchema Annotation hasn't been implemented yet.
 - Adding some tests about this. 

- [x] Contents updates:
 - Method getContents deleted. 
 - Method getContent improved, solving this issue: #2403. 
-  applyProduces replaced by applyMediaType (now receives String arrays)

Bugfix: #2403 
Since the RequestBody, ApiResponse and Parameter all have content
I think we can have 1 method called
getContent(Content[] content, String[] mediaTypes)
so each type can pass in value() of @consumes or @Produces
or null if no mediaTypes specified.
if mediaTypes is null set it to "/


modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/Reader.java
```diff
@@ -280,28 +280,12 @@ protected void processParameter(List<Parameter> parameters, Operation operation,
...
-                                        for (String value : classConsumes.value()) {
-                                            setMediaTypeToContent(parameter.getSchema(), content, value);
-                                        }
-                                    } else {
-                                        setMediaTypeToContent(parameter.getSchema(), content, DEFAULT_MEDIA_TYPE_VALUE);
-                                    }
-
-                                    requestBody.setContent(content);
-                                }
-                                if (parameter.getSchema() != null) {
```
frantuma: Why has the above been removed? its purpose was populating request body schema/content from parameter schema if not set in annotation.

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/annotations/parameters/ParametersTest.java
```diff
@@ -53,6 +55,16 @@ public void testSimpleCallback() {
...
+                "      - name: arrayParameter\n" +
+                "        in: query\n" +
+                "        required: true\n" +
+                "        content:\n" +
+                "          '*/*':\n" +
+                "            schema:\n" +
+                "              type: int\n" +
+                "              description: the generated id\n" +
+                "              format: id\n" +
+                "              readOnly: true\n" +
```
frantuma: what should be taking precedence (array or content or schema)? did you discuss this with @webron?

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/annotations/parameters/ParametersTest.java
```diff
@@ -98,7 +110,21 @@ public void testSimpleCallback() {
...
+                                ),
+                                schema = @Schema(
+                                        type = "int",
+                                        format = "id",
+                                        description = "the generated id",
+                                        readOnly = true),
+                                content = @Content(schema = @Schema(type = "int",
+                                        format = "id",
+                                        description = "the generated id",
+                                        readOnly = true)))
```
frantuma: what should be taking precedence (array or content or schema)? did you discuss this with @webron?

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/annotations/requests/RequestBodyTest.java
```diff
@@ -237,6 +237,7 @@ public Response methodWithRequestBodyWithoutAnnotation(
         @Path("/user")
         @Operation(summary = "Delete user",
                 description = "This can only be done by the logged in user.")
+        @Consumes({"application/json", "application/xml"})
```
frantuma: why has this be added? this is not needed for the test and should be removed I'd say

ralphdoe: Done!

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/Reader.java
ralphdoe: This is now treated inside the getContent method, the consumes and produces are now treated as String[] mediaTypes.

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/annotations/parameters/ParametersTest.java
ralphdoe: https://github.com/swagger-api/swagger-core/pull/2416 Suggested by Ron in this PR.

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/annotations/parameters/ParametersTest.java
ralphdoe: #2416 Suggested by Ron in this PR.

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/Reader.java
frantuma: @ralphdoe Can you point me to code implementing the above? getContent doesn't seem to use parameter schema if not set in annotation, but maybe I am wrong

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/annotations/parameters/ParametersTest.java
frantuma: @ralphdoe What do you mean? I couldn't find any reference there about the question above

ralphdoe: I meant this PR: https://github.com/swagger-api/swagger-core/pull/2333.
webron 28 days ago Owner
We should specify this, schema and content are mutually exclusive, and describe the behavior if more than one is specified. I'd suggest content > array > schema. @ralphdoe - what's the current behavior of the reader?

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/Reader.java
ralphdoe: This is the getRequestBody method in OperationParser: https://github.com/swagger-api/swagger-core/blob/bugfx/2395-reader-annotations-improvent/modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java#L341-L360. Also you can see: https://github.com/swagger-api/swagger-core/blob/bugfx/2395-reader-annotations-improvent/modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java#L452-L465

frantuma: @ralphdoe I might be wrong, but that code seems to populate requestBody with values from @RequestBody and @Consumes annotations, which is not what the removed code did, which was applying the so far resolved parameter schema to requestBody if requestBody doesn't specify it

ralphdoe: My bad, code recovered.

PR #2406, opened 2017-08-29
fix examples map on header

gracekarina:
Examples in header should be a map instead of a List, header should follow the structure of a parameter as says in:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#header-object

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/oas/models/headers/Header.java
+++ b/modules/swagger-models/src/main/java/io/swagger/oas/models/headers/Header.java
@@ -218,7 +218,7 @@ public class Header {
     return this;
   }
 
-  public Header addExamples(String key, Example examplesItem) {
+  public Header addExample(String key, Example examplesItem) {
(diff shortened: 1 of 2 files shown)


modules/swagger-models/src/main/java/io/swagger/oas/models/headers/Header.java
```diff
@@ -196,30 +198,31 @@ public Header schema(Schema schema) {
...
-  public Header examples(List<Example> examples) {
+  public Header examples(Map<String, Example> examples) {
     this.examples = examples;
     return this;
   }
 
-  public Header addExamplesItem(Example examplesItem) {
-    if(this.examples == null) {
-      this.examples = new ArrayList<Example>();
+  public Header addExamples(String key, Example examplesItem) {
```
webron: This should be `addExample` (and should be changed in the Parameter as well).

Issue #1847, opened 2016-07-07
When XmlAccessType.NONE is annotated, values annotated with XmlAttribute are not parsed


wapolinar:
After the implementation of #1649, values annotated (in Java) with XmlAttribute are not parsed and must be annotated with XmlElement (which is a nice workaround). As far as I understand it, the annotation of XmlAttribute should also result in parsing the value.
````
@XmlRootElement
@XmlAccessorType(XmlAccessType.NONE)
@ApiModel(description = "A class that describes a user profile")
public class Profile {
	 
	@XmlAttribute
	@ApiModelProperty(value = "The name", required = true)
	private String name;
	
	@XmlElement
	@ApiModelProperty(value = "The address", required = true)
	private String address;

        private String secretKey;
}
````
Here, address would be parsed, the name would not be parsed.

fehguy:
This is a bug--if you have time for a PR for a test and/or fix, it'd be appreciated.

clovis-guillemot:
Same issue for me. Any workaround for this ?

wapolinar:
Annotate with XmlElement instead. This should work.

clovis-guillemot:
@fehguy can you please merge this PR #2409 to fix this issue. Thanks !

PR #2409, opened 2017-08-30
fix 1847-XmlAttribute are not parsed when XmlAccessType.NONE

clovis-guillemot:
This commit fix bug #1847 
When XmlAccessType.NONE is annotated, values annotated with XmlAttribute are not parsed


clovis-guillemot:
@webron @frantuma : this PR should resolve 2 issues #1847 and #2487 
I just resolved some conflicts, can you have a look on this PR please.

PR #2415, opened 2017-08-31
Make Reader and ServletConfig fields protected in BeanConfig

vanta:
Fixing #1884


vanta:
Is anyone reviewing PRs here?

Issue #2412, opened 2017-08-31
Examples Implementation


ralphdoe:
It's necessary to implement the examples treatment in the reader.
 - [x] MediaType Object
 - [x] Parameter Object

frantuma:
implemented in #2416

PR #2416, opened 2017-09-02
Bugfix/issue 2412 examples implementation

ralphdoe:
Examples annotation and examples object.
#2412

### DIFF
--- a/modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
@@ -116,8 +116,8 @@ public @interface Parameter {
   boolean hidden() default false;
 
   /**
-   * An array of examples used to show the use of the associated schema.
-   * @return the list of examples
+   * An array of examples  of the schema used to show the use of the associated schema.
+   * @return array of examples of the parameter
    **/
   ExampleObject[] examples() default {};
 
--- a/modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
@@ -25,6 +25,7 @@ import java.lang.annotation.Target;
 import io.swagger.oas.annotations.enums.Explode;
 import io.swagger.oas.annotations.media.ArraySchema;
 import io.swagger.oas.annotations.media.Content;
+import io.swagger.oas.annotations.media.ExampleObject;
 import io.swagger.oas.annotations.media.Schema;
 
 /**
@@ -113,12 +114,12 @@ public @interface Parameter {
    * @return whether or not this parameter is hidden
    */
   boolean hidden() default false;
-  
+
   /**
-   * Provides an array examples of the schema.  When associated with a specific media type, the example string shall be parsed by the consumer to be treated as an object or an array.  Ignored if the properties content or array are specified.
-   * @return the list of examples for this parameter
+   * An array of examples used to show the use of the associated schema.
+   * @return the list of examples
    **/
-  String[] examples() default {};
+  ExampleObject[] examples() default {};
 
   /**
    * Provides an example of the schema.  When associated with a specific media type, the example string shall be parsed by the consumer to be treated as an object or an array.  Ignored if the properties examples, content or array are specified.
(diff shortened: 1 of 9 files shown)


modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
```diff
@@ -113,12 +114,12 @@
    * @return whether or not this parameter is hidden
    */
   boolean hidden() default false;
-  
+
   /**
-   * Provides an array examples of the schema.  When associated with a specific media type, the example string shall be parsed by the consumer to be treated as an object or an array.  Ignored if the properties content or array are specified.
-   * @return the list of examples for this parameter
+   * An array of examples used to show the use of the associated schema.
```
frantuma: change to "An array of examples of the schema"

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
```diff
@@ -113,12 +114,12 @@
...
+   * @return the list of examples
```
frantuma: change to @return array of examples of the parameter

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
```diff
@@ -78,6 +78,10 @@
             parameterObject.setIn(parameter.in());
             isEmpty = false;
         }
+        if (StringUtils.isNotBlank(parameter.example())) {
+            parameterObject.setIn(parameter.example());
```
frantuma: I guess this is wrong, would be `parameterObject.setParameter` instead? also it means no tests current cover the case, and should be added

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
```diff
@@ -422,35 +436,6 @@ public static void applyProduces(Produces classProduces, Produces methodProduces
         return Optional.empty();
     }
 
-    public static Optional<MediaType> getMediaType(MediaType mediaType, ExampleObject example) {
```
frantuma: remove related unused imports IOException and Json

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/resources/PetResource.java
```diff
@@ -21,6 +21,7 @@
 import io.swagger.oas.annotations.Operation;
 import io.swagger.oas.annotations.Parameter;
 import io.swagger.oas.annotations.media.Content;
+import io.swagger.oas.annotations.media.ExampleObject;
```
frantuma: unused import

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
ralphdoe: Changed!

modules/swagger-annotations/src/main/java/io/swagger/oas/annotations/Parameter.java
ralphdoe: Changed!

modules/swagger-jaxrs2/src/test/java/io/swagger/jaxrs2/resources/PetResource.java
ralphdoe: Removed!

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
ralphdoe: Removed!

modules/swagger-jaxrs2/src/main/java/io/swagger/jaxrs2/OperationParser.java
ralphdoe: Replaced by parameterObject.setExample and test added.

frantuma:
@ralphdoe examples related code looks good, the one thing is that this branch/pr seems to also include commits from #2404 which is still in discussion, not sure why you added them here also; are these changes needed for the example related stuff? if so we must wait for finalization of #2404, otherwise can you remove those, so to keep here only stuff related to examples?

PR #2423, opened 2017-09-12
Feature/spec filter 3.0

ralphdoe:
Spec Filter New Version
Classes, Methods, Variables Renaming.
New OpenApiSpecFitler Interface
AbsctractFilter implementation
Starting the new SpecFilter Implementation
Filters added:
 - OpenAPI
 - Operations
 - Parameters
 - PathItem
 - Schemas related with Parameters
 - Tests added


modules/swagger-core/src/main/java/io/swagger/core/filter/SpecFilter.java
```diff
@@ -4,314 +4,120 @@
...
+                filteredPathItem.setHead(filterOperation(filter, pathItem.getHead(), resourcePath, HEAD, params, cookies, headers));
+                filteredPathItem.setOptions(filterOperation(filter, pathItem.getOptions(), resourcePath, OPTIONS, params, cookies, headers));
+                filteredPathItem.setDescription(pathItem.getDescription());
+                filteredPathItem.set$ref(pathItem.get$ref());
+                filteredPathItem.setSummary(pathItem.getSummary());
+                clonedPaths.addPathItem(resourcePath, filteredPathItem);
             }
+            clone.paths(clonedPaths);
+            clone.setComponents(openAPI.getComponents());
+            clone.setSecurity(openAPI.getSecurity());
```
frantuma: these should be called outside the loop

frantuma: also other elements are not cloned? e.g. extensions, externalDocs, servers

modules/swagger-core/src/main/java/io/swagger/core/filter/SpecFilter.java
```diff
@@ -4,314 +4,120 @@
...
+            if (filteredOperation.isPresent()) {
+                List<Parameter> filteredParameters = new ArrayList<>();
+                Operation filteredOperationGet = filteredOperation.get();
+                List<Parameter> parameters = filteredOperationGet.getParameters();
+                if (parameters != null) {
+                    for (Parameter parameter : parameters) {
+                        Parameter filteredParameter = filterParameter(filter, operation, parameter, resourcePath, key, params, cookies, headers);
+                        if (filteredParameter != null) {
+                            filteredParameter.setSchema(filterProperty(filter, parameter.getSchema(), resourcePath, key, params, cookies, headers));
+                            filteredParameters.add(filteredParameter);
```
frantuma: see also general review comment, filterProperty should be used for schemas inside components also

modules/swagger-core/src/main/java/io/swagger/core/filter/SpecFilter.java
ralphdoe: Done!

frantuma:
included in #2452

PR #2426, opened 2017-09-13
Support the JsonValue annotation

liam-jackson-hpe:
Currently the very useful `@JsonValue` annotation isn't supported by swagger-core. Schemas generated by swagger core will **not** be correct for the output of serializing models making use of this annotation.

https://fasterxml.github.io/jackson-annotations/javadoc/2.0.0/com/fasterxml/jackson/annotation/JsonValue.html

"Marker annotation similar to XmlValue that indicates that results of the annotated "getter" method (which means signature must be that of getters; non-void return type, no args) is to be used as the single value to serialize for the instance. Usually value will be of a simple scalar type (String or Number), but it can be any serializable type (Collection, Map or Bean). "

e.g. if you have these classes
```
public class A {
    public B b = B();
}
public class B {
    @JsonValue
    public string getValue() {
        return "value";
    }
}
```
`A` would **not** marshal to `{"b": {"value":"value"}}` but just to `{"b":"value"}`.

This PR attempts to add support for `@JsonValue` to swagger-core 2.x. It would be difficult to add to 1.x without a certain amount of refactoring due to the difference between `Property` and `Model`.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2017 SmartBear Software
+ * (c) Copyright 2017 Hewlett Packard Enterprise Development LP
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package io.swagger.jackson;
 
 import com.fasterxml.jackson.annotation.JsonIdentityInfo;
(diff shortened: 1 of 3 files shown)


modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
```diff
@@ -202,8 +219,11 @@ public Schema resolveProperty(JavaType propType,
 
                     if("object".equals(mi.getType())) {
                         // create a reference for the property
-                        final BeanDescription beanDesc = _mapper.getSerializationConfig().introspect(propType);
-                        String name = _typeName(propType, beanDesc);
+                        String name = mi.getName();
+                        if (StringUtils.isEmpty(name)) {
```
liam-jackson-hpe: I believe it should be acceptable to trust the name if its been set in the `mi`.

However, I'm not 100% confident because I'm not aware of the intention behind the below re-introspection of `propType` to get `name`. Also the tests still pass if I completely remove this `if` and just rely on `mi.getName()`.

This feature does require using the name from `mi` and not the one from `_typeName(propType, beanDesc)` as it would be wrong for classes using `@JsonValue`.

modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
```diff
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2017 SmartBear Software
+ * (c) Copyright 2017 Hewlett Packard Enterprise Development LP
```
webron: This is going to be tricky.

liam-jackson-hpe: Hi @webron, please could you elaborate on the problem here?

webron: By donating the code, the ownership is not going to be HPE's.

liam-jackson-hpe: I've been advised that if I wrote a portion of the code, then that portion is owned by my employer.  They are happy to contribute the code to your project, but feel that an appropriate copyright notice needs to be present for two reasons:
      * Regardless of whether the copyright line is included, the code I wrote is really owned by them.
      * Not notifying other potential users of the complete copyright ownership can only lead to misunderstandings down the road.

frantuma:
closing as support introduced elsewhere

Issue #2328, opened 2017-07-20
Missing annotation "Encoding"


arthurdm:
From spec:  https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#encodingObject

Should be linked from the Content / MediaType annotation

frantuma:
merged in #2437

PR #2437, opened 2017-09-20
Issue #2328: added encoding annotation 

a-saf:
Encoding annotation only; jaxrs2 implementation will be in a separate PR.


a-saf:
This PR is for issue #2328

Issue #2466, opened 2017-09-29
List<String> in @BeanParam results in array of type "object" in generated JSON/YAML


ChristianCiach:
This bug has been introduces by commit 26f9572a6a92cd2375ef876baabe8667dbbf9dcd, which wrongfully replaced `.getGenericType()` by `.getRawType()`, even if "`.getType()`" would have been the correct non-deprecated replacement.

Pull-Request incoming.

ChristianCiach:
Pull requests for master and 2.0 are ready. Please merge at your earliest convenience.

frantuma:
implemented in #2548

PR #2467, opened 2017-09-29
#2466 retain generic types for attributes of BeanParam

ChristianCiach:
As described in #2466, this pull request will fix the wrong schema generation if BeanParams with generic attributes are used. The bug was originally introduced by commit 26f9572a6a92cd2375ef876baabe8667dbbf9dcd.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/DefaultParameterExtension.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/DefaultParameterExtension.java
@@ -125,7 +125,7 @@ public class DefaultParameterExtension extends AbstractSwaggerExtension {
      */
     private void handleAdditionalAnnotation(List<Parameter> parameters, Annotation annotation, 
         final Type type, Set<Type> typesToSkip) {
-        if (CLASS_BEAN_PARAM != null && CLASS_BEAN_PARAM.isAssignableFrom(annotation.getClass())) {
+        if (shouldHandleAdditionalAnnotation(annotation)) {
             // Use Jackson's logic for processing Beans
             final BeanDescription beanDesc = mapper.getSerializationConfig().introspect(constructType(type));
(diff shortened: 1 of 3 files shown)


ChristianCiach:
@webron 
Please consider this pull request for the next 1.5-release.

webron:
@pjfanning - inviting you to review this PR as well.

ChristianCiach:
An identical pull request for the 2.0 branch has already been merged.

pjfanning:
The change seems ok but I think there should be unit tests added to avoid a regression

pjfanning:
@ChristianCiach can you provide a test case or give some indication of how to reproduce the issue being fixed here?

ChristianCiach:
@pjfanning I will try to create a test case.

ChristianCiach:
I've committed a unit test. It's pretty ugly, because there is no `BeanParam` annotation in JAX-RS-API < 2.0, so I had to introduce my own annotation called "`MyBeanParam`". I tried to contain the ugliness in the class `BeanParamTest.java`. 

To apply this test to the 2.0 branch, `MyBeanParam` can be removed in favour of the real `BeanParam`.

pjfanning:
This looks good to merge - @ChristianCiach  could you add a new pull request for the 2.0 branch with your suggested test for that branch?

ChristianCiach:
*sigh* :D

ChristianCiach:
Pull request for 2.0: https://github.com/swagger-api/swagger-core/pull/2546

By the way: There are multiple instances where a class extends a standard collection like LinkedHashMap, like `Paths.java` and `Callback.java. This is considered bad practice, because there are many pitfalls. The mentioned classes override equals(), so there is the problem that `anyMap.equals(paths)` can be true, but `paths.equals(anyMap)` can not, which violates the contract of equals(). Also, I would never expect the classes Paths and Callback to be Maps.

frantuma:
included in #2548

PR #2521, opened 2017-11-13
Use dynamic port number for integration test

reftel:
Also makes the tests get skipped insted of failing when run from outside of maven.

### DIFF
--- a/modules/swagger-jaxrs/pom.xml
+++ b/modules/swagger-jaxrs/pom.xml
@@ -49,6 +49,26 @@
                 </configuration>
             </plugin>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


modules/swagger-jaxrs/pom.xml
```diff
@@ -49,6 +49,26 @@
                 </configuration>
             </plugin>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <version>1.6</version>
```
frantuma: it possibly makes sense using latest plugin version `3.0.0`

reftel: Fixed. Thanks!

webron:
Should this be migrated to the 2.0 branch too?

frantuma:
applied also to 2.0 in #2535

PR #2561, opened 2017-12-02
Ensure default empty string is set as Null on BooleanProperty

### LINKED ISSUE #2484: `ApiModelProperty` annotation generates automatically `example` for `Boolean`
I'm working in Scala.
I defined my case class as follow:
```Scala
import io.swagger.annotations.{ApiModel, ApiModelProperty}

@ApiModel(value = "TokenIntrospectionResponse")
case class TokenIntrospectionResponse(@(ApiModelProperty@field) @BeanProperty val active: Boolean)
```

When I read the generated schema, I see:
```JSON
"definitions": {
   "TokenIntrospectionResponse": {
      "type": "object",
      "properties": {
        "active": {
            "type": "boolean",
            "example": false
         }
      }
   }
}
```

If I replace the type of `active` property from `Boolean` to `String`, the result of the Swagger doc generation is:
```JSON
"definitions": {
   "TokenIntrospectionResponse": {
      "type": "object",
      "properties": {
         "active": {
            "type": "string"
         }
      }
   }
}
```

If I remove the `ApiModelProperty` annotation, the `example` is not generated for type `Boolean`:
```JSON
"definitions": {
   "TokenIntrospectionResponse": {
      "type": "object",
      "properties": {
         "active": {
            "type": "boolean"
         }
      }
   }
}
```

- Is it a bug ?
- Is there a way to keep this `ApiModelProperty` annotation and delete this `example` property for boolean property ?
- If not, is there an existing workarround ?

For your information, I need this annotation to specify this property is required.
I need to erase this `example` property because Amazon API Gateway is not supporting this one ([source](http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html)):
> - API Gateway supports most of the Swagger specification, with the following exceptions:
>    - [...]
>   - **The __example__ tag is not supported.**

Thanks in advance.

simonamc:
fix #2484

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/properties/BooleanProperty.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/properties/BooleanProperty.java
@@ -41,7 +41,9 @@ public class BooleanProperty extends AbstractProperty implements Property {
     @Override
     public void setExample(Object example) {
         if (example instanceof String) {
-            this.example = Boolean.parseBoolean((String)example);
+            if (!((String) example).isEmpty()) {
+                this.example = Boolean.parseBoolean((String) example);
+            }
         } else {
             this.example = example;
         }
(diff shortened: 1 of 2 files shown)


frantuma:
Thanks!

Issue #1857, opened 2016-07-19
@ApiModelProperty required=false is overriden by @NotNull annotation


m-radzikowski:
Please consider this example:

``` java
@ApiModelProperty(required = false, value = "Required only for Ipsum group")
@NotNull(groups = Ipsum.class)
private int lorem;
```

In this case we have `@NotNull` annotation, but with specified validation group. Because of that in swagger this property shouldn't be marked as required so we add `required=false` in `@ApiModelProperty`. Unfortunately `@NotNull` annotation seems to have priority over Swagger own `@ApiModelProperty` and `lorem` is still marked as required.

keir-rex:
This has cost me a major headache too.

patrickstubner:
I could use this weekness [https://github.com/swagger-api/swagger-core/issues/1705#issuecomment-220629231](url) to achieve a workaround, that works for me.
Create an own NotNull (PartialNotNull) annotation.
Your custom annotation now can use the NotNull annotation inside to get the validation right.
This way Swagger don't recognise your custom annotation to be a NotNull validation annotation and generats the property as not required.

Use it like this
`
@PartialNotNull(groups = Ipsum.class)
private int lorem;
`

This is the custom annotation:
`
@NotNull
@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER })
@Retention(RUNTIME)
@Documented
@Constraint(validatedBy = { })
public @interface PartialNotNull {
  String message() default "{javax.validation.constraints.NotNull.message}";
  Class<?>[] groups() default { };
  Class<? extends Payload>[] payload() default { };
  @Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER })
  @Retention(RUNTIME)
  @Documented
  @interface List {
    PartialNotNull[] value();
  }
}
`

fehguy:
It's kind of hard to support every case with this.  To most people, `@NotNull` means "cannot be null" and "value is required".  Thus that annotation is winning.  There's no way to handle conflicting configurations without making someone unhappy.

I can see a case where the `@ApiModelProperty` will _win_ no matter what.  Maybe that's the answer but still someone won't like it.

m-radzikowski:
Well, it's weird that external (`@NotNull`) annotation has priority over library's own annotation (`@ApiModelProperty`) when it comes to displaying library-specific stuff...

fehguy:
[...]

PR #2606, opened 2018-01-04
#1857 Use @ApiModelProperty required value if present instead of @NotNull

m-radzikowski:
As stated in #1857, in this situation:
```java
@ApiModelProperty(required = false, value = "Required only for Ipsum group")
@NotNull(groups = Ipsum.class)
private int lorem;
```
field `lorem` is marked as required, as it has `@javax.validation.constraints.NotNull` annotation. This change modifies behavior so if `@ApiModelProperty` is present, it will always be used it's `required` value. Swagger annotations should have precedence over other annotations.

Only drawback is that in existing code this:
```java
@ApiModelProperty(value = "Smth")
@NotNull
private int lorem;
```
currently is marked as required (because of `@NotNull`), and after change it will be marked as not-required (as `required=false` is default value in `@ApiModelProperty`).

Every other code will act as before, so this:
```java
@NotNull
private int lorem;
```
will be marked as required, since there is no Swagger annotation that would override it.

This change can be done for 2.0 version as well, so if it will be accepted I can make another PR for 2.0. Or maybe this change should go only to 2.0, as it somehow changes previous behavior. Please let me know what do you think.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
@@ -472,6 +472,11 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
                 }
 
                 if (property != null) {
+                    // apply properties from bean validation annotations first,
+                    // so they could be overriden by Swagger annotation values
+                    // (see https://github.com/swagger-api/swagger-core/issues/1857)
+                    applyBeanValidatorAnnotations(property, annotations);
+
                     property.setName(propName);
 
                     if (mp != null && !mp.access().isEmpty()) {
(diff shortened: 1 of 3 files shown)


frantuma:
Thanks with your effort with this one! I don't think we can go this way actually, because of the issue you mention in your comment about `required` defaulting to `false` which wouldn't allow to use the annotation "freely". A solution similar to what done for `readOnly` in #2675 for 1.5 and #2695 for 2.0 should be applicable here as well, adding an annotation field named e.g. `necessity` (or better naming..) with related enum with values like `AUTO`, `REQUIRED`, `NOT_REQUIRED`, with related support in reader/resolver code, and deprecating `required`. 

Please let me know if you can help, and/or any comment.

MariusKl:
I face the same issue when working around the lack of swagger supporting regex with (?i)
 
`import javax.validation.constraints.Pattern;

@Pattern(regexp = "^(?i)(new|closed)$")
@Schema(description ="In fact the field allows the pattern ^(?i)(new|closed)$ but this causes an invalid swagger document", pattern = "^(new|closed|NEW|CLOSED)$")`

Would be great to favor the own annotations over external ones.

Issue #2621, opened 2018-01-15
Deserializer doesn't support allOf in properties


Tiller:
The current implementation is not able to handle object properties that are defined using "allOf"
"allOf" is supported at the root of definitions through the "ComposedModel", but not much anywhere else.

The following example cannot be deserialized properly:

```
definitions:
  standaloneDef:
    type: object
    description: Standalone
    properties:
      name:
        type: string
        description: name
        
  defWithDep:
    type: object
    description: With dep
    required:
    - id
    properties:
      id:
        type: string
        description: id
        example: "0x00"
      def:
        type: object
        allOf:
        - $ref: "#/definitions/standaloneDef"
        - type: object
          properties:
            otherProp:
              type: string
```

The 'def' attribute in 'defWithDep' ends up with no property at all after deserialization. It is read as the following:
```
{
  "swagger" : "2.0",
  "info" : {
    "version" : "0.0.1",
    "title" : "..."
  },
  "basePath" : "...",
  "paths" : {},
  "definitions" : {
    "standaloneDef" : {
      "type" : "object",
      "properties" : {
        "name" : {
          "type" : "string",
          "description" : "name"
        }
      },
      "description" : "Standalone"
    },
    "defWithDep" : {
      "type" : "object",
      "required" : [ "id" ],
      "properties" : {
        "id" : {
          "type" : "string",
          "example" : "0x00",
          "description" : "id"
        },
        "def" : {
          "type" : "object",
          "properties" : { }
        }
      },
      "description" : "With dep"
    }
  },
  "parameters" : { },
  "responses" : { }
}
```

Regarding how I ended up in this scenario is because of a long time running conversation here: vendor extensions next to references. I need to do the following:
```
defWithDep:
  type: object
  properties:
    def:
      $ref: "#/definitions/standaloneDef"
      x-thingy: thing
```

but the (one of the) only way to do it, is by using allOf:
```
defWithDep:
  type: object
  properties:
    def:
      x-thingy: thing
      allOf:
      - $ref: "#/definitions/standaloneDef"
```

which is - afaik - compliant with the spec (swagger editor manages it well)

frantuma:
implemented in #2890

PR #2622, opened 2018-01-15
#2621 Basic implementation of 'allOf' for properties.

Tiller:
Linked to #2621

- Creation of a new property implementation: ComposedProperty that contains a field "allOf" listing all the included properties
- Creation of the associated UT
- Update of the PropertyDeserializer to handle this new implementation

Note: this PR adds basic support of the functionnality. It probably doesn't handle all the cases of the spec. But I'd say minimal support is better than no support.


ymohdriz:
Hi @frantuma,

On which swagger-core lib version, this change will be available as it's not available in 1.5.20 version.

Thanks,
Mohammed

frantuma:
@ymohdriz Thanks for the PR! it's included in #2890 with a small change

PR #2668, opened 2018-02-19
Take into account parentResponses and components.getResponses()

ilgrosso:
At the moment it seems that `parentResponses` and `components.getResponses()` are barely ignored.

### DIFF
--- a/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
+++ b/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
@@ -720,6 +720,19 @@ public class Reader implements OpenApiReader {
                 parentResponses);
     }
 
+    private void addApiResponses(Operation operation, Map<String, ApiResponse> responses) {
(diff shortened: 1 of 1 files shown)


ilgrosso:
> Responses in components are relevant only if referenced in an operation response, and are not meant to be added to operation responses.

Ah ok, sorry I did not get this; so, what if I want to add some default responses to *all* operations? With Swagger I could annotate a [super-interface](https://github.com/apache/syncope/blob/2_0_X/common/rest-api/src/main/java/org/apache/syncope/common/rest/api/service/JAXRSService.java#L27-L37) to obtain such a result, I thought `components.getResponses()` was the OpenAPI equivalent.

> Also for parentResponses, actually a cleanup would remove that argument, as subresource locators method consider only responses defined on the subresource

Understand.

frantuma:
@ilgrosso Just merged #2685 implements the behaviour you mention also for 2.0, allowing `ApiResponse` annotations at class level and adding them to each operation of the class

ilgrosso:
Thanks @frantuma , indeed.

Issue #2669, opened 2018-02-19
Add Xml property to ArrayModel class


rbsilva:
| Q                                 | A
| --------------------------------- | -------
| Bug or feature request?           |  feature
| Which Swagger-Core version?       | 1.5.19
| Which Java version?               | Java 8
| Which JAX-RS framework & version? | -

Actual behaviour:

When creating a ArrayModel:
```
new ArrayModel().items(....);
```
The output json is:
```
{
   "type":"array",
   "items":{"$ref":"#/definitions/Payment"}
}
```

Expected behaviour:

When creating a ArrayModel with Xml:
```
new ArrayModel().items(....).xml(new Xml().wrapped(true).name("payments"));
```
The output json is:
```
{
   "type":"array",
   "items":{"$ref":"#/definitions/Payment"},
   "xml":{"name":"payments","wrapped":true}
}
```

frantuma:
implemented in #2684

rbsilva:
Thanks :)

PR #2670, opened 2018-02-19
Add Xml property to ArrayModel class #2669

rbsilva:
In reference to issue: #2669


frantuma:
Thanks! slightly refactored in merged #2670

PR #2674, opened 2018-02-20
Inner properties may have XML declared (rebased)

### LINKED ISSUE #2316: items.xml not deserialized
The **xml** declaration on the "items" property is ignored by `PropertyDeserializer`:

```yml
     progdoi:
        type:                       array
        xml:
          wrapped:                  true
        items:
          type:                     string
          xml:
            name:                   doi
```

According to the docs, the "items.xml" declaration is valid:

> When defined within the Items Object (items), it will affect the name of the individual XML elements within the list. When defined alongside type being array (outside the items), it will affect the wrapping element and only if wrapped is true. If wrapped is false, it will be ignored.

Running Swagger with `debugSwagger=true` and `debugParser=true` shows that the JSON includes the "items.xml" element, but the "xml" declaration does not make it to Swagger's model.

zregvart:
This is a rebase of pull request #2313 by @mobreza

Fixes #2316


mobreza:
@zregvart 🤞

frantuma:
included and replaced by #2688

Issue #2492, opened 2017-10-24
Fields not specific to body parameter should be in AbstractSerializableParameter


ymohdriz:
Hi team,

In general, fields which are not specific to body parameter should be in AbstractSerializableParameter and fields applicable to all parameters in AbstractParameter.

But field 'allowEmptyValue' which is not specific to body parameter should be in AbstractSerializableParameter. 

Is there any reason that it was added in AbstractParameter instead of AbstractSerializableParameter?

Thanks,
Mohammed

frantuma:
implemented in #2686

PR #2686, opened 2018-02-24
refs #2492 - Relocate Parameter allowEmptyValue

ralphdoe:
Property allowEmptyValue reubicated. 
Solving this Issue: https://github.com/swagger-api/swagger-core/issues/2492

### DIFF
--- a/modules/swagger-jaxrs/src/test/java/io/swagger/SimpleReaderTest.java
+++ b/modules/swagger-jaxrs/src/test/java/io/swagger/SimpleReaderTest.java
@@ -682,7 +682,7 @@ public class SimpleReaderTest {
         Parameter readOnlyParam = swagger.getPath("/fun/readOnly").getGet().getParameters().get(0);
         assertTrue(readOnlyParam.isReadOnly());
 
-        Parameter allowEmptyParam = swagger.getPath("/fun/allowEmpty").getGet().getParameters().get(0);
+        SerializableParameter allowEmptyParam = (SerializableParameter) swagger.getPath("/fun/allowEmpty").getGet().getParameters().get(0);
         assertTrue(allowEmptyParam.getAllowEmptyValue());
     }
 }
--- a/modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
@@ -37,6 +37,7 @@ public abstract class AbstractSerializableParameter<T extends AbstractSerializab
     protected String example;
     private Integer maxItems;
     private Integer minItems;
+    protected Boolean allowEmptyValue;
 
     @JsonIgnore
     protected List<String> _enum;
@@ -385,6 +386,16 @@ public abstract class AbstractSerializableParameter<T extends AbstractSerializab
         this.minItems = minItems;
     }
 
+    @Override
+    public Boolean getAllowEmptyValue() {
+        return allowEmptyValue;
+    }
+
+    @Override
+    public void setAllowEmptyValue(Boolean allowEmptyValue) {
+        this.allowEmptyValue = allowEmptyValue;
+    }
+
     @JsonProperty("x-example")
     public Object getExample() {
         if (example == null) {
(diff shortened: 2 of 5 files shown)


modules/swagger-models/src/main/java/io/swagger/models/parameters/Parameter.java
```diff
@@ -37,7 +37,6 @@
 
     void setReadOnly(Boolean readOnly);
 
-    Boolean getAllowEmptyValue();
 
-    void setAllowEmptyValue(Boolean allowEmptyValue);
+
```
frantuma: these 2 methods should also be moved to SerializableParameter interface

modules/swagger-jaxrs/src/test/java/io/swagger/SimpleReaderTest.java
```diff
@@ -682,7 +682,7 @@ public void scanClassWithReadOnlyAndEmptyValueParams() {
         Parameter readOnlyParam = swagger.getPath("/fun/readOnly").getGet().getParameters().get(0);
         assertTrue(readOnlyParam.isReadOnly());
 
-        Parameter allowEmptyParam = swagger.getPath("/fun/allowEmpty").getGet().getParameters().get(0);
+        QueryParameter allowEmptyParam = (QueryParameter) swagger.getPath("/fun/allowEmpty").getGet().getParameters().get(0);
```
frantuma: you would use here `SerializableParameter` instead of `QueryParameter`

modules/swagger-models/src/main/java/io/swagger/models/parameters/Parameter.java
ralphdoe: SerializableParameter extends Parameter Interface, so there's no need to remove it here. It was being inherited from the Parameter Interface.

modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java
```diff
@@ -530,8 +539,14 @@ public boolean equals(Object obj) {
...
         } else if (!type.equals(other.type)) return false;
         if (uniqueItems == null) {
-            if (other.uniqueItems != null) return false;
-        } else if (!uniqueItems.equals(other.uniqueItems)) return false;
+            if (other.uniqueItems != null) {
+                return false;
+            }
+        } else if (!uniqueItems.equals(other.uniqueItems)) {
+            return false;
+        } else if (allowEmptyValue != null ? !allowEmptyValue.equals(other.allowEmptyValue) : other.allowEmptyValue != null) {
```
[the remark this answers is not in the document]
ralphdoe: Deleted ```else```. The condition of allowEmptyValue must not be dependent on the uniqueItems one.

modules/swagger-jaxrs/src/test/java/io/swagger/SimpleReaderTest.java
ralphdoe: This attribute must be valid only to QueryParameter and FormParameter as the documentation says: ```This is valid only for either query or formData parameters and allows you to send a parameter with a name only or an empty value```.

frantuma: We are moving the property from the more generic interface/abstract class `Parameter/AbstractParameter` which is also implemented by `BodyParameter`, to the less generic `SerializableParameter/AbstractSerializableParameter` which is not implemented by BodyParameter; this is implemented also by other parameters (not query or formData) but it's a good compromise of effort/result for this scope.

modules/swagger-models/src/main/java/io/swagger/models/parameters/Parameter.java
frantuma: Don't understand the comment, however the above comment applies here also:
We are moving the property from the more generic interface/abstract class `Parameter/AbstractParameter` which is also implemented by `BodyParameter`, to the less generic `SerializableParameter/AbstractSerializableParameter` which is not implemented by BodyParameter; this is implemented also by other parameters (not query or formData) but it's a good compromise of effort/result for this scope.

modules/swagger-jaxrs/src/test/java/io/swagger/SimpleReaderTest.java
ralphdoe: Now I've understood the change. Thanks! Changed to SerializableParameter.

modules/swagger-models/src/main/java/io/swagger/models/parameters/Parameter.java
ralphdoe: Methods moved.

PR #2696, opened 2018-03-04
Fixing Guava version OSGi imports as reflections-0.9.11 is using 20.0

reta:
Hey guys,

Run into this unexpected issue recently while integrating `Swagger 2.0` into Karaf OSGi container (`4.1.x` releases) for JAX-RS services. So Swagger `2.0` has a dependency on Guava `23.0` while the underlying `reflections-0.9.11` library is declaring dependency on Guava `20.0`. Obviously, the `23.0` is being picked up at the end, which messes up the  OSGi import constraints. One of the possible solutions (this PR) is to relax the requirement for Guava versions, `[20.0, 24)` instead of `[23.0, 24)`. With that, and the fix for `reflections-0.9.11` (https://github.com/apache/servicemix-bundles/pull/111), things are working properly. 

Best Regards,
    Andriy Redko

### DIFF
--- a/modules/swagger-jaxrs2/pom.xml
+++ b/modules/swagger-jaxrs2/pom.xml
@@ -38,6 +38,7 @@
                         </Export-Package>
                         <Import-Package>
                             javax.ws.rs*;version="2.0",
+                            com.google.common.base;version="[20.0,24)",
                             *
                         </Import-Package>
                     </instructions>


frantuma:
Hi, actually guava dep has been removed in #2618 for swagger 2.0, so this PR shoudn't be necessary

reta:
Hi @frantuma ,

Thanks for the reply, I see it in the `swagger-core`'s, `pom.xml`
```
<guava-version>23.0</guava-version>
```
It may not be used by Swagger directly but the Guava deps come from `reflections` library as well. However, due to the fact that Swagger asks for higher version, this is what is generated in the OSGi manifest for `swagger-jaxrs2` module:

```
Import-Package: ... com.google.common.base;version="[23.0,24)", ...
```
Alternatively, it would be simpler to downgrade the Guava in  `swagger-core`'s, `pom.xml` to `20.0`, the problem will be gone as well.

```
<guava-version>20.0</guava-version>
```

What do you think? Thanks!

Best Regards,
    Andriy Redko

frantuma:
Hi @reta, what about fix in #2698? it removes guava from root deps management, and should be equivalent as result to downgrading to 20.0. Can you check if it works ok in your scenario?

reta:
Hi @frantuma,

Yes, certainly, that would solve it. Confirmed by the manifest:

```
Import-Package: ..., com.google.common.base;version="[20.0,21)",
```

Thank you.

Best Regards,
    Andriy Redko

reta:
Closing in favor of https://github.com/swagger-api/swagger-core/pull/2698

PR #2698, opened 2018-03-05
refs #2696 - remove guava from deps management

frantuma:


reta:
This will also put back the proper transitive Guava dependency to `jaxrs2` module, thank you @frantuma

Issue #2707, opened 2018-03-15
Enum Names Not Overriden with @JsonProperty


PeaceInColour:
| Q                                 | A
| --------------------------------- | -------
| Bug or feature request?           | Bug
| Which Swagger-Core version?       |1.5.18
| Which Java version?               |1.8u151
| Which JAX-RS framework & version? |RESTEasy 3.0.19 with JAX-RS 2.0
 

Using `@JsonProperty` to override enum names isn't working. The following example below:

```java
public enum SomeEnum {
	@JsonProperty("num-one")
	NUMONE,
	
	@JsonProperty("num-two")
	NUMTWO
}
```
results in an OpenAPI file that looks like:

```json
"enum": [
	"NUMONE",
	"NUMTWO"
]
```

instead of the expected:

```json
"enum": [
	"num-one",
	"num-two"
]
```

The issue is in ModelResolver which is using the deprecated Jackson Databind method `findEnumValue` to read in the enum name. This method does not look for any overrides and just defaults to the name in the code. Using the replacement method `findEnumValues` fixes this problem.

PeaceInColour:
Once I get permission to push (or figure out what's wrong if the issue is on my end) I can push a fix for this.

frantuma:
Thanks for the fix! to open a PR, please fork the repo and push the branch to your own repo, and open the PR from there targeting the main repo.

frantuma:
fixed in #2796

rabiori:
any workarounds for someone stuck in the old version

jlous:
@rabiori Ever find a workaround?

DeepakPatankar:
I am also stuck in the old version, can anyone suggest a workaround?

jlous:
We used a custom ModelConverter on 2.1.0, which has a regression on this issue (fixed in 2.1.3)
Don't know if that option applies to 1.5

PR #2712, opened 2018-03-19
#2707 Overriding Enum Names with @JsonProperty

PeaceInColour:
Issue is [here](https://github.com/swagger-api/swagger-core/issues/2707)

Enum Names are currently not changed when `@JsonProperty` is used. This PR fixes this bug and allows for `@JsonProperty` to be used for renaming enums.


PeaceInColour:
Would I be able to get someone to review this PR?

PeaceInColour:
Would be really nice to get this fix in (or some feedback at least) since this issue is important for me. Otherwise, I'd have to keep using snapshots where I've manually added this change to the code.

frantuma:
Thanks for the fix! included in merged #2796

Issue #2705, opened 2018-03-14
200 Responses Not Overwritten at the Class Level


PeaceInColour:
| Q                                 | A
| --------------------------------- | -------
| Bug or feature request?           | Bug
| Which Swagger-Core version?       |1.5.18
| Which Java version?               |1.8u151
| Which JAX-RS framework & version? |RESTEasy 3.0.19 with JAX-RS 2.0
 

I have a fix for this issue. I'm just creating this issue for the PR.

Any default 200 (or whatever the default responses are set to) responses cannot be overridden with `@ApiResponse` at the class level. This isn't an issue for method level `@ApiResponse` annotations. 

The bug occurs in Reader.java at lines 945 - 955:

```java
for (ApiResponse apiResponse : apiResponses) {
    addResponse(operation, apiResponse, jsonViewAnnotation);
}
    // merge class level @ApiResponse
    for (ApiResponse apiResponse : classApiResponses) {
        String key = (apiResponse.code() == 0) ? "default" : String.valueOf(apiResponse.code());
        if (operation.getResponses() != null && operation.getResponses().containsKey(key)) {
            continue;
        }
        addResponse(operation, apiResponse, jsonViewAnnotation);
}
``` 
Here, method level `@ApiResponse` annotations are added first, then class level `@ApiResponse` annotations are added. This works in every other scenario since the class level for loop checks to see if the response code has already been included before adding the class level response to stop the class level responses from overriding the method level responses. These for loops do not take into account that the default responses are added to operation earlier in the code which results in the bug.

To fix this bug I changed the order the responses are being added to operation to be: default->class level->method level. This way each time a response is found the response will always override any response with the same response code.

PeaceInColour:
I'm getting a permission denied message when trying to push my branch. Do I need to get permission to push my branch or is there something wrong on my end?

frantuma:
Thanks for the fix! to open a PR, please fork the repo and push the branch to your own repo, and open the PR from there targeting the main repo.

PeaceInColour:
For PRs do we have to assign someone to them ourselves?

webron:
@PeaceInColour - no, we monitor and review the PRs as we can.

frantuma:
see also [this comment](https://github.com/swagger-api/swagger-core/pull/2713#issuecomment-386583554) in PR #2713

MichakrawSB:
Closing this issue as part of our https://github.com/swagger-api/swagger-core/issues/5074.
[...]

PR #2713, opened 2018-03-19
#2705 Oerwritting Default Responses at Class Level

PeaceInColour:
Issue is [here](https://github.com/swagger-api/swagger-core/issues/2705)

Class level `@ApiResponse` annotations are not able to overwrite default responses (eg. 200). This PR allows for default responses to be overwritten at the class level while simplifying the order responses are added to the Operation object.

### DIFF
--- a/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
+++ b/modules/swagger-jaxrs/src/main/java/io/swagger/jaxrs/Reader.java
@@ -942,18 +942,16 @@ public class Reader {
             }
         }
 
+        //class level @ApiResponse
+        for (ApiResponse apiResponse : classApiResponses) {
+            addResponse(operation, apiResponse, jsonViewAnnotation);
+        }
+        
+        //method level @ApiResponse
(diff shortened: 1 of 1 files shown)


frantuma:
@PeaceInColour Thanks for your effort with this one. As you mentioned currently responses are overwritten and not merged; this makes sense for method level annotations (as you can specify return type in annotation `response` field) while for class level it is not applicable (as different methods would return different types), on top of the original issue of #2705.

A possible direction, also to maintain backward compatibility, would be to introduce a new `ApiResponse` annotation boolean field `applyReturnType` or similar naming, along with the change of order as done in your commit;  with `applyReturnType = true`, the code would merge the returnType (if present) into the response from annotation, be it at class or method level. 

Let me know if you can help with that (and related tests..), and/or any comment you might have

PeaceInColour:
Thanks @frantuma! This is definitely on my list of things to fix right now. I was hoping to avoid requiring the use of an annotation for this but I don't think it's a big deal if I have to. I'll give that a shot and see what happens!

MichakrawSB:

PR #2782, opened 2018-04-25
"Operation.isDeprecated()" should return false as default

ackintosh:
As described below, the default value of 'deprecated' is `false`.
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#user-content-operationDeprecated

So I think that `Operation.isDeprecated()` should behave same as the description: it should return true or false, never returns null.

related issue in swagger-parser: https://github.com/swagger-api/swagger-parser/issues/697

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/Operation.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/Operation.java
@@ -259,10 +259,10 @@ public class Operation {
 
     public void setDeprecated(Boolean value) {
         if (value == null) {
-            this.deprecated = null;
-        } else {
-            this.deprecated = value;
+            return;
         }
+
+        this.deprecated = value;
     }
 
     @JsonAnyGetter
--- a/modules/swagger-models/src/main/java/io/swagger/models/Operation.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/Operation.java
@@ -258,7 +258,7 @@ public class Operation {
     }
 
     public void setDeprecated(Boolean value) {
-        if (value == null || value.equals(Boolean.FALSE)) {
+        if (value == null) {
             this.deprecated = null;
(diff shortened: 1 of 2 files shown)


ackintosh:
https://travis-ci.org/swagger-api/swagger-core/builds/370936121?utm_source=github_status&utm_medium=notification

The currently failing tests is asserting 'whole of the spec' with the 'whole of the Swagger object'.

[ParameterSerializationTest.java#L392](https://github.com/ackintosh/swagger-core/blob/10736b2ba93ed97d79732ac4964b44ab5ea59a09/modules/swagger-core/src/test/java/io/swagger/parameter/ParameterSerializationTest.java#L392)

```
[ERROR] testIssue1765(io.swagger.parameter.ParameterSerializationTest)  Time elapsed: 0.01 s  <<< FAILURE!
java.lang.AssertionError: 
Serialized object:
{"swagger":"2.0","paths":{"/test":{"get":{"parameters":[{"name":"days","in":"path","required":true,"type":"integer","format":"int32","enum":[1,2,3,4,5]}],"responses":{"default":{"description":"great"}},"deprecated":false}}}}
does not equal to expected serialized string:
{"swagger":"2.0","paths":{"/test":{"get":{"parameters":[{"name":"days","in":"path","required":true,"type":"integer","format":"int32","enum":[1,2,3,4,5]}],"responses":{"default":{"description":"great"}}}}}}
```

By this PR, Swagger object is deserialized `deprecated: false` as default, and caused the failures.

frantuma:
Thanks for spotting this, just merged #2795 addresses the issue by allowing to set deprecated as `false` in `Operation.setDeprecated()`, therefore whenever explicitly set as `false` (e.g. while deserializing, or by setting programmatically), `getDeprecated()` will return correctly `false`

Issue #2809, opened 2018-05-17
Swagger core 1.5 does not allow spec filtering of definitions


shauera:
We have a use case in which we want to produce different swagger docs derived from the same API. Implementing the `AbstractSpecFilter` is what we need. However, it lacks the hook for filtering specific definitions. Using ` isRemovingUnreferencedDefinitions` is too broad for our use case.

We suggest add an ` isDefinitionAllowed` method to address this use case.

frantuma:
implemented in #2820

PR #2810, opened 2018-05-17
#2809 Allow filtering of specific definitions in SpecFilter.

shauera:
This addresses the use cases where filtering out specific definitions is required.

#2809


shauera:
What are the chances of getting this PR into 1.5.20?

joyfulelement:
+1 Would be great if this enhancement gets in.

frantuma:
@shauera we can sneak this in, but we would need a test case added to SpecFilterTest, can you possibly take care of that, and ping me here when done?

frantuma:
included and replaced by #2820

shauera:
Thanks for merging this change. I was about to add a unit test but I see you have added one. Would you still want another one?

frantuma:
nope not needed thx

Issue #2638, opened 2018-01-30
$ref property is missing in callback.java (Open API 3.0)


ymohdriz:
Hi team,

Unable to refer a callback which is available in the component because $ref property is missing in the callback.java.

All the other models in the components like schema, ApiResponse, Link etc contains $ref.

Expected Behavior:
$ref should be added

Thanks,
Mohammed

ymohdriz:
Hi team,

Raised a PR for this issue - https://github.com/swagger-api/swagger-core/pull/2834

Thanks,
Mohammed

frantuma:
implemented in #2882

PR #2834, opened 2018-06-08
From bugfix:ymohdriz

ymohdriz:
Hi team,

Unable to refer a callback which is available in the component because $ref property is missing in the callback.java. 
This PR is to add $ref property in the Callback.java.

Find the detailed info here https://github.com/swagger-api/swagger-core/issues/2638

Please review the PR and merge the same.

Thanks,
Mohammed


ymohdriz:
Hi team,

Can you please ignore one commit (Add summary property in ApiResponse ) out of 3 commits and consider the remaining 2 commits for merge.

Thanks,
Mohammed

PR #2843, opened 2018-06-15
Validating relative/internal Refs 

gracekarina:
References:
- https://github.com/swagger-api/swagger-parser/issues/727
- https://github.com/swagger-api/swagger-parser/issues/713

### DIFF
--- a/modules/swagger-models/pom.xml
+++ b/modules/swagger-models/pom.xml
@@ -105,6 +105,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.5</version>
+        </dependency>
+        <dependency>
             <groupId>asm</groupId>
             <artifactId>asm</artifactId>
             <scope>test</scope>
--- a/modules/swagger-models/src/main/java/io/swagger/models/refs/GenericRef.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/refs/GenericRef.java
@@ -1,5 +1,7 @@
 package io.swagger.models.refs;
 
+import org.apache.commons.lang3.StringUtils;
+
 /**
  * A class the encapsulates logic that is common to RefModel, RefParameter, and RefProperty.
  */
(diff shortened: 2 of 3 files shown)


modules/swagger-models/pom.xml
```diff
@@ -104,6 +104,11 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.5</version>
+        </dependency>
```
frantuma: We try not to add deps if not strictly necessary, use instead indexOf / charAt directly in code

modules/swagger-models/src/main/java/io/swagger/models/refs/GenericRef.java
```diff
@@ -1,5 +1,7 @@
 package io.swagger.models.refs;
 
+import org.apache.commons.lang3.StringUtils;
+
```
frantuma: We try not to add deps if not strictly necessary, use instead indexOf / charAt directly in code

modules/swagger-models/src/main/java/io/swagger/models/refs/GenericRef.java
```diff
@@ -91,12 +93,18 @@ private static String computeSimpleRef(String ref, RefFormat format, RefType typ
...
         if (ref.startsWith("http:") || ref.startsWith("https:")) {
             result = RefFormat.URL;
         } else if (ref.startsWith("#/")) {
             result = RefFormat.INTERNAL;
         } else if (ref.startsWith(".") || ref.startsWith("/")) {
             result = RefFormat.RELATIVE;
+        }else if (!ref.contains(":") &&   // No scheme
+                !ref.startsWith("#") && // Path is not empty
+                !ref.startsWith("/")&& // Path is not absolute
+                StringUtils.countMatches(ref, ".") == 1) { // Path does not start with dot but contains "." (file extension)
```
frantuma: We try not to add deps if not strictly necessary, use instead indexOf / charAt directly in code, probably `indexOf(".") != -1` is sufficient

Issue #2833, opened 2018-06-06
Property requestBody should be of type String instead of RequestBody in Links.java (OAS 3.0)


ymohdriz:
Hi team,

In Links.java, property requestBody should be of type String instead of RequestBody as it's clearly mentioned as below in https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#linkObject

requestBody | Any \| {expression} | A literal value or {expression} to use as a request body when calling the target operation.
-- | -- | --

Note: We used Map<String, **String**> for parameters field in Links.java

Thanks,
Mohammed

frantuma:
fixed in #2877

PR #2866, opened 2018-07-02
Fix for 2833 (Property requestBody should be of type String instead of RequestBody in Links.java (OAS 3.0))

ymohdriz:
Hi team,

This PR is to solve the issue #2833 
Please review the PR and merge the same.

Thanks,
Mohammed


frantuma:
Thanks for spotting this! just merged #2877 which updates the datatype to `Object` (handling also the "Any" case) and adds a test, and replaces this PR.

PR #2871, opened 2018-07-09
Implementation of Some Customization for Swagger Core (Issue 2994)

duckladydinh:
For the issue: https://github.com/swagger-api/swagger-core/issues/2994
Dear, I want to propose some customization options for OpenAPI like:
+ OpenAPIBuilderOptions.USE_FULLNAME: Boolean -> Including package name for models (in case you do not need to generate code and only want document)
+ OpenAPIBuilderOptions.OMIT_GENERIC: Boolean -> Omit the generic part of generic type. For example, consider class Query<T>, it will exclude the <T> part (Object).
+ OpenAPIBuilderOptions.USE_ENUMNAME: Boolean -> Force use of enum.name() instead of enum.toString().
+ OpenAPIBuilderOptions.RECYCLE_ENUM: Boolean -> Recycle enums (you do not need to inline enum all the time, just declare 1 as a normal enum class and use reference to it) 
+ OpenAPIBuilderOptions.HIDE_PARENTS: Boolean -> Adding ability to hide all properties from parent classes that are provided in Schema(allOf = {...})
+ OpenAPIBuilderToolbox.addModel(OpenAPI, Type) -> Add a method to directly add additional models and their dependencies to OpenAPI object.
+ Expose JsonObjectMapper in ObjectMapperFactory.defaultMapper to control visibility.


dtaveras:
This is great work. Any timeline on getting it merge to master?

duckladydinh:
> This is great work. Any timeline on getting it merged to master?

Thank you. I am the author of this. I have tried to contact the team long ago but it seems that they have no interest :(. So I guess it will... never be merged.

PR #2881, opened 2018-07-19
Added ability to apply Reflections filtering by package.

boatmisser:
When using a "fat" application (e.g. Dropwizard) Reflections parses the entire JAR not just the resource package requested, this is slow and produces large numbers of keys and values.
This PR adds the ability to configure Reflections package filtering so that we can ensure that only the packages  we require are actually parsed.

**Example Result**
Before:
Reflections took 1203 ms to scan 1 urls, producing 15550 keys and 29087 values
After:
Reflections took 66 ms to scan 1 urls, producing 18 keys and 19 values


frantuma:
closing as switched to Classgraph

PR #2883, opened 2018-07-19
new attribute "allowableValuesEnumClass" to manage allowable values from Enum Class

dedece35:
**Abstract**
To make it easiest the Enum values managment, I created a new attribute named "allowableValuesEnumClass" on following annotations : "ApiParam" and "ApiImplicitParam".
This new attribute was made on the same way as "allowableValues" attribute.

**Context**
My problem is that I have several Java Enums and I have to create a string field in each Enum with all values of this Enum. Then this string field is given to "allowableValues" attribute of ApiParam annotation.
I have to do this because "allowableValues" attribute can only take a string value and not a function like "MyEnum.values()" because Java annotations want a static string value.

**Solution**
A new attribute of ApiParam (and ApiImplicitParam) born, named "allowableValuesEnumClass".
Why this name ? because user has to give the Enum class (with full package name like "com.mypackage.MyEnum") containing values to get.
If this new "allowableValuesEnumClass" is valued, then the system will generate a supplier class for the enum class provided.
This supplier class source code will be generated in a Temp directory, then put in memory, then compiled, then loaded, then instanciated and at the end the only one existing method ("getValues") will be invoked to give the list of string values of the enum provided.
Of course, I documented my code and added LOGGER (in debug) to follow differents steps.

**Tests**
Units Tests : Of course, each new class created and modified was tested with a new Test Class.
Real word test : I made several manual tests with my company Rest API and values of my Enum class was provided and displayed on Swagger documentation

**Other**
If you have questions, I will answer it with pleasure.
if you want I describe this new feature in a documentation, I can write this, no problem.


MichakrawSB:

PR #2913, opened 2018-08-12
Update Reader.java

poazy:
Fix the problem of adding a @RequestBody annotation to the JAX-RS POST method can cause openapi.json 's requestBody not to read the @FormParam parameter


joschi:
@frantuma I think we can safely close this PR.

Issue #2940, opened 2018-08-30
Update jackson library to latest stable


scadgek:
I see the 1.5.x stable version currently uses jackson version 2.9.5. Shouldn't there be any issues with updating it to 2.9.6 and producing a release to Maven Central? Specifically I need this to avoid dependency version mismatch in my project.

frantuma:
we could and will update, but you can declare the dependency to 2.9.6 at top level, and that will be picked up

scadgek:
I'm aware of the workarounds, but ideally I'd like to avoid them. Having version 2.9.6 in the swagger-core itself will provide a better guarantee that it's an acceptable version for proper work of the library. Although I understand that it's not an issue in this specific case (where the version mismatch is 2.9.5. vs 2.9.6).

frantuma:
implemented in #3123  and #3091

PR #2941, opened 2018-08-30
Update jackson version to the latest stable

scadgek:
This is a fix for the issue #2940


jmini:
This can be closed as it is fixed with #3091

PR #2968, opened 2018-09-27
:triangular_ruler: Preserve order in maps accross the OAS model

quilicicf:
### Issue

When adding elements in a map in the OAS POJOs then serializing, the order in which the elements were added is not preserved.

Although JSON maps are supposed not to enforce an order on their elements, in practice this is implemented in all serializers and used in lots of products. 

We are converting API definitions from different languages and need the order to be preserved.

We already contributed [a similar fix](https://github.com/swagger-api/swagger-core/pull/2022) in version 2 of OAS.


frantuma:
Thanks!

quilicicf:
You're welcome :-)

Issue #2966, opened 2018-09-24
io.swagger.v3.jaxrs2.Reader#resolveApplicationPath does not work with proxies


nelsongraca:
The current implementation of resolveApplicationPath fails to find the @ApplicationPath annotation if the application is a proxy, it should search for the annotation in the superclass too.

I can provide a PR with a fix if needed.

frantuma:
Thanks for reporting this, and yes a PR is welcome!

nelsongraca:
Ok, I will work on it and submit it ASAP since this would be useful for me.

frantuma:
I guess it's a good approach, shouldn't have side effecs

nelsongraca:
Hi the PR is submitted: https://github.com/swagger-api/swagger-core/pull/2974

frantuma:
Thanks! PR merged, closing ticket

jgangemi:
this issue is still broken w/ version `2.0.8`. i'm using `helidon`, which in turn uses `weld` and stepping through the code, it seems the logic is reversed. the `ResourceConfig` is found where the weld proxy is looked for and the proxy is found where the resource config is looked for, specifically when `application.getClass().getMethod("getApplication", null);` is invoked.

PR #2974, opened 2018-10-08
Fix for Issue #2966

nelsongraca:
The annotation @ApplicationPath is now scanned on the superclasses in order to work as intended with Weld proxies.


frantuma:
Thanks!

Issue #2994, opened 2018-10-26
Swagger's Lack of Customization


duckladydinh:
Hello, I'd love to use Swagger, but there are certain things that I cannot achieve when trying to generate Swagger Script from Java code using the swagger-core dependencies. Some of which are:

1. Fullname problem: When generating Swagger Script, two classes of the same name in two different packages get messed up. Is there a way to support package name also?

2. Unnecessary (Object) in generic types like Query<T> will generate Query(Object) instead of Query. Can we remove the Object part?

3. Enum: As everyone knows, enum has 2 different methods name() and toString() and in my case name() is needed. Is there any way to use name() in Swagger?

4. It is really annoying when all the enums are inlined. It is better to have 1 enum class and have a reference to it when we need. How can we achieve that?

5. Schema(allOf = {Parent}) is specified but all the fields of the parents are still available. It does not make any sense at all, right? How can we hide them?

6. I prefer building Swagger Scripts 'manually' by creating a OpenAPI object first and add classes to it since the automated method is poorly implemented. How can I add classes to it easier?

Last, but not least, sometimes I want to expose private fields also, but Jackson completely lacks the means to achieve that. Can you maybe expose the JsonObjectMapper so that we can control the visibility of the fields?

Thank you. Looking forward to hearing from you.

MichakrawSB:
Closing this issue as part of our https://github.com/swagger-api/swagger-core/issues/5074.
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.
Thank you for your understanding!

PR #2996, opened 2018-10-26
Fixing Issue 2994 - Adding Various Customizations for Swagger

duckladydinh:
For the issue: https://github.com/swagger-api/swagger-core/issues/2994
Dear, I want to propose some customization options for OpenAPI like:

OpenAPIBuilderOptions.USE_FULLNAME: Boolean -> Including package name for models (in case you do not need to generate code and only want document)
OpenAPIBuilderOptions.OMIT_GENERIC: Boolean -> Omit the generic part of generic type. For example, consider class Query, it will exclude the part (Object).
OpenAPIBuilderOptions.USE_ENUMNAME: Boolean -> Force use of enum.name() instead of enum.toString().
OpenAPIBuilderOptions.RECYCLE_ENUM: Boolean -> Recycle enums (you do not need to inline enum all the time, just declare 1 as a normal enum class and use reference to it)
OpenAPIBuilderOptions.HIDE_PARENTS: Boolean -> Adding ability to hide all properties from parent classes that are provided in Schema(allOf = {...})
OpenAPIBuilderToolbox.addModel(OpenAPI, Type) -> Add a method to directly add additional models and their dependencies to OpenAPI object.
Expose JsonObjectMapper in ObjectMapperFactory.defaultMapper to control visibility.


MichakrawSB:

Issue #2972, opened 2018-10-02
Where there are multiple fields with same generic definition Map<K, V> (e.g Map<String, String>) then only the first field is rendered in the example text


camlane:
Where there are multiple fields with same generic definition Map<K, V> (e.g Map<String, String>) then only the first field is rendered in the example text

When I have a response class with more than one field of type Map<String, String> only the first field will render 
in the 'Example Value' part of the response definition in swagger ui

```
private Map<String, String> myField1;
private Map<String, String> myField2;
```

Only the below is output
```
"myField": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    "additionalProp3": "string"
  }
```

However I can see the below lines being output to the log twice.

io.swagger.v3.core.converter.ModelConverterContextImpl resolve: resolve [map type; class java.util.Map, [simple type, class java.lang.String] -> [simple type, class java.lang.String]]

wcicola:
I have been dealing with this issue for the better part of a week and finally managed to nail down the root. The issue begins at line 613 of ModelResolver:

`property = context.resolve(aType);`

ModelConverterContextImpl  (line 77) returns a cached Schema instance if it has already been processed:

` if (processedTypes.contains(type)) {
            return modelByType.get(type);
}`

ModelResolver at line 666 changes the name of that instance and adds it to the list of properties:
`property.setName(propName);
...
props.add(property);
`

If you use a debugger and inspect _props_ after this line, you can see that it actually contains multiple Schema instances--**each** with the name of the most recently parsed field. Then at line 674, each instance gets added to _modelProps_ :
`
for (Schema prop : props) {
            modelProps.put(prop.getName(), prop);
        }
`
And of course, since they have all been assigned the same name, this results in only the final one being retained in the map. 

I changed ModelConverterContextImpl to return a clone of the cached instance, and it now includes all fields:

`{
  "type" : "object",
  "properties" : {
    "stringsMap1" : {
      "type" : "object",
      "additionalProperties" : {
        "@class" : "io.swagger.v3.oas.models.media.StringSchema",
        "type" : "string"
      }
    },
    "stringsMap2" : {
      "type" : "object",
      "additionalProperties" : {
        "@class" : "io.swagger.v3.oas.models.media.StringSchema",
        "type" : "string"
      }
    }
  }
}
`
I will refine this code to only do the cloning when the AnnotatedType is of a Map and do some more testing, and can then create a PR if no other issues arise.

camlane:
[...]

PR #2997, opened 2018-10-26
ref #2972 - Proper handling of multiple Map fields with same generic …

wcicola:
Fix for issue #2972:
ModelConverterContextImpl now returns clone of cached Schema instance when that Schema's type is a Map in order to avoid bug where ModelResolver operates on the same instance.


frantuma:
replaced by #3001

Issue #2189, opened 2017-04-28
SubTypes including annotated class causes java.util.ConcurrentModificationException


rasmusvhansen:
Using Swagger 1.5.12

```java
@ApiModel(subTypes = {OtherClass.class, SameClass.class})
public class SameClass {
...
```

Causes `java.util.ConcurrentModificationException` in `ModelResolver`

Workaround of course is to not make that mistake in the annotation.
However, tracking down the error took a long time, so I would suggest that Swagger either:

1. Ignored the superfluous annotation or
2. Provided a better error message

SamSuffit:
Hello, I had the same problème. Si I make a PR which correct it. 
I'll have to used it in my product, any idea if it will be included ?
And if so then when the next 1.5 release is due ?

Felk:
The same thing happens with `@JsonSubTypes` annotations, like this:

    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME)
    @JsonSubTypes({
            @JsonSubTypes.Type(value = SameClass.class, name = "SameClass")
    })
    public class SameClass {
        // ...
    }

Using `io.swagger.core.v3:swagger-jaxrs2:2.0.8`.

Stacktrace:

```
java.util.ConcurrentModificationException
		at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:719)
		at java.base/java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:751)
		at java.base/java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:749)
		at deployment.mydeployment.war//io.swagger.v3.core.jackson.ModelResolver.removeParentProperties(ModelResolver.java:1257)
		at deployment.mydeployment.war//io.swagger.v3.core.jackson.ModelResolver.resolveSubtypes(ModelResolver.java:1214)
		at deployment.mydeployment.war//io.swagger.v3.core.jackson.ModelResolver.resolve(ModelResolver.java:708)
		at deployment.mydeployment.war//io.swagger.v3.core.converter.ModelConverterContextImpl.resolve(ModelConverterContextImpl.java:90)
		at deployment.mydeployment.war//io.swagger.v3.core.jackson.ModelResolver.resolveSubtypes(ModelResolver.java:1157)
		at deployment.mydeployment.war//io.swagger.v3.core.jackson.ModelResolver.resolve(ModelResolver.java:708)
[...]

PR #3031, opened 2018-11-26
Take current class list in subTypes ConcurrentModificationException error fixes(#2189)

SamSuffit:
Hi,

I've the problem myself.
If a parent class is not bastract it failed.
So here is a pr to correct ths bug. 

Can you keep me inform ?


frantuma:
Thanks! #2189 fixed by #3299 for 1.5 and #3298 for 2.x, closing this PR

PR #3052, opened 2018-12-12
Removed dependency on the Servlet API from modules/swagger-jaxrs2

uldall:
Some web servers (such as Grizzly) can use JAX-RS without the Servlet API.
This change is to support that scenario.

This PR is not ready to merge, but could serve as a starting point for removing the dependency on the Servlet API from modules/swagger-jaxrs2.


MichakrawSB:

PR #3069, opened 2019-01-10
use JakartaEE Maven coordinates

deki:
The Maven coordinates for JAXB und REST API changed and should be update. See also https://wiki.eclipse.org/New_Maven_Coordinates

The new Servlet API artifact is not available on central and can be changed later. See also discussion here: https://www.eclipse.org/lists/servlet-dev/msg00069.html


frantuma:
Thanks, EE coords updated in #3467

Issue #3082, opened 2019-01-23
jaxrs generic return type is abstract class.


andrejpetras:
Hi,
I get always the abstract (super) type and not the generic type in the response definition for my jaxrs service.
Java code example

```java
@Stateless
@Path("token")
@Consumes(MediaType.APPLICATION_JSON)
@Produces(MediaType.APPLICATION_JSON)
public class ProcessTokenRestService extends AbstractEntityRestService<ProcessTokenDTO> {
}

public abstract class AbstractEntityRestService<DTO extends PersistentDTO> implements EntityRestService<DTO> {

    public DTO create(DTO object) throws Exception {
        return null;
    }

}
```
Part of the openapi.json
```json
  "paths" : {
    "/token" : {
      "post" : {
        "operationId" : "create",
        "requestBody" : {
          "content" : {
            "application/json" : {
              "schema" : {
                "$ref" : "#/components/schemas/ProcessTokenDTO"
              }
            }
          }
        },
        "responses" : {
          "default" : {
            "description" : "default response",
            "content" : {
              "application/json" : {
                "schema" : {
                  "$ref" : "#/components/schemas/PersistentDTO"
                }
              }
            }
          }
        }
      }
    }
  },
```
I would like to have this for the response.
```json
        "responses" : {
          "default" : {
            "description" : "default response",
            "content" : {
              "application/json" : {
                "schema" : {
                  "$ref" : "#/components/schemas/ProcessTokenDTO"
                }
              }
            }
          }
        }
      }
```

The response type should be ProcessTokenDTO (as input parameter) and not PersistentDTO. 
I use the test1.zip project to reproduce this issue. How to reproduce: mvn clean install

Thanks
Andrej

[test1.zip](https://github.com/swagger-api/swagger-core/files/2788755/test1.zip)

andrejpetras:
I check the Reader class in the jaxrs module. The parameter types are taken from the AnnotatedMethod object. For the return type, the type from the java method is used. 
I create simple fix and a pull request.

Could you please look at it if it is ok or create another solution for this issue.

Thanks
Andrej

frantuma:
Thanks! merged #3096 (includes #3083)

PR #3083, opened 2019-01-23
#3082 jaxrs generic return type is abstract class

andrejpetras:
Issue #3082 

I check the Reader class in the jaxrs module. The parameter types are taken from the AnnotatedMethod object. For the return type, the type from the java method is used. 
I create simple fix and a pull request.

Could you please look at it if it is ok or create another solution for this issue.

Thanks
Andrej


frantuma:
Thanks! included in #3096

Issue #3084, opened 2019-01-24
Models should not be equal if they have different properties


nzolotova:
Good day everyone,
we faced with an issue while generating an openApi with the version 1.5. 

Two models, that have different properties, occurred to be equal, but shouldn't be. 

I created a test in Kotlin to compare two models with properties, that have different nested properties, and it failed:
```
@Test
	fun `should compare two models with different properties`(){

		val model1 = givenModel("order")
		val model2 = givenModel("attributes")

		then(model1 == model2).isFalse()
	}

	private fun givenModel(subPropertyName: String): ModelImpl {
		val model = ModelImpl()
		model.properties = givenProperties(subPropertyName)

		return model
	}

	private fun givenProperties(subPropertyName: String): Map<String, ObjectProperty> {
		return mapOf("_links" to givenProperty(), "page" to givenProperty(), "_embedded" to givenProperty(mapOf(subPropertyName to givenProperty())))
	}

	private fun givenProperty(properties: Map<String, ObjectProperty> ? = emptyMap()): ObjectProperty {
		return ObjectProperty(properties)
	}
```
For me it looks like the issue happens, because ```ObjectProperty``` does not overwrite ```equals()``` method from ```AbstractProperty```, that's why map with nested properties defined only in the class ```ObjectProperty```  is not compared. 

Kind Regards,
Natalia Zolotova

frantuma:
implemented in #3085

PR #3085, opened 2019-01-24
Consider properties for equality of ObjectProperty

mduesterhoeft:
Addresses https://github.com/swagger-api/swagger-core/issues/3084

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/properties/ObjectProperty.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/properties/ObjectProperty.java
@@ -142,6 +142,10 @@ public class ObjectProperty extends AbstractProperty implements Property {
 
     @Override
     public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+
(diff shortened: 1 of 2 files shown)


mduesterhoeft:
Hi there, what do you think? Does this have a chance to be merged soon? @frantuma

modules/swagger-models/src/main/java/io/swagger/models/properties/ObjectProperty.java
```diff
@@ -139,4 +139,25 @@ public ObjectProperty example(Object example) {
         this.setExample(example);
         return this;
     }
+
+    @Override
+    public boolean equals(Object o) {
```
frantuma: can you possibly just add
```
if (this == o) {
  return true;
}
```

frantuma:
Thanks!

mduesterhoeft:
Thanks for merging.

mduesterhoeft:
@frantuma do you already have plans to release the next 1.5 version?

frantuma:
It will be within 1-2 weeks

PR #3102, opened 2019-02-01
fix #3094 - avoid duplicate fields in schema classes

### LINKED ISSUE #3094: Schema subclasses contain duplicate fields
The way in which the various Schema subclasses (e.g. ArraySchema, StringSchema, etc.) are defined means that they are duplicating certain fields.   For example, StringSchema is defined like this:
```
public class StringSchema extends Schema<String> {
    private String type = "string";
...
}
```
This causes StringSchema to have a DIFFERENT type field than is found in Schema<T>.

To alleviate this, I'm proposing a slight change to Schema like this to introduce a ctor:
```
public class Schema<T> {
    ...
    public Schema(String type, String format) {
        this.type = type;
        this.format = format;
    }
    ...
}
```
and then StringSchema could be changed like this to use this new ctor:
```
public class StringSchema extends Schema<String> {

    public StringSchema() {
        super("string", null);
    }
    ...
}
```
Then BinarySchema could be changed like this:
```
public class BinarySchema extends Schema<byte[]> {

    public BinarySchema() {
        super("string", "binary");
    }
    ...
}
```
And ByteArraySchema could be changed like this:
```
public class ByteArraySchema extends Schema<byte[]> {

    public ByteArraySchema() {
        super("string", "byte");
    }
    ...
}
```
and most of the other Schema subclasses could be changed in a similar way.

Before I submit a PR for these changes, I just wanted to introduce the issue and see if there is general agreement on the solution.   I don't think this will make any functional difference in the classes but would make things a little easier when debugging and would just make good "java sense" I think.
Comments?
cc: @wing328

padamstx:
Why:
The various Schema subclasses (ArraySchema, BinarySchema, etc.) are
defined such that they actually duplicate some of the fields from
the Schema super class (e.g. type, format, default, etc.).  This
can create confusion when examining instances of these classes within
a debugger as you will see two "type" fields for example, the base class'
"type" field will be set to null and the subclass' "type" field will be set
to "array" (or "binary", or "boolean", etc.).   We have similar issues with
the format and default fields.

What:
In this PR, I've fixed this by adding a protected ctor to Schema, and default ctors to
the subclasses, plus I've removed some getters/setters from the subclasses
that are no longer needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ atlassian-ide-plugin.xml
 *.iml
 .java-version
 sonar-project.properties
+test-output/
--- a/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/ArraySchema.java
+++ b/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/ArraySchema.java
@@ -22,26 +22,15 @@ import java.util.Objects;
  * ArraySchema
  */
 
-public class ArraySchema extends Schema {
-    private String type = "array";
-    private Schema items = null;
+public class ArraySchema extends Schema<Object> {
+    private Schema<?> items = null;
 
-    /**
-     * returns the type property from a ArraySchema instance.
-     *
-     * @return String type
-     **/
-
-    public String getType() {
-        return type;
-    }
-
-    public void setType(String type) {
-        this.type = type;
+    public ArraySchema() {
+        super("array", null);
     }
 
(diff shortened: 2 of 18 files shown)


padamstx:
Please disregard the name of my bug branch.   I don't know where I got "342".  Should have named it "issue-3094" :)

padamstx:
cc: @frantuma

frantuma:
Thanks!

PR #3108, opened 2019-02-05
Throw exception when values in Parameter and Schema are invalid

### LINKED ISSUE #3097: Deserializer not failing when data types are invalid
For a definition containing an invalid schema data type, the deserializer doesn't fail.
In the following example, `type: float` of `schema` is invalid.
```
openapi: 3.0.0
info:
  title: Test
  description: This is a test spec
  version: 0.0.1
paths:
  /item:
    get:
      summary: "Provides item"
      parameters:
        - name: "itemId"
           in: query
           description: "Id of item"
           required: false
           schema:
             type: float
           example: 102873
    responses:
      200:
        description: "Success"
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/Response"
```
Same if `in` is set to an invalid value.

I believe [ParameterDeserializer](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/ParameterDeserializer.java) and [ModelDeserializer](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/ModelDeserializer.java) should throw exceptions when the values don't match any of the possible values.

timemanx:
Fixes #3097


MichakrawSB:

PR #3112, opened 2019-02-05
fix #3111 Some primitive types in @ApiResponse.content are reported i…

stauvel:
For reporting ApiResponse.content().schema() types, I'm using the corresponding PrimitiveType when possible or else the current getName().toLowerCase() method.

Thanks team !

PR #3114, opened 2019-02-06
Primitive types format

stauvel:
#3112 doublon with a Suggestion to customize format with implementation attribute.

So you can reach a binary object like this :

    @Schema(implementation = Byte.class, format = "binary")


frantuma:
Thanks! included and replaced by #3468

PR #3119, opened 2019-02-11
Supporting  inside allOf properties

okalentiev:
Currently, composed properties are not working correctly with `allOf` properties that contain $ref. It adds them as properties, but not matching correctly with the sibling properties. As mentioned in https://github.com/swagger-api/swagger-core/issues/2621 the following structure is valid:

```
defWithDep:
  type: object
  properties:
    def:
      x-thingy: thing
      allOf:
      - $ref: "#/definitions/standaloneDef"
```

This PR contains a proposal for handling this case.


MichakrawSB:

PR #3125, opened 2019-02-16
fix #3088 - fixes boolean @type annotation

### LINKED ISSUE #3088: Schema attribute "type" doesn't work for value "boolean"
`@Schema(`type` = "boolean") test_boolean:MyBooleanClass = false`

Specifying type=boolean in the schema always produces type:Object in the resulting json. I have tracked the issue down to PrimitiveType.java in package io.swagger.v3.core.util, line 215

It appears that
`dms.put("boolean", "boolean");`
Should be :
`dms.put("boolean_", "boolean");`
Because of this function
```
public static PrimitiveType fromTypeAndFormat(String type, String format) {
        if (StringUtils.isNotBlank(type) && type.equals("object")) {
            return null;
        }
        return fromName(datatypeMappings.get(String.format("%s_%s", StringUtils.isBlank(type) ? "" : type, StringUtils.isBlank(format) ? "" : format)));
    }
```

always formats the string with an underscore.

AdrianMF:
`@Schema(`type` = "boolean") test_boolean:MyBooleanClass = false`

Specifying type=boolean in the schema always produces type:Object in the resulting json. I have tracked the issue down to PrimitiveType.java in package io.swagger.v3.core.util

It appears that
`dms.put("boolean", "boolean");`
Should be :
`dms.put("boolean_", "boolean");`

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
@@ -212,7 +212,7 @@ public enum PrimitiveType {
         dms.put("string_date-time", "date-time");
         dms.put("string_partial-time", "partial-time");
         dms.put("string_password", "password");
-        dms.put("boolean", "boolean");
+        dms.put("boolean_", "boolean");
         dms.put("object_", "object");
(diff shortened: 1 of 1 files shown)


frantuma:
well spotted, thanks!

PR #3142, opened 2019-03-01
Removed guava from swagger-jaxrs

albertotn:
My proposal is to remove guava from swagger-jaxrs, because:

- not needed at compile time
- not needed at test time
- not needed a runtime

As main benefit, guava jar will not be part of dependencies and also all transitive dependencies


frantuma:
closing as guava is used by module (even if removed would be picked as transitive deps from core

Issue #3133, opened 2019-02-22
Feature Request: Swagger Maven Plugin should be able to read Swagger config


patrickfav:
Currently the maven plugin allows for reading a `openapiFilePath` which expects the file to have the `openApi` element as root, like so:

```
info:
  version: '1.0.0'
  title: REST API
  description: 'The REST API description.'
  contact:
    email: mail@mail.com
  license:
    name: All Rights Reserved
    url: https://license.com
  x-logo:
    url: https://logo.svg
    backgroundColor: "#FFFFFF"
externalDocs:
  description: Docs
  url: 'http://docs/Glossar'
```

However, if one already uses Swagger to dynamically serve the doc over a e.g. Servlet, one usually has a Swagger configuration file with this scheme:

```
resourcePackages:
  - my.package
prettyPrint: false
cacheTTL: 0
openAPI:
  info:
    version: '1.0.0'
    title: REST API
    description: 'The REST API description.'
    contact:
      email: mail@mail.com
    license:
      name: All Rights Reserved
      url: https://license.com
    x-logo:
      url: https://logo.svg
      backgroundColor: "#FFFFFF"
  externalDocs:
    description: Docs
    url: 'http://docs/Glossar'
```

As there is no trivial way to include yaml in yaml, I now would have to maintain 2 openApi configs.

**My suggestion is therefore this:**
Provide a new config named `configurationFilePath` ~swaggerConfigurationFilePath~ which accepts a path to a configuration file as described in https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Integration-and-Configuration#configuration-file. This file would be parsed before the other configs and has less precedence as a maven plugin defined config. E.g. if provided a `resourcePackages` defined config in maven pom would have higher precedence of the same config defined in the swagger yaml config.

-----------------

Is the team interested in such a feature? I am willing to provide a PR for this.

frantuma:
This sound very good indeed, and it is for sure interesting, possibly applicable also to gradle-plugin. Maybe the config name could be `configurationFilePath`

patrickfav:
Alright, then Id prepare the PR in the next couple of days. I would start with the maven plugin. If we agree on the exact requirement and behaviour, this can then be ported to the Gradle plugin.

> Maybe the config name could be `configurationFilePath`

You are right, this is probably better.

frantuma:
excellent thx!

frantuma:
#3150 merged, thanks!

[...]

PR #3150, opened 2019-03-10
Add swagger config file option to maven plugin (#3133)

patrickfav:
Currently the maven plugin only accepted OpenAPI files. This adds
the option to provide a swagger configuration (e.g. prettyPrint),
so if you already have a config file for say, serving a life api doc,
you don't have to define it twice.

This heavily refactors the maven plugin and adds a bunch of tests. Also
the readme is extended to reflect the new parameter. The error handling
for file reading was changed to fail-fast, meaning, if a user provides
the config, the file must exist and must be readable, otherwise an error
is thrown. This makes it easier to correctly set the plugin up.

refs #3133


modules/swagger-maven-plugin/README.md
```diff
@@ -80,3 +109,15 @@ Parameter | Description | Required | Default
...
+When providing both `openapiFilePath` and `configurationFilePath` and other maven configuration, this
+is the other in which the settings are applied (higher number is lower precedence - first 1 is applied, then 2 and so on):
+
+1. Swagger configuration file (_Swagger_ + _OpenAPI_)
+1. OpenAPI configuration file (_OpenAPI_)
+1. Maven configuration parameters (_Swagger_)
+1. Embedded annotations (_OpenAPI_)
+
+Note that some layers merge the OpenAPI model and some just overwrite it completely. This feature is not
+meant to merge different configs.
```
frantuma: This section is not fully clear to me, particularly the meaning of `Embedded annotations` and the swagger/openapi within parenthesis); Possibly it would be sufficient something like:


_Since version `2.0.8`, `configurationFilePath` parameter is available, allowing to specify a path to a `[swagger configuration file](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Integration-and-Configuration#configuration)`; If single maven configuration parameters (e.g. `prettyPrint`) are also defined, these will overwrite any value set in configuration file; the same applies to  `openapiFilePath` which takes precedence over `openAPI` field in configuration file._

modules/swagger-maven-plugin/src/main/java/io/swagger/v3/plugin/maven/SwaggerMojo.java
```diff
@@ -191,19 +334,38 @@ public void execute() throws MojoExecutionException, MojoFailureException
...
+
     public String getOpenapiFilePath() {
         return openapiFilePath;
     }
 
+    String getConfigurationFilePath() {
+        return configurationFilePath;
+    }
+
+    void setContextId(String contextId) {
```
frantuma: is this just for tests?

patrickfav: Correct. That is why I left it with package visibility. Do you have a suggestion to better mark it as test method (Im familiar with a `@VisibleForTesting` annotation found in `com.google.common.annotations`)?

modules/swagger-maven-plugin/README.md
patrickfav: Yours is probably clearer, will change.

patrickfav: Done.

patrickfav:
Hi, I've updated the Readme. Please let me know if there is anything you feel needs to be changed.

PR #3162, opened 2019-03-28
Fix bug "always true" condition.

amatskevich:
jasonSerialize.as() never return null according API. So there is replacing correct JavaType class to java.lang.Void in case missed parameter "as". As result I've got empty body of example on UI.
See, https://static.javadoc.io/com.fasterxml.jackson.core/jackson-databind/2.9.8/com/fasterxml/jackson/databind/annotation/JsonSerialize.html#as--


frantuma:
replaced by #4603

PR #3164, opened 2019-04-02
Clone filtered operation

nicolashenry:
The filtered operations should be cloned because the first time the parameters are filtered, the original operations are also modified.

Note that this PR also prevent the filtered operation to create an empty array of parameters which does not exist in the original operation.


nicolashenry:
@frantuma Does this PR miss something in order to be reviewed?

frantuma:
eventually.. thanks!

PR #3165, opened 2019-04-02
Use recursive search for filtered definitions

nicolashenry:
Currently, the filtered definitions are not searched recursively so if you declare MyClass1 > MyClass2 > MyClass3 only the first two levels are retrieved (MyClass1 and MyClass2) because MyClass1 schema only contains the reference to MyClass2.
This PR fixes this.


nicolashenry:
Seems to be fixed

PR #3186, opened 2019-04-23
Ticket-3115: Documentation for the swagger-gradle-plugin appears to give incorrect artifact path

bronogard:
- Fix: show correct GAV coordinates for swagger-gradle-plugin when using the Gradle buildscript Syntax


frantuma:
Thanks!

PR #3204, opened 2019-05-17
Remove Jaxb Dependency

MarcMil:
As far as I know, the javax.xml.bind package is part of the JRE since at least version 6. As such, it should not be necessary to pull in a third party package containing an implementation of that very package.

I could successfully build swagger without it and I am not sure why it is necessary, so I'd suggest to remove it.

### DIFF
--- a/modules/swagger-core/pom.xml
+++ b/modules/swagger-core/pom.xml
@@ -54,11 +54,6 @@
     </build>
     <dependencies>
         <dependency>
-            <groupId>javax.xml.bind</groupId>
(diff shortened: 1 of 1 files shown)


bronogard:
The `javax.xml.bind:jaxb-api:2.3.0` library is relevant once running with *JDK version >= 9*.

MarcMil:
Oh I see. That's unfortunate. Seems to be counter intuitive with the module concept, but it seems indeed the case that for the newest Java versions it is the recommended way.

PR #3206, opened 2019-05-22
Fix BaseApiListingResource when concurrent execution

awojcicki:
In some scenarios `process` method doesn't return proper Swagger definition (scan isn't performed)

Sample replication scenario:
- two threads are inside process method
- new empty Swagger instance is created for both on them
- synchronized block is fired for first one producing correct swagger and marks as initialized 
- seconds thread starts synchronized block but because it was marked as initialized scan is not performed
- second thread returns incorrect swagger definition


awojcicki:
@frantuma hey, we identified issue in `BaseApiListingResource` If you have some time please take a look on this fix. Thank you!

PR #3209, opened 2019-05-24
Respect required status of form parameters.

UKreher:
When merging form parameters to one schema for the response body, the information about whether a parameter is taken over to the merged schema.


frantuma:
Thanks!

PR #3210, opened 2019-05-24
Do not set default type string for composed schema.

UKreher:
A composed schema usually has sub schemas. So there is no need to set the default type "string" unless explicitly set. Setting it may lead to an unexpected semantic of the composed schema.


frantuma:
Thanks for the PR! can you just update the expected test spec in `io.swagger.v3.jaxrs2.ReaderTest.testGetResponsesWithComposition` (which was indeed not good as it had `string` as type for the composed schema)?

thanks!

UKreher:
Sure, now that I know where to find the test. ;-)

Thank you!

PR #3211, opened 2019-05-24
Make parseMethod protected thus improving extensibility.

UKreher:
Making the main Reader.parseMethod protected greatly improves extensibility.

The method being private requires me to override the 3 corresponding public methods. Additionally, I need a callback reference (`java.util.function.Function`) for each of them just to simulate a `super` call in the one “overriding“ method.


frantuma:
Thanks!

Issue #2992, opened 2018-10-24
Properties without @Schema annotation of type LocalTime missing in openapi.json


markusdemetz:
I use `swagger-annotations (2.0.5)` to generate my `openapi.json` file.

When using a property of type `java.time.LocalTime` named `a`, it produces the following (shown in Swagger UI):

![a](https://user-images.githubusercontent.com/30077873/47442241-58df9b80-d7b2-11e8-9947-6b498af2a2d5.JPG)

When I have 2 or more properties of type `java.time.LocalTime` and these properties are not annotated with `@Schema` and a name attribute, then only the last one is present in the `openapi.json` file.
I.e.:
```
private LocalTime a;
private LocalTime b;
private LocalTime c;
```
with corresponding getters and setters on an API model only outputs the last property of type `java.time.LocalTime`:

![schemaarticle](https://user-images.githubusercontent.com/30077873/47443480-ac52e900-d7b4-11e8-83c9-09db55d8fb80.JPG)

Usually I do not annotate my properties with `@Schema` when not needed.
Do I miss something here or is this a bug?
Thanks!

frantuma:
Thanks for reporting this; it's indeed a bug, fixed in latest `2.0.6-SNASPHOT` available on sonatype; alternatively, #2993 /  `2.0.6-SNASPHOT` also adds support for `LocalTime` resolved globally as a primitive string type with [rfc3339](https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14) format `partial-time`, either using convenient 

`PrimitiveType.enablePartialTime();` 

or 

```
PrimitiveType.customClasses().put("org.joda.time.LocalTime", PrimitiveType.PARTIAL_TIME);
PrimitiveType.customClasses().put("java.time.LocalTime", PrimitiveType.PARTIAL_TIME);
``` 
in any bootstrap code.

in this case, the outcome would be something like;

```
a:
  type: string
  format: partial-time
b:
  type: string
  format: partial-time
c:
  type: string
  format: partial-time
```

Closing ticket, please reopen if you're still experiencing issues.

emigioria:
org.joda.time.LocalDateTime should be added to DATE_TIME group.

ryancerium:
I opened #3212 to address `org.joda.time.LocalDateTime` @emigioria .

frantuma:
fixed in #3212, thanks!

PR #3212, opened 2019-05-24
Serialize org.joda.time.LocalDateTime as a string

ryancerium:
Also alphabetized the classes in the parameter list.


ryancerium:
Related to issue #2992

frantuma:
thanks!

Issue #3197, opened 2019-04-30
allOf is not appearing properly in generated schema


zaverrabadiya:
I am using io.swagger.core.v3  <version>2.0.8</version> to generate OpenAPI Schema. I am using inheritance with polymorphism to define my models. When I generate the schema, subtype models are not generated properly. `allOf` should have model definition as child of it (inline) but in my schema `allOf` and rest of the subtype schema are siblings. Which breaks generation of client, particularly properties are not generated. 

For example,
my subtype `RaceCar` looks like: `RaceCar:
      required:
      - carMetaData
      type: object
      properties:
        id:
          type: integer
          format: int64
        model:
          type: string
      allOf:
      - $ref: '#/components/schemas/Car'` 

which has `allOf`, `type` and `properties` as siblings. 

It supposed to be: `RaceCar:
      allOf:
        - $ref: '#/components/schemas/Car'
        - type: object
          required:
            - id
          properties:
            id:
              type: integer
              format: int64
            model:
              type: string
          x-translation:
            - model`

### My java models ###
```
@Schema(discriminatorProperty = "type", discriminatorMapping = {
        @DiscriminatorMapping(value = "RaceCar", schema = RaceCar.class),
        @DiscriminatorMapping(value = "SportCar", schema = SportCar.class)
},
oneOf = {
        RaceCar.class,
        SportCar.class
})
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME,
        property = "type")
@JsonSubTypes({
    @JsonSubTypes.Type(value = RaceCar.class),
    @JsonSubTypes.Type(value = SportCar.class)
})
public abstract class Car {

    private CarMetaData carMetaData;

    public CarMetaData getCarMetaData() { return carMetaData; }
}

public class RaceCar extends Car {

    @JsonProperty(required = true)
    public Long id;

    public String model;

    @Schema(required = true)
    public CarMetaData carMetaData;
}

public class SportCar extends Car {

    @JsonProperty(required = true)
    public Long id;

    public String model;
}
```

### Actual generated schema ###
```
schemas:
    Car:
      type: object
      properties:
        carMetaData:
          $ref: '#/components/schemas/CarMetaData'
      discriminator:
        propertyName: type
        mapping:
          RaceCar: '#/components/schemas/RaceCar'
          SportCar: '#/components/schemas/SportCar'
      oneOf:
      - $ref: '#/components/schemas/RaceCar'
      - $ref: '#/components/schemas/SportCar'
[...]

Issue #3030, opened 2018-11-22
Wrong schema generation for composed types


m-kay:
When configuring inheritance through allOf property in the `@Schema` annotation, the generated OAS 3 specification is wrong. It is indeed shown corectly in swagger-ui, however when using the specification to generate client code, the generated code is not correct.

With this java classes
```
class Parent {
	private String sharedProperty;
}

@Schema(allOf = Parent.class)
class Child extends Parent {
	private String property;
}
```
the following OAS schemas are generated:
```
...
{
  "components": {
    "schemas": {
      "Parent": {
        "type": "object",
        "properties": {
          "sharedProperty": {
            "type": "string"
          }
        }
      },
      "Child": {
        "type": "object",
        "properties": {
          "property": {
            "type": "string"
          }
        },
        "allOf": [
          {
            "$ref": "#/components/schemas/Parent"
          }
        ]
      }
    }
  }
}
```
According to the [OAS](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/) correct would be:
```
...
{
  "components": {
    "schemas": {
      "Parent": {
        "type": "object",
        "properties": {
          "sharedProperty": {
            "type": "string"
          }
        }
      },
      "Child": {
        "allOf": [
          {
            "$ref": "#/components/schemas/Parent"
          },
          {
            "type": "object",
            "properties": {
              "property": {
                "type": "string"
              }
            }
          }
        ]
      }
    }
  }
}
```

dalewking:
FYI, in your example you named it baseProperty and in the OAS you switched to sharedProperty

dalewking:
See #3049 which points out that if this we're fixed to generate the correct OAS, that OAS would not be parsed correctly and would end up duplicating the "property" property.

dalewking:
I wonder if the order of items in the allOf matters. You show it with the $ref last. Can you try it with the $ref first?

m-kay:
Yes exactly, refs have to go after the actual object

m-kay:
Sorry I was wrong. Actually ref should be first, otherwise the Child class does not inherit from Parent at all.
See the example https://github.com/m-kay/oasgenerateallofexample, generate the code with `mvn clean install` and compare the generated code in the packages reffirst and reflast.

frantuma:
[...]

PR #3228, opened 2019-06-27
Fix inheritance

nitinjavakid:
This should fix issue #3197


UKreher:
This may also fix #3030.

frantuma:
Thanks! included and replaced by #3262

Issue #3253, opened 2019-07-26
Swagger does not generate documentation when implementing an  interface mehtod with a List<Some> as parameter.


rnacher:
hi!
swagger does not generate any endpoint documentation when the method implemented has a list a parameter.
Given the interface

`@Path("resources")
@Tag(name = "resource")
public interface TicketRNRAnnotatedInterface<T extends Number> {
  @POST
  Response postResource(List<T> payload);
}`

and the implementation

`public class TicketRNRConcreteImplementation implements TicketRNRAnnotatedInterface<Long> {
  @Override
  public Response postResource(List<Long> payload) {
    return null;
  }
}`

it does not generate any path for this method.
`Reader reader = new Reader(new OpenAPI());
OpenAPI openAPI = reader.read(TicketRNRConcreteImplementation.class);
 System.out.println(openAPI);`

> class OpenAPI {
    openapi: 3.0.1
    info: null
    externalDocs: null
    servers: null
    security: null
    tags: null
    paths: null
    components: null
}

I should expect to get the post method listed.

rnacher:
Added a PR with a failing test in order to show the bug.
https://github.com/swagger-api/swagger-core/pull/3254

rnacher:
Bug fix uploaded.

MichakrawSB:
Closing this issue as part of our https://github.com/swagger-api/swagger-core/issues/5074.
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.
Thank you for your understanding!

PR #3254, opened 2019-07-26
Swagger does not generate documentation when implementing an interface mehtod with a List<Some> as parameter. #3253

rnacher:
This PR is related to ticket 3253.
For now it just contains a test to show the bug.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/ReflectionUtils.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/ReflectionUtils.java
@@ -51,38 +51,53 @@ public class ReflectionUtils {
     }
 
     /**
-     * Checks if the method methodToFind is the overridden method from the superclass or superinterface.
+     * Checks if the method methodToFind is the overridden method from the superclass or superinterface. In most cases getMethods should be called instead.
      *
      * @param methodToFind is method to check
      * @param cls          is method class
      * @return true if the method is overridden method
      */
     public static boolean isOverriddenMethod(Method methodToFind, Class<?> cls) {
-        Set<Class<?>> superClasses = new HashSet<>();
-        for (Class c : cls.getInterfaces()) {
-            superClasses.add(c);
-        }
+      return !getMethods(cls).contains(methodToFind);
+    }
 
-        if (cls.getSuperclass() != null) {
-            superClasses.add(cls.getSuperclass());
-        }
-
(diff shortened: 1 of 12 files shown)


modules/swagger-core/src/main/java/io/swagger/v3/core/util/ReflectionUtils.java
```diff
@@ -51,39 +51,54 @@ public static Type typeFromString(String type) {
...
-                    if (method.getName().equals(methodToFind.getName()) && method.getReturnType().isAssignableFrom(methodToFind.getReturnType())
-                            && Arrays.equals(method.getParameterTypes(), methodToFind.getParameterTypes()) && !Arrays.equals(method.getGenericParameterTypes(), methodToFind.getGenericParameterTypes())) {
-                        return true;
+    /**
+     * Returns the list of methods of the given class and its hierarchy but without overridden methods.
+     * It is similar to Class.getMethods but overridden ones are not listed.
+     *
+     * @param cls          is the class
+     * @return the list of methods but without overridden methods
+     */
```
rnacher: The Class.getMethods already scans over the hierarchy so, It is not needed to load the hierarchy.

modules/swagger-core/src/test/java/io/swagger/v3/core/util/reflection/ReflectionUtilsTest.java
```diff
@@ -52,9 +71,11 @@ public void isOverriddenMethodTest() throws NoSuchMethodException {
 
         for (Method method : IParent.class.getMethods()) {
             if ("parametrizedMethod5".equals(method.getName())) {
```
rnacher: I think that there was a bug in the test. That method parametrizedMethod5 was hidden! and it should be listed.

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
```diff
@@ -374,19 +374,14 @@ public OpenAPI read(Class<?> cls,
         globalParameters.addAll(ReaderUtils.collectFieldParameters(cls, components, classConsumes, null));
 
         // iterate class methods
```
rnacher: Reflection.getMethods just returns the desired methods. No further filtering should be done on overriding.

rnacher:
This PR is deprecated by https://github.com/swagger-api/swagger-core/pull/3261

Issue #3257, opened 2019-07-29
External Spec validation of required properties isn't possible


Hoovs:
Hi team,
Not 100% the format so if wrong please let me know.

I am using the 1.5.22 tag of swagger. I am working on building a smaller swagger validator based on this tool chain. I want to produce an error saying you specified a required field but it's not in the definition. I have a swagger spec definition like below:
```
definitions:
  Thing:
    type: object
    required:
      - id
      - notHere
    properties:
      id:
        type: string
      with_underscore:
        type: string
```
So I wrote some code like:
```
Swagger spec = null;
try {
    spec = new SwaggerParser().read(path);
} catch (Exception e) {
// TODO
}

Map<String, Model> definitions = spec.getDefinitions();
Model model = definitions.get("Thing");
if (model instanceof ModelImpl) {
    ModelImpl modelImpl = (ModelImpl) model;
    if (modelImpl.getRequired() != null && modelImpl.getRequired().size() > 0) {
        Map<String, Property> properties = modelImpl.getProperties();
        List<String> requiredProperties = modelImpl.getRequired(); // This returns only id.
        for (String requiredProperty : requiredProperties) {
            if (!properties.containsKey(requiredProperty)) {
                //TODO something here
.....
```
So I looked at the code and getRequired only returns properties that are on the object that match the required.

Is there a way to make this work?

I added code to ModelImpl like:
```
@JsonIgnore
public List<String> getSpecSpecifiedRequired() {
    return required;
}
```
Which does work but is this the best way to do this and what are the odds of something like this making it in?

Thank you again for any help.

frantuma:
merged #3258

PR #3258, opened 2019-07-29
Per Issue 3257 here is an attempt at a fix

Hoovs:
This exposes the internal required field for external spec validation.
With this code the caller sees the actual fields needed in the spec vs
what the swagger parser produced.


joschi:
Refs #3257

frantuma:
thanks!

PR #3266, opened 2019-08-07
sort methods to make api generations reproducible

slavb18:


webron:
There's actually a downside to this. When it comes to the spec, the order is meaningless. I can see why some people may want it ordered, but they don't necessarily want it alphabetically. We used to have a way to control ordering in the past, but since the order information was removed from the spec, we decided to take it out.

We appreciate the effort with the PR but are inclined to not merge it.

slavb18:
downside of not ordering is that naming of operations is different
e.g. there is two methods with same name from different subresources

e.g. method "getData": resource.getData, resource.subresource.getData

one of them gets operation name "getData" second  "getData_1", 
and this depends on order of methods.

May be there should be configuration option to configure ordering of methods ?
like  JVM, NAME_ASCENDING in Junit @FixMethodOrder annotation

frantuma:
A solution has been implemented in #3740, see [PR comment](#3740) for details and usage scenarios. Closing this ticket, any further issues in this area should be reported in a new or not closed ticket.

PR #3283, opened 2019-09-04
Make scanning all JAX-RS packages explicit

reta:
#### Description
The regression [1] has been reported by one of the CXF users with respect to `Swagger Core` releases after `2.0.6` , here is the issue.

The JAX-RS service has instance of the `Application` class which has only one resource class returned by `getClasses()` method. The service uses Swagger `2.0.8` + OpenAPI v3 and does not specify any resources classes and resource packages to scan (assuming those will be taken from the  `Application` instance). Before `2.0.7` it was the case but since than the `Swagger Core`  changed implementation from `Reflections` to `ClassGraph`, and now **always** scan the complete class path if resource classes and resource packages are empty. The classes from the `Application`  instance are added on top and there is no way to tailor this behavior.

There are couple of solutions to make the behavior more controllable. The one presented in this PR is very simple but intuitive: introduce explicit marker to scan all packages.

```java
openApiConfiguration.setResourcePackages(Collections.singleton("*"));
```

Another option would be to check `Application`  instance for resource classes and allow the scanning of all packages only if this set is empty. 

The feedback would be really appreciated, for now the desired results could be achieved with some smelly workarounds but it would be great to make the scanning outcomes obvious. 

[1] https://issues.apache.org/jira/browse/CXF-8103

Thank you guys. 
CC @frantuma

### DIFF
--- a/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/JaxrsAnnotationScanner.java
+++ b/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/JaxrsAnnotationScanner.java
@@ -18,6 +18,7 @@ import java.util.Map;
 import java.util.Set;
 
 public class JaxrsAnnotationScanner<T extends JaxrsAnnotationScanner<T>> implements JaxrsOpenApiScanner {
+    private static final String ALL_PACKAGES = "*";
 
     static final Set<String> ignored = new HashSet();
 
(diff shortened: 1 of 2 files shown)


frantuma:
Thanks for reporting this! the default behaviour has indeed changed in such a scenario since `2.0.7` (more as a side effect..) and wasn't documented; just merged #3284 introduces [JaxrsApplicationAndResourcePackagesAnnotationScanner](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/JaxrsApplicationAndResourcePackagesAnnotationScanner.java) (see also [updated wiki](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Integration-and-Configuration#resource-scanning)) which behaves as JaxrsApplicationAndAnnotationScanner in versions up to `2.0.6` and can be provided in  [configuration](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Integration-and-Configuration#resource-scanning) (therefore replacing this PR).

In current versions `2.0.6` - `2.0.9`, the same behaviour can be achieved either using `ApplicationScanner` (only considering custom `Application.classes()`), or providing a [custom scanner](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Integration-and-Configuration#resource-scanning).

Let me know if this solves your issue and in this case please close PR.

reta:
Thanks a lot, @frantuma , makes perfect sense! Would you be interested in tests for `JaxrsApplicationAndAnnotationScanner` and `JaxrsAnnotationScanner` (in a separate PR)? I haven't found much tests for scanners, thank you!

frantuma:
tests are more than welcome, thanks!

PR #3284, opened 2019-09-04
refs #3283 - add scanner limited to defined resourcePackages

frantuma:
As mentioned in #3283, since version `2.0.7` default resource scanning behaviour has changed in scenarios where no resource packages or classes are defined; in versions up to `2.0.6` no resources would have been added except the ones defined in custom `Application.classes()`; since `2.0.7` the whole classpath is instead scanned. 

This PR adds a [scanner implementation](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Integration-and-configuration#resource-scanning)  `JaxrsApplicationAndResourcePackagesAnnotationScanner` which behaves as pre `2.0.7` default mechanism


vivekmalhotragithub:
How do we use this new Scanner? Is there an example usage?

frantuma:
@vivekmalhotragithub programmatic usage is e.g. [in this test](https://github.com/swagger-api/swagger-core/pull/3284/files#diff-dee4270ef2b510198f702986648b24daR60), otherwise you can configure it via [configuration](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Integration-and-configuration#resource-scanning) providing full class name `io.swagger.v3.jaxrs2.integration.JaxrsApplicationAndResourcePackagesAnnotationScanner`

PR #3285, opened 2019-09-04
Adding test cases for different JAX-RS scanners

reta:
#### Description 
Adding test cases for different JAX-RS scanners (as per https://github.com/swagger-api/swagger-core/pull/3283 discussions). Thanks guys!

CC @frantuma


frantuma:
Thanks a lot!

PR #3294, opened 2019-09-22
Fix #3124 - Move enums to top level schema definitions

### LINKED ISSUE #3124: ModelResolver: How to ref enum type with enum as schema component?
In the ModelResolver, there is this [piece of code](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L241-L244).

Given an enum property and model class 
```java
public class Test {
    public State state;
    public enum State { UP, DOWN }
}
```

This generates a `string` property with the enum values exploded out.
```json
      "Test" : {
        "type" : "object",
        "properties" : {
          "state" : {
            "type" : "string",
            "enum" : [ "UP", "DOWN" ]
          }
        }
      }
```
---
According to the Enums [documentation](https://swagger.io/docs/specification/data-models/enums/), under the section **Reusable Enums**, it could instead be formatted like so:

```json
      "Test" : {
        "type" : "object",
        "properties" : {
          "state" : {
            "$ref": "#/components/schemas/State"
          }
        }
      },
      "State" : {
        "type" : "string",
        "enum" : [ "UP", "DOWN" ]
      }
```

How does one achieve this?

---
I tried the following variants:

### 1
```java
public class Test {
    public State state;
    
    @Schema
    public enum State { UP, DOWN }
}
```
```json
      "Test" : {
        "type" : "object",
        "properties" : {
          "state" : {
            "type" : "string",
            "enum" : [ "UP", "DOWN" ]
          }
        }
      }
```

### 2
```java
public class Test {
    public State state;

    @Schema(
        type = "string",
        implementation = String.class,
        allowableValues = ["UP", "DOWN"]
    )
    public enum State { UP, DOWN }
}

```
```json
      "Test" : {
        "type" : "object",
        "properties" : {
          "state" : {
            "type" : "string",
            "enum" : [ "UP", "DOWN" ]
          }
        }
      }
```

### 3
```java
public class Test {
    public State state;

    @Schema(
        type = "string",
        implementation = String.class,
        allowableValues = ["UP", "DOWN"],
        ref = "#/components/schemas/State"
    )
    public enum State { UP, DOWN }
}
```
```json
      "Test" : {
        "type" : "object",
        "properties" : {
          "state" : {
            "$ref" : "#/components/schemas/State"
          }
        }
      },
      "State" : {
        "$ref" : "#/components/schemas/State"
      }
```

### 4
```java
public class Test {
    @Schema(ref = "#/components/schemas/State")
    public State state;

    @Schema(
        type = "string",
        implementation = String.class,
        allowableValues = ["UP", "DOWN"]
    )
    public enum State { UP, DOWN }
}
```
```json
      "Test" : {
        "type" : "object",
        "properties" : {
          "state" : {
            "type" : "string"
          }
        }
      }
```

### 5
```java
public class Test {
    @Schema(ref = "#/components/schemas/State")
    public State state;

    @Schema(
        type = "string",
        allowableValues = ["UP", "DOWN"]
    )
    public enum State { UP, DOWN }
}
```
```json
      "Test" : {
        "type" : "object",
        "properties" : {
          "state" : {
            "type" : "string",
            "enum" : [ "UP", "DOWN" ]
          }
        }
      }
```

### 6
```java
public class Test {
    @Schema(ref = "#/components/schemas/State")
    public State state;

    @Schema(
        implementation = String.class,
        allowableValues = ["UP", "DOWN"]
    )
    public enum State { UP, DOWN }
}
```
```json
      "Test" : {
        "type" : "object",
        "properties" : {
          "state" : {
            "type" : "string"
          }
        }
      }
```

mtraynham:
This fixes #3124 by moving enums to top level schema properties, original inspiration from #2996.

@frantuma I know the tests are likely going to break because this changes how enums work.  I was hoping to get your feedback on how to make this configurable, like #2996 originally proposed.  Should it be an annotation on the class or a configuration property of the OpenAPI?

### DIFF
--- a/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/media/Schema.java
+++ b/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/media/Schema.java
@@ -310,6 +310,13 @@ public @interface Schema {
     boolean hidden() default false;
 
     /**
+     * Allows enums to be represented as top level references.
+     *
+     * @return whether or not this schema is a top level enum reference
+     */
+    boolean enumAsRef() default false;
+
+    /**
      * An array of the sub types inheriting from this model.
      */
     Class<?>[] subTypes() default {};
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -263,7 +263,10 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
         }
 
         if (model == null && type.isEnumType()) {
-            model = new StringSchema().$ref(name).name(name);
+            model = new StringSchema();
+            if (resolvedSchemaAnnotation != null && resolvedSchemaAnnotation.enumAsRef()) {
+                model = model.$ref(name).name(name);
+            }
             _addEnumProps(type.getRawClass(), model);
             isPrimitive = true;
         }
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -264,9 +264,6 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
 
         if (model == null && type.isEnumType()) {
             model = new StringSchema();
-            if (resolvedSchemaAnnotation != null && resolvedSchemaAnnotation.enumAsRef()) {
-                model = model.$ref(name).name(name);
-            }
             _addEnumProps(type.getRawClass(), model);
             isPrimitive = true;
         }
@@ -321,6 +318,14 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
                 resolveArraySchema(annotatedType, schema, resolvedArrayAnnotation);
                 schema.setItems(model);
                 return schema;
+            }
+            if (type.isEnumType() &&
+                    resolvedSchemaAnnotation != null &&
+                    resolvedSchemaAnnotation.enumAsRef()) {
+                // Store off the ref and add the enum as a top-level model
+                context.defineModel(name, model, annotatedType, null);
+                // Return the model as a ref only property
+                model = new Schema().$ref(name);
             }
             return model;
         }
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -263,7 +263,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
         }
 
         if (model == null && type.isEnumType()) {
-            model = new StringSchema();
+            model = new StringSchema().$ref(name).name(name);
             _addEnumProps(type.getRawClass(), model);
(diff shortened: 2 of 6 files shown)


frantuma:
@mtraynham thanks for looking into this! I share here some thoughts.

1. I guess the changed code is not sufficient (I also don't understand the change to `Components`) to address the issue; we would need to convert the resolved enum into a ref and add that to the components sections, as done for non primitive types (depending on scenario); also specific tests would need to be added

2. About the way to control behaviour, I can see 2 ways, independent from each other:

A. implement a new annotation e.g `@EnumAsRef`(name = "")` and check it in code for point 1 above.

B. add it as a `Configuration` member; this is more complicated, as atm configuration is not passed down to `ModelConverter` instances (e.g. `ModelResolver`) and related classes; we would probably need to slightly refactor converter related code to gain access to config property; the alternative based on shared static variables sounds more limited and not really consistent, also not applicable e.g. in maven plugin or in scenarios with no java code related to swagger-core config.

mtraynham:
@frantuma heh, I also do not entirely understand why the change in #2996 to `Components` is necessary, but it does seem to be.  I removed it and Swagger instead produces an empty block for enums with no properties.

I'm not against the `@EnumAsRef` or maybe represented as `@Schema(enumAsRef=true)`, my only concern there is consistency throughout a user's API.  But, I also am unfamiliar with how this code works, so it's likely the safer route for me than configuration changes.

xiaoxiangmoe:
@mtraynham What is the current state of this feature? Thanks.

mtraynham:
@xiaoxiangmoe This works, but I haven't really heard back from @frantuma.

mtraynham:
Alright, I think the initial implementation was a bit strange, since it had a `Components` change which nulled the `$ref`.  But after following some of this recursion in `ModelResolver`, I think my latest push is a better implementation.  It simply adds the enum model as a top level schema to the context and returns a simple `Schema` with a `ref`.  I reverted the `Components` change and added a test for this.

mtraynham:
By the way, I have found that if you have an enum that uses a different method of serializing and deserializing with say Jackson, such as an `integer` enum, you can accomplish top level enums in a similar fashion with:

```java
    @Schema(enumAsRef = true, type = "integer", allowableValues = {"1", "2"})
    public enum NumberedEnum {
        NUMBER_1(1),
        NUMBER_2(2);

        private final Integer value;
        NumberedEnum(Integer value) {
            this.value = value;
        }

        @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
        public static NumberedEnum fromValue(Integer value) {
            for (NumberedEnum other : NumberedEnum.values()) {
                if (other.value.equals(value)) {
                    return other;
                }
            }
            throw new IllegalArgumentException("Unsupported value " + value);
        }
    }
```

xiaoxiangmoe:
Is there any feedback on this pr? @frantuma

xiaoxiangmoe:
@frantuma Please have a look. Thanks.

mtraynham:
@frantuma any updates on this?

frantuma:
It's in our radar, we plan to have it in for next release

xiaoxiangmoe:
@frantuma When will the next edition be released?

frantuma:
No exact ETA, but within 2 weeks

xiaoxiangmoe:
@frantuma any updates on this?

frantuma:
Thanks a lot! this is included and replaced by #3352 which adds also global support via sys prop / static var

Issue #3303, opened 2019-09-26
$refs within items don't get added to list of schemas when `isRemovingUnreferencedDefinitions`


ethan92429:
When the schema has an arraySchema that has items that are in turn a $ref, that ref doesn't get followed to add all of its children refs. In fact, I think this happens anytime there are refs of refs. This logic only goes one ref level deep. I was able to work around this in a PR I will post and link to this issue.

ethan92429:
Tried to make the PR through GH's online edit tool, but it got mangled. Trying again now through git.

ethan92429:
OK, I changed the branch to adhere to the contribution guidelines and cleaned up the code. Hopefully it'll pass now!

ethan92429:
Please let me know if I am missing anything in the way of style or logic!

frantuma:
Thanks for reporting this, fixed in #3317 (which includes #3305)

ethan92429:
@frantuma thanks for taking care of this so quickly! And thanks for helping with a great product 😄

PR #3305, opened 2019-09-26
Bug nested refs filtering

ethan92429:
Added logic to keep resolving refs until all children refs are entirely resolved down to their leaves. See here for the discussion: #3303


frantuma:
Thanks! included and replaced by #3317 (adds a test)

PR #3349, opened 2019-11-11
Fix 3348 missing properties

rozzamozza:
Clones all resolved properties so may fix other bugs due to renaming of previously parsed properties


frantuma:
Thanks! included in #3351

PR #3356, opened 2019-11-15
Fix copy-paste issues in AbstractSerializableParameter.getExample 

xpoinsard:
The warning message was speaking of DefaultValue instead of Example.
The warning message was using defaultValue instead of example.
The Boolean conversion was using  defaultValue instead of example.


frantuma:
Thanks!

PR #3415, opened 2020-01-16
fix #3414 - fix enumsAsRef global property behavior

### LINKED ISSUE #3414: ModelResolver.enumsAsRef behavior broken
There is an if-statement in ModelResolver:324:
```java
if (type.isEnumType() &&
                    (resolvedSchemaAnnotation != null && resolvedSchemaAnnotation.enumAsRef()) ||
                    ModelResolver.enumsAsRef
            ) {
```

I believe the way it should be written is:
```java
if (type.isEnumType() &&
                    ((resolvedSchemaAnnotation != null && resolvedSchemaAnnotation.enumAsRef()) ||
                    ModelResolver.enumsAsRef)
            ) {
```

The way it is now seems to attempt to resolve every type as an enum if `ModelResolver.enumsAsRef == true`, resulting in NPEs with `context.defineModel(name, model, annotatedType, null);` passing a `name` that is NULL:
```
java.lang.NullPointerException: null
	at java.base/java.util.TreeMap.put(TreeMap.java:561) ~[na:na]
	at io.swagger.v3.core.converter.ModelConverterContextImpl.defineModel(ModelConverterContextImpl.java:58) ~[swagger-core-2.1.0.jar:2.1.0]
	at io.swagger.v3.core.jackson.ModelResolver.resolve(ModelResolver.java:329) ~[swagger-core-2.1.0.jar:2.1.0]
```

Am I correct or am I misinterpreting the purpose of the `enumsAsRef` property?

daniel-frak:
- Fixes if-statement in ModelResolver to not treat non-enum models as enums when enumsAsRef is globally enabled
- Adds a comment explaining the purpose of the enumsAsRef global property in ModelResolver

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -91,6 +91,10 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
     public static final String SET_PROPERTY_OF_ENUMS_AS_REF = "enums-as-ref";
 
     public static boolean composedModelPropertiesAsSibling = System.getProperty(SET_PROPERTY_OF_COMPOSED_MODEL_AS_SIBLING) != null ? true : false;
+
+    /**
+     * Allows all enums to be resolved as a reference to a scheme added to the components section.
+     */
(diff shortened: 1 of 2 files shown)


frantuma:
Thanks!

Issue #3399, opened 2020-01-03
Change javax to jakarta artifact dependencies


mohanaraosv:
Use the jakarta maven co-ordinates 

https://wiki.eclipse.org/New_Maven_Coordinates

Ref : https://www.eclipse.org/lists/jakartaee-platform-dev/msg00029.html

webron:
That seems only relevant if we update to the latest version, isn't it?

mohanaraosv:
Yes, I will raise PR if you agree on [cxf Jakarta adoption](https://github.com/apache/cxf/pull/500/files)

reta:
In the same vein, it would be probably good to update Bean Validation to 2.0 (1.1 is still used), since there are no Jakarta artifacts for Bean Validation 1.1 released (and very unlikely will be in a future). I am wondering @webron if there are reasons to support Bean Validation 1.1? Thank you!

frantuma:
Thanks for reporting and working on this! implemented in #3467

PR #3422, opened 2020-01-19
Moving to jakarta maven co-ordinates and updated bean validation

mohanaraosv:
Updated maven coordinates for javax and bean validation  [issue](https://github.com/swagger-api/swagger-core/issues/3399)


mohanaraosv:
There a problem with sync I will another PR

PR #3424, opened 2020-01-21
Fix typo in AnnotationsUtils; Fixes #3423

### LINKED ISSUE #3423: Specifying `requiredProperties` on model class yields values in `enum`
I have the following model class:

```
@Schema(requiredProperties = { "recordedBy" })
public class Something {

    @Schema(required = true)
    @JsonProperty
    private Instant recordedAt;

    @Schema(required = true) // this isn't respected for reference values
    @JsonProperty
    private EmployeeHead recordedBy;

    @Schema(required = true)
    @JsonProperty
    private BigDecimal hours;
}
```

which causes the following OpenAPI schema to be generated:
```
    Something:
      required:
      - hours
      - recordedAt
      - recordedBy
      type: object
      properties:
        recordedAt:
          type: string
          format: date-time
        recordedBy:
          $ref: '#/components/schemas/EmployeeHead'
        hours:
          type: number
      enum:
      - recordedBy
```

It appears the `requiredProperties` are treated as possible enum values even though model in question is not an enum.

It should be noted I am using `requiredProperties` in this case since the `required = true` assignment on the schema annotation for `recordedBy` is not being respected (relevant: #3259, #3370). Omitting `requiredProperties` results in `recordedBy` not being listed under `required` in the generated schema. This has been a viable workaround to this date, but the aforementioned behavior breaks it.

jannispl:
Fixes #3423

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
@@ -1696,10 +1696,10 @@ public abstract class AnnotationsUtils {
 
             @Override
             public String[] allowableValues() {
-                if (master.requiredProperties().length > 0 || patch.requiredProperties().length == 0) {
-                    return master.requiredProperties();
+                if (master.allowableValues().length > 0 || patch.allowableValues().length == 0) {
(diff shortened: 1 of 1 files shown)


frantuma:
Thanks!

Issue #3244, opened 2019-07-12
Inheritance getting lost by type re-resolution


UKreher:
Swagger 2.0.8 may generate OpenAPI with lost inheritance information. This occurs with the following type hierarchy:
```
@JsonSubTypes({@Type(B.class)})
class A<T extends X> {
  List<T> list;
}
class B extends A<Y> {}

@JsonSubTypes({@Type(Y.class)})
class X {}
class Y extends X {}
```
When resolving `X` (via `ModelResolver.resolve(AnnotatedType, ModelConverterContext, Iterator<ModelConverter>)`), a model for the subtype `Y` will be defined (`ModelResolver:1219` calling `ModelConverterContext.defineModel(String, Schema, AnnotatedType, String)`) using `allOf`. If `B` gets resolved later, `Y`will get re-resolved. This re-resolution does not respect the extension of `X` by `Y` and thus does not have the `allOf` as it should. Now this wrong model for `Y` is simply defined (`ModelResolver:468` and `ModelResolver:701`) thus overwriting the correct model.
Examining the models already defined in the context is not enough, in my case the model will be defined several times using different contexts. Therefore the model will also be added several times to `Reader.components` with the last model winning.

Unfortunately, the problem depends on the order in which types are resolved. This order probably depends on the order of the methods where these types are used as parameters which seems rather non-deterministic. That is, the resulting model for `Y` may respect the inheritance but after a restart, it may not.

ctytgat:
I've got the exact same problem on swagger 1.5.24. 
We have a quite complex api structure with lots of inheritance. We noticed that the inheritance info was sometimes lost and after intensive debugging came to the same conclusion.
It's basically a lottery ... :(

ctytgat:
FYI, using a custom swagger like this seems to help: (but it's just a hack)

```
public class CustomSwagger extends Swagger {
    public void addDefinition(String key, Model model) {
        if (this.definitions == null || !(definitions.get(key) instanceof ComposedModel)) {
            super.addDefinition(key, model);
        }
    }
}
```

UKreher:
Unfortunately, the hack/my committed changes fail to create the schema in case of the following inheritance:
```
@JsonSubTypes({@Type(T.class), @Type(U.class)})
class S{}
class T extends S {
  List<S> list;
}
class U extends S {}
```
The cyclic dependency needs the re-resolution with a non-ComposedModel. When blocking this, `list` will not be part of the schema.

[...]

PR #3428, opened 2020-01-22
ref #3244 - keep existing ComposedSchema whenever it exists

UKreher:
This PR

-  ignores schemas that overwrite an existing one whereas the existing one is a ComposedSchema and the new one is not
- allows to overwrite a ComposedSchema without a discriminator with a new ComposedSchema with a discriminator
- avoids re-resolution of an existing schema

All changes only avoid overwriting of composed schemas considering inheritance. They do not fix the cause for this behaviour! It would be much better to handle the composed schema before creating a replacement for it.


MichakrawSB:

PR #3429, opened 2020-01-27
Re-ordering Stream APIs

HirotoTanaka114:
We are conducting a research project on re-ordering Stream APIs for improving program execution speed.

Before and after re-ordering, we measured running time of tests which execute target Stream.
Command for running tests is : 
`time mvn -DfailIfNoTests=false -Dtest=CompositionTest test`

and the result is :
```
the order of Stream API      test running time
===============================================
distinct().filter()                9.728s
filter().distinct()                9.710s
```


joschi:
@HirotoTanaka114 Do you have any other benchmarks or research promoting the suggested change?

0.01 seconds don't seem to be very convincing and could as well be an error of measurement or pure chance.

frantuma:
closing as stale and possibly not significant performance improvement

PR #3430, opened 2020-01-28
Updated maven coordinates for javax and bean validation issue

mohanaraosv:
Updated maven coordinates for javax and bean validation [issue](https://github.com/swagger-api/swagger-core/issues/3399)


frantuma:
Thanks a lot! included and replaced by #3467

Issue #3475, opened 2020-03-13
Deterministic ordering of maps in the OpenAPI response.


ethan92429:
I've found that the ordering of the schemas and the paths can appear to be pretty random. It would be nice if these would be always alphabetically ordered... I will try switching all LinkedHashMaps and the like to be TreeMaps and see how it goes in a PR. Unless there is some profound meaning to return everything in the insertionOder? I can't imagine there is, since the read method takes in what could be an unordered set.

ethan92429:
@frantuma here's another PR that should help me fix all the other tests in #3469 by clarifying the failures if there are any - because #3469 seems to switch insertion order by being more efficient in not re-processing types multiple times.

frantuma:
A solution has been implemented in #3740, see [PR comment](#3740) for details and usage scenarios. Closing this ticket, any further issues in this area should be reported in a new or not closed ticket.

PR #3476, opened 2020-03-13
Converts all LinkedHashMaps into TreeMaps to sort paths and schemas a…

ethan92429:
…lphabetically rather than by insertion order. This is to address #3475


ethan92429:
@frantuma this passed with just two test changes to ordering in the JSON. Doesn't change anything but the internal types used from linkedHashMaps to TreeMaps... what do you think?

Issue #3443, opened 2020-02-11
java 8 time: Instant.class is treated as a Long/int64 rather than Data-time 


StephenOTT:
Given: https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java#L248

why is Instant.class treated as a LONG rather than date-time?

frantuma:
Thanks for reporting this! It has been fixed in latest snapshot in #3477 (meaning `java.time.Instant` is resolved as `date-time` by default instead of `integer` / `int64`, as it makes more sense);

In previous versions the behaviour can be customized similarly to what mentioned [here](https://github.com/swagger-api/swagger-core/issues/2992#issuecomment-433369074) and [here](https://github.com/swagger-api/swagger-core/issues/2680#issuecomment-517398335), via e.g.:

```
PrimitiveType.customClasses().put(java.time.Instant.class.getName(), 
                                  PrimitiveType.DATE_TIME);
```

The same mechanism can be applied to resolve e.g. to `long` for any reason in versions `2.1.2+`:

```
PrimitiveType.customClasses().put(java.time.Instant.class.getName(), 
                                  PrimitiveType.LONG);
```

PR #3477, opened 2020-03-16
refs #3443 - Resolve java.time.Instant as date-time instead of long

frantuma:


skel2007:
Could you please port this code (or allow custom mappings) to the 1.6.x? We can't migrate to 2.x for now because it has another API.

PR #3519, opened 2020-04-16
Utilize the OpenApiReader interface in ReaderListener method signatures

### LINKED ISSUE #3517: Change ReaderListener methods to utilize the OpenApiReader interface
I was attempting to extend the [`Reader`](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java) class from the [`swagger-jaxrs2`](https://github.com/swagger-api/swagger-core/tree/master/modules/swagger-jaxrs2) module so I could better control (i.e. customize) the process of reading classes but I ran into this issue with the `ReaderListener` interface. This is a simple change and I may be the only one asking for it. Thus, I'll open a PR soon unless there is an objection.

datagitlies:
overall the purpose of this pull request is to make it easier to extend the Reader class which enables control / customization of the processes within.

Closes #3517


frantuma:
thanks a lot!

Issue #2475, opened 2017-10-04
IntegerSchema doesn't support Long values


frantuma:
Atm IntegerSchema only supports Integer, deserializing or assigning a Long value (e.g. defaultValue or enum) results in error or misbehaviour while casting to integer.

e.g.

```
        schema:
          type: integer
          format: int64
          enum:
          - 2147483647
          - 3147483647
          - 31474836475505055
```

when fixed, also related code in parser should be fixed/adapted, replacing code in :  https://github.com/swagger-api/swagger-parser/pull/536/files

frantuma:
implemented in #2510

PR #3544, opened 2020-05-04
Fix deserializing enum with negative Long values

TomEvers:
Extension to changes made in #2475. 

IntegerSchema supports deserializing or assigning a Long value, but at the moment a small negative Long value isn't casted correctly. This PR adds support for negative Long values as well.


frantuma:
thanks!

Issue #2828, opened 2018-06-01
Serialisation of OpenAPI is not stable


frantuma:
_moved from swagger-api/swagger-parser/issues/728 reported by @jmini_

We have observed a strange behaviour with the Serialisation of an OpenAPI instance with `Yaml.mapper().writeValueAsString(openAPI)` produces different results:

```diff
--- a/src/main/resources/openapi.yaml
+++ b/src/main/resources/openapi.yaml
@@ -97,11 +97,11 @@ paths:
           type: array
           items:
             type: string
-            default: available
             enum:
             - available
             - pending
             - sold
+            default: available
       responses:
         200:
           description: successful operation
@@ -678,20 +678,20 @@ paths:
           type: array
           items:
             type: string
-            default: $
             enum:
             - '>'
             - $
+            default: $
       - name: enum_header_string
         in: header
         description: Header parameter enum test (string)
         schema:
           type: string
-          default: -efg
           enum:
           - _abc
           - -efg
           - (xyz)
+          default: -efg
       - name: enum_query_string_array
         in: query
         description: Query parameter enum test (string array)
@@ -701,20 +701,20 @@ paths:
           type: array
           items:
             type: string
-            default: $
             enum:
             - '>'
             - $
+            default: $
       - name: enum_query_string
         in: query
         description: Query parameter enum test (string)
         schema:
           type: string
-          default: -efg
           enum:
           - _abc
           - -efg
           - (xyz)
+          default: -efg
       - name: enum_query_integer
         in: query
         description: Query parameter enum test (double)
@@ -743,18 +743,18 @@ paths:
                   description: Form parameter enum test (string array)
                   items:
                     type: string
-                    default: $
                     enum:
                     - '>'
                     - $
+                    default: $
                 enum_form_string:
                   type: string
                   description: Form parameter enum test (string)
-                  default: -efg
                   enum:
                   - _abc
                   - -efg
                   - (xyz)
+                  default: -efg
       responses:
         400:
           description: Invalid request
@@ -1214,11 +1214,11 @@ components:
         name: Name
     EnumClass:
       type: string
-      default: -efg
       enum:
       - _abc
       - -efg
       - (xyz)
+      default: -efg
     List:
       type: object
       properties:
```

From a semantic point of view, the content is the same. But my expectation is to get the content always serialised the same way.

I did not understood what is producing the difference.
Maybe it depends from the JVM. 

[...]

Issue #2775, opened 2018-04-18
Swagger file rendering - Elements sorted alphabetically 


bibiwan:
I need to render swagger files using POJO classes which reads description of an API from CSV files, and I use your API to get this done.

CSV files describes elements from a business point of view and are **not** sorted alphbetically.

Example : 
Property 1 :
  sub-property B
  sub-property C
  sub-property A

When I use your lib to generate swagger, here is what I get : 
Property 1 :
  sub-property A
  sub-property B
  sub-property C

As you can see, sub-properties are sorted.

Here is my method's code to get yaml string :
`return Yaml.pretty().writeValueAsString(factory);`
where factory is a io.swagger.models.Swagger instance.

Would it be possible not to sort the properties when rendering Yaml ?

webron:
Where does the CSV come to play here? The use case is a bit unclear.

bibiwan:
I want to get an api description in swagger format for Oracle's Policy
Modeling projects, that's why I use CSV files : they describe my api and
data I need to use to communicate with webservices designed with Oracle
product. So, I create a swagger object with your api, and populate this
object with informations stored inside my CSV files to create operations,
properties,... It's working perfectly well, except for this sort issue. So
I'd like to know if there's a way to disable the alphabetical sort of the
yaml output, or if this feature could be planned into the future releases.

bibiwan:
Hi,

I'm back about my request. Could we just forget the CSV part ? Let's focus about this : is it possible that elements added to a Swagger instance are not sorted alphabetically when calling the method `Yaml.pretty().writeValueAsString(factory);` ?

MichakrawSB:
Closing this issue as part of our https://github.com/swagger-api/swagger-core/issues/5074.
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.
Thank you for your understanding!

Issue #2400, opened 2017-08-25
Swagger-Models: Make definitions use a sorted Map instead of HashMap


domdorn:
io.swagger.models.Swagger#definitions is using a HashMap as its implementation. This leads to unstable results when serializing the spec to json. 

I'd like to change the implementation to use a TreeMap<String, Model> instead. This makes sure, that the map is always sorted the same way. 

I'm happy to provide a pull request.

ChristianLutz:
Currently this [PR](https://github.com/swagger-api/swagger-core/pull/2968) has been applied to oas3. It uses an `LinkedHashMap` instead of a `TreeMap`. I would also prefer the later.

hiddewie:
The maven plugin from OpenAPI tools has to sort the output manually now. This is a problem for projects where the OpenAPI specification is committed. Also see openapi-tools/swagger-maven-plugin#49.

PR #3549, opened 2020-05-06
ticket-3475

fkardashov:
this will avoid randomization in the output file


joschi:
Refs #3475
Refs #2828

This PR looks similar to #3476.
While it reduces the entropy in the generated specification, it doesn't cover all cases.
For example elements in a list (such as enums) aren't sorted and will still depend on various other factors.

hiddewie:
Also ref:
- #2775
- #3266
- #2400

frantuma:
A solution has been implemented in #3740, see [PR comment](#3740) for details and usage scenarios. Closing this ticket, any further issues in this area should be reported in a new or not closed ticket.

PR #3551, opened 2020-05-07
feature: listing registered converters

ymajoros:
Sometimes, it's useful to be able to list registered converters to know if we need to add some converter. For example, we needed that for the springdoc-openapi project to allow tests running in parallel (avoiding them to all add the same converters when different spring contexts are present).


frantuma:
thanks!

PR #3572, opened 2020-05-29
Added support for filterClass to OpenApiServlet

espenaf:
This adds support for the flterClass option in OpenApiServlet. Making the servlet behave more like the BaseOpenApiResource.


modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
```diff
@@ -80,4 +106,43 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+        if (StringUtils.isEmpty(queryString)) {
+            return queryParameters;
+        }
+
+        String[] parameters = queryString.split("&");
+
+        for (String parameter : parameters) {
+            String[] keyValuePair = parameter.split("=");
+            queryParameters.put(keyValuePair[0], Collections.singletonList(keyValuePair[1]));
+        }
```
joschi, line 122: This would overwrite query parameters which occur multiple times.

For example with `?param1=value1&param2=value2&param1=value3`, the `queryParameters` map would only contain `{param1=value3, param2=value2}`.

Consider using [`MultivaluedHashMap`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/MultivaluedHashMap.html) instead.

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
```diff
@@ -80,4 +106,43 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+    }
+
+    private static Map<String, String> getCookies(Cookie[] cookies) {
+        Map<String, String> mapOfCookies = new HashMap<>();
+        if (cookies != null) {
+            for (Cookie cookie : cookies) {
+                mapOfCookies.put(cookie.getName(), cookie.getValue());
+            }
+        }
+        return mapOfCookies;
```
joschi, line 133: There can be multiple cookies with the same name but we would only return the last cookie for any name in the map.

Consider using [`MultivaluedHashMap`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/MultivaluedHashMap.html) instead.

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
```diff
@@ -80,4 +106,43 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+    }
+
+    public static Map<String, List<String>> getHeaders(HttpServletRequest req) {
+        if (req.getHeaderNames() == null) {
+            return new HashMap<>();
```
joschi, line 138: Should we return `Collections.emptyMap()` here?

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
espenaf: Good point. Fixed.

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
espenaf: I agree, but the cookie parameter for io.swagger.v3.core.filter.SpecFilter#filter() and the underlying abstract and interface (OpenAPISpecFilter) only takes in an Map<String, String>, so a MultivaluedHashMap won't work. Skimming through these classes, it seems the cookies aren't used for anything yet, I can't find an filterCookies() in SpecFilter so that may be the reason it is only taking in an Map<String, String>, instead of and Map<String, List<String>>. Do you want be to refactor the whole thing, or just leave it as it is. Could create another PR for this refactoring covering both OpenApiServlet and BaseOpenApiResource.

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
espenaf: I agree.

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
joschi: Fair enough, let's keep the "flat" map for now.

@espenaf Could you please create an issue for refactoring the Cookie handling accordingly? 😃

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
```diff
@@ -80,4 +107,43 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+
+        if (StringUtils.isEmpty(queryString)) {
+            return queryParameters;
+        }
+
+        String[] parameters = queryString.split("&");
+
+        for (String parameter : parameters) {
+            String[] keyValuePair = parameter.split("=");
+            queryParameters.put(keyValuePair[0], Collections.singletonList(keyValuePair[1]));
```
joschi, line 122: Since using `MultivaluedHashMap`, we only need to put the flat value instead of a list with the value here:
```suggestion
            queryParameters.put(keyValuePair[0], keyValuePair[1]);
```

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
```diff
@@ -80,4 +107,43 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+        MultivaluedHashMap<String, String> queryParameters = new MultivaluedHashMap<>();
+
+        if (StringUtils.isEmpty(queryString)) {
+            return queryParameters;
+        }
+
+        String[] parameters = queryString.split("&");
+
+        for (String parameter : parameters) {
+            String[] keyValuePair = parameter.split("=");
```
joschi, line 121: Query parameters could only consist of a key (`https://example.com/path?my-parameter`) but we're only handling key/value pairs here.

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
espenaf: Issue created in https://github.com/swagger-api/swagger-core/issues/3582

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/integration/OpenApiServlet.java
espenaf: The MultivaluedHashMap only accepts an List<String> as value, to flattening it won't work. Digging deeper into how this is handeled in BaseOpenApiResource is see that we really would want a proper request query handler with URI Decoder like the one found in org.glassfish.jersey.uri.UriComponent#decodeQuery() and used by UriInfo that BaseOpenApiResource pulls inn. This seem to require adding an dependency to jersey-common. Is that OK, or should be just try to handle it all here.

joschi: > The MultivaluedHashMap only accepts an List as value, to flattening it won't work.

See [`MultivaluedHashMapl#add(K, V)`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/AbstractMultivaluedMap.html#add-K-V-).

My suggested change was incorrect and was still using `put()` instead of `add()`. 😄

joschi: > This seem to require adding an dependency to jersey-common. Is that OK, or should be just try to handle it all here.

Given that we want to support all valid JAX-RS implementations and not only Jersey, I don't think this is an option.

espenaf: Turns out the getParmeterMap() method from the request gives me an map with all keys and values, but the values are String[]. So I have rewritten the method, added URI Decoding support, moved it into it's own class and added some tests.

frantuma:
@joschi @espenaf thanks a lot, does the PR look ok to both?

joschi:
@frantuma I think it's good to merge.

espenaf:
I am happy with it, so just merge.

Issue #3554, opened 2020-05-07
update jackson to latest 2.11.1+


frantuma:
see 
* https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.11
* https://medium.com/@cowtowncoder/jackson-2-11-features-40cdc1d2bdf3
* https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.11.1

Xexanos:
Are there any news on a timeline for this update?

frantuma:
no fixed ETA, probably within next release; things should work probably fine though with 2.11.1+ if you define version in your POM / build system

nhenneaux:
I have tested using jackson 2.11.0 on our project and it was failing since a deprecated method was removed.
```
java.lang.NoSuchMethodError: com.fasterxml.jackson.databind.introspect.AnnotatedMember.getType(Lcom/fasterxml/jackson/databind/type/TypeBindings;)Lcom/fasterxml/jackson/databind/JavaType;
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:424)
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:202)

**Comment by qwazer** [comment]
The same issue when use 
```
        <dependency>
            <groupId>io.swagger</groupId>
            <artifactId>swagger-annotations</artifactId>
            <version>1.6.1</version>
        </dependency>
```
and 
```
  <parent>
    <groupId>com.fasterxml.jackson</groupId>
    <artifactId>jackson-base</artifactId>
    <version>2.11.0</version>
  </parent>
```


```
Caused by: java.lang.NoSuchMethodError: 'com.fasterxml.jackson.databind.JavaType com.fasterxml.jackson.databind.introspect.AnnotatedMember.getType(com.fasterxml.jackson.databind.type.TypeBindings)'
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:425) ~[swagger-core-1.6.1.jar:1.6.1]
	at io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:203) ~[swagger-core-1.6.1.jar:1.6.1]

```
Description:
An attempt was made to call a method that does not exist. The attempt was made from the following location:
    io.swagger.jackson.ModelResolver.resolve(ModelResolver.java:425)
The following method did not exist:
[...]

PR #3576, opened 2020-06-03
Update to Jackson 2.11.0

joschi:
* Update to Jackson 2.11.0
* Make tests independent of YAML formatting (whitespace, single/double quotes)

Refs #3554
Closes #3531


frantuma:
not sure here, 2 thoughts:

1. the change introduces a "more relaxed" check which would e.g not cover changes like the ones discussed in #3544; while this is already true for many tests where `compareAsYaml` or equivalent is used, the ones doing string comparison still provide final output comparison.

2. still I'd wait for outcome of https://github.com/FasterXML/jackson-dataformats-text/pull/201, to possibly switch directly to `2.11.1`

joschi:
With FasterXML/jackson-dataformats-text#201 merged and scheduled for Jackson 2.11.1, I would also close this PR and upgrade directly to Jackson 2.11.1 once it has been released. 👍

PR #3578, opened 2020-06-04
Take `@Deprecated` into account for operations

### LINKED ISSUE #3500: swagger-jaxrs2 (2.1.2) scanner ignores "@Deprecated"
The following request will not be marked as deprecated in the openapi.json
```
@Produces(MediaType.APPLICATION_JSON)
@Consumes(MediaType.APPLICATION_JSON)
@Api("Some API")
@Path("2.0/api")
@Local
public interface PersonResource
{
    @GET
    @Path("/someRequest")
    @Deprecated
    Response someRequest();
}
```
But according to the javadoc it should be marked as deprecated:
http://docs.swagger.io/swagger-core/v2.1.2/apidocs/io/swagger/v3/oas/annotations/Operation.html#deprecated--

> Allows an operation to be marked as deprecated. Alternatively use the `@Deprecated` annotation

In the old 1.5 swagger-jaxrs it was supported too.

joschi:
Mark operations as deprecated which have been declared on a deprecated method or class using the `java.lang.Deprecated` annotation.

Fixes #3500

### DIFF
--- a/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
+++ b/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
@@ -954,7 +954,10 @@ public class Reader implements OpenApiReader {
         }
 
         if (apiOperation != null) {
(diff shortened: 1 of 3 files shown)


frantuma:
thanks! included and replaced by #3590

PR #3588, opened 2020-06-10
Bug/3587 fix parameter examples order

### LINKED ISSUE #3587: Parameter examples order is lost during processing
When working with multiple examples it is useful to control the order that the parameter examples are defined so that when they are displayed the first example can be setup to be a reasonable happy path for the api.

Specifically I'm using Swagger UI to render generated OpenAPI documentation from a CXF/JaxRs service definition in a Spring Boot application.  For one of the optional query parameters I've defined a collection of examples including a blank one.  Ideally the blank one should be first (in this particular case) so that when you go to the Swagger UI it will be the first example and there for the default when you click 'Try it Out'.  That way you don't have to twiddle with each parameter example for the nominal/default case.

Parameter annotation example:
```
 @Parameter(
  examples = {
    @ExampleObject(
      summary = "All Fields",
      name = "All Fields - the query parameter is blank or not set"
    ),
    @ExampleObject(
      summary = "ID Only",
      name = "ID Only - ...",
      value = "id"
    ),
    @ExampleObject(
      summary = "Unrecognized Field",
      name = "Unrecognized Field - ...",
      value = "fieldThatDoesNotExist"
    )
  }
)
```

Because the ordering is lost, this is the resulting OpenApi document snippet (which Swagger UI dutifully renders in this order):
```
 "examples" : {
  "ID Only - ..." : {
    "summary" : "ID Only",
    "description" : "ID Only - ...",
    "value" : "id"
  },
  "Unrecognized Field - ..." : {
    "summary" : "Unrecognized Field",
    "description" : "Unrecognized Field - ...",
    "value" : "fieldThatDoesNotExist"
  },
  "All Fields - the query parameter is blank or not set" : {
    "summary" : "All Fields",
    "description" : "All Fields - the query parameter is blank or not set"
  }
}
```

I think the fix is as simple as modifying https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/ParameterProcessor.java#L110 to use a LinkedHashMap instead of a HashMap for exampleMap, but I haven't investigated beyond that point to see if the order is also mangled further on.

Does that seem like an appropriate fix?

mandragorn:
fixes #3587 
This pull request ensures that the order that examples are defined in the @Parameter.examples property is the same as the order in the generated openapi document.


mandragorn:
Will add tests before marking it ready for review.

mandragorn:
@joschi I do not have permission to merge this.  What is the process for getting it merged and where do I find out when the next release will be?

joschi:
@mandragorn No worries, @frantuma or another maintainer of `swagger-core` will merge the PR when they have time for it.

mandragorn:
:+1: awesome.  Thank you!

frantuma:
Thanks a lot!

PR #3592, opened 2020-06-15
Fixes issue with multiple cookie values with same name when used by an filterClass. 

### LINKED ISSUE #3582: OpenAPISpecFilter does not handle cookies correctly
The OpenAPISpecFilter.filter* methods and underlying abstract an implementations takes inn an Map<String, String> for cookies, while headers and parameters are defined as Map<String, List<String>>. Cookies might have multiple values for the same name, making this structure only providing the last cookie extracted from the incoming request. Fortunaltly, for the moment there is no filterCookie operation in the SpecFilter class, so the values are not used yet. My suggestion is that this structure is changed to an Map<String, List<String>>. Se https://github.com/swagger-api/swagger-core/pull/3572 for some reference into discovering the issue.

espenaf:
All methods in the OpenAPISpecFilter interface have a Map of cookies, but this map is only <String, String>. This PR refactors it into an Map<String, List<String>>, in the same way as params and headers are created. Because we can have cookies with the same name and multiple values.  This should fix #3582.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/filter/AbstractSpecFilter.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/filter/AbstractSpecFilter.java
@@ -16,42 +16,42 @@ import java.util.Optional;
 public abstract class AbstractSpecFilter implements OpenAPISpecFilter {
 
     @Override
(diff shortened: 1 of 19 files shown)


MichakrawSB:

PR #3594, opened 2020-06-19
add dependabot.yml for dependency management

sullis:
Dependabot version 2:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/


sullis:
LGTM

sullis:
@frantuma

sullis:
WDYT? @frantuma

frantuma:
Sorry about late response, a quick update: we are considering this, but still have doubts about real added value, as tests cannot cover 100% of possible regressions, and careful checks must be applied; this means that PRs will be little more than a "notification" of update existence and not evident breaking, and this could be possibly not so useful noise.

But this doesn't mean we will not apply it, we have it in sight and keep updates here.

PR #3613, opened 2020-07-02
Deterministic and alphabetical orderding of JSON and YAML output

hiddewie:
References:

- https://github.com/OpenAPITools/openapi-generator/pull/3763
- https://github.com/swagger-api/swagger-core/issues/2828
- https://github.com/swagger-api/swagger-parser/issues/728
- https://github.com/OpenAPITools/openapi-generator/pull/233
- https://github.com/swagger-api/swagger-core/pull/3549
- https://github.com/swagger-api/swagger-core/pull/3476
- https://github.com/swagger-api/swagger-core/issues/3475
- https://github.com/openapi-tools/swagger-maven-plugin/pull/49
- #2775
- #3266
- #2400

### Problem

When using the Swagger library to generate OpenAPI specifications from source code, the output should be fully deterministic. Running the tool on the same source code should always give the same result. This is not the case at the moment.

Because the output is non-deterministic, the contents will change after running the tool again. This causes problems when the output is committed to a version control system such as Git. Even when not changing the input, a huge diff may be generated because the content of the file has changed.

### This PR

This PR contains two parts:
- There were two parts of the core library that were non deterministic, the reflection access of fields and of methods. These instances in the code have been sorted for determinism.
- Many parts of the OpenAPI specification are stored in `Map`s, specifically unordered `Map`s. This causes problems because the insertion order is ususally the iteration order. This problem has been solved by having the Object mapper (Jackson) sort the output properties.


a-abella:
please build

swarth100:
I've also been looking into deterministic serialization of the OpenAPI spec and I believe the above changes might not be enough.

Individual classes (such as `Schema`) could still serialize their fields in a different order.

My current workaround is to acquire Swagger's object mapper and set property ordering myself:

```
Json.mapper().addMixIn(Paths.class, AlphabeticallySorted.class);

...

@JsonPropertyOrder(alphabetic = true)
public abstract static class AlphabeticallySorted {}
```

hiddewie:
@swarth100 That also seems like a good (additional) solution. Could you make a test case with a specification which fails on non-determinism for this PR?

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
```diff
@@ -372,8 +375,13 @@ public OpenAPI read(Class<?> cls,
         // look for field-level annotated properties
         globalParameters.addAll(ReaderUtils.collectFieldParameters(cls, components, classConsumes, null));
 
+        // Make sure that the class methods are sorted for deterministic order
+        // See https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getMethods--
+        final List<Method> methods = Arrays.stream(cls.getMethods())
+                .sorted(Comparator.comparing(Method::getName))
```
jeff-lauterbach-by, line 381: According to the linked docs, `Class.getMethods()` does not return methods in any particular order and multiple methods can have the same name... Should some additional properties be added to the comparator?

hiddewie: Good point. I will add a custom comparator that also sorts based on the arguments of the method.

mariantirlea:
Hello, I am looking forward to see an updated on this. This feature will really help us because we need a test where the current OpenApi file is compared to a reference file to see if there are any differences (this way we avoid unwanted mistakes in the API)

frantuma:
included and replaced by #3740, see [PR comment](#3740) for details and usage scenarios

PR #3553, opened 2020-05-07
Support for overriding enum values with @JsonProperty or @JsonValue

joschi:
This PR adds support for overriding enum values with the Jackson-specific annotations [`@JsonProperty`](https://fasterxml.github.io/jackson-annotations/javadoc/2.10/com/fasterxml/jackson/annotation/JsonProperty.html) or [`@JsonValue`](https://fasterxml.github.io/jackson-annotations/javadoc/2.10/com/fasterxml/jackson/annotation/JsonValue.html).

Those annotations allow customizing how enum values are serialized and deserialized by Jackson and this PR consolidates the generated OpenAPI specification with this customization.

Java example enums (both serializing and deserializing identically when using Jackson):

```java
public enum SomeEnum {
  @JsonProperty("num-one")
  NUMONE,
	
  @JsonProperty("num-two")
  NUMTWO
}
```

```java
public enum SomeEnum {
  NUMONE("num-one"), NUMTWO("num-two");

  private final String value;

  public SomeEnum(String value) {
    this.value = value;
  }

  @JsonValue
  public String getValue() {
    return value;
  }
}
```

Results in the generated OpenAPI specification:
```yaml
someEnum:
  type: string
  enum:
    - num-one
    - num-two
```

Closes #3552

PR #3617, opened 2020-07-05
Backport from master to 1.5 for support to override enum values

### LINKED ISSUE #3616: Support overriding enum values for branch 1.5
I am using the 1.x.x version of the swagger-maven-plugin to generate OpenAPI v2 specification from my annotated model classes. Unfortunately, switching to OpenAPI v3 is not an option at the moment.

I need to generate enum values that have spaces or other special characters. This is currently not possible with the 1.x.x version. The generator always picks the name of the enum from the Java class, which has limits from the java language syntax. The latest release of the 2.x.x. swagger-api had an enhancement for that problem (issue #3552). This enhancement allows to annotate the enum values with `@JsonProperty` or a getter method with `@JsonValue`.

It would be nice, if this feature would also be available when generating OpenAPI v2 specs with the 1.x.x version.

hhat42:
This PR adds the feature from #3553 to the 1.5 branch. It allows to override enum values with the Jackson-specific annotations @JsonProperty or @JsonValue.

Since the 1.5 branch uses source target setting for Java 7, the `stream` API and `Optional` are not available. The code from the master branch was adapted to work without these. Except from that, the changes from the #3553 PR have all been ported into this PR, including the corresponding tests.

Closes #3616

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
@@ -14,7 +14,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 
 import javax.validation.constraints.DecimalMax;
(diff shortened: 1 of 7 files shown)


frantuma:
Thanks!

PR #3637, opened 2020-07-16
implements exampleSetFlag to flag deserialization of null example vs missing

frantuma:


cvgaviao:
@frantuma, I think I'm getting a collateral effect of this change in the Parser. When serializing the OpenAPI object, the ` exampleSetFlag` is being written in my YAML file.
```
  parameters:
    ParamP-App-id:
      name: app-id
      in: path
      required: true
      style: simple
      explode: false
      schema:
        type: integer
        format: int32
        exampleSetFlag: false
```

frantuma:
@cvgaviao 

it's possible that you're not using `swagger-core` [Json](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/Json.java) or [Yaml](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/Yaml.java) objects; these guys basically wrap a customized `ObjectMapper`, among the other things registering [SchemaMixin](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/mixin/SchemaMixin.java#L24) which is excluding `exampleSetFlag` from serialization.

You would use them e.g by calling `Yaml.pretty(openAPI)`.

cvgaviao:
> @cvgaviao
> 
> it's possible that you're not using `swagger-core` [Json](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/Json.java) or [Yaml](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/Yaml.java) objects; these guys basically wrap a customized `ObjectMapper`, among the other things registering [SchemaMixin](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/mixin/SchemaMixin.java#L24) which is excluding `exampleSetFlag` from serialization.
> 
> You would use them e.g by calling `Yaml.pretty(openAPI)`.

@frantuma, well I'm using exactly this with openapi-generator and swagger-parser 2.0.20:
```
      String openapiYaml = Yaml.pretty(openAPI);
      Files.write(outputFile.resolve(outputFile),
                  openapiYaml.getBytes(StandardCharsets.UTF_8));
```
And it being working well, but after moving to the 2.0.22-SNAPSHOT the problem appears.

frantuma:
@cvgaviao I am not able to reproduce this, it successfully tests in core and latest parser.

Possibly the ObjectMapper used gets customized somehow, disabling `MixIns` or something.

If you're still experiencing the issues, please provide a reproducible test case

cvgaviao:
@frantuma, I'm in a rush here and I downgraded the version in order to keep things running ok.

I was not able to find any Serialization test in this code, just found Deserialization ones. Could you point it to me?  I'll try to look at it, and if were necessary to create a reproducer this weekend.

frantuma:
@cvgaviao e.g. [this test](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/test/java/io/swagger/v3/core/deserialization/JsonDeserializationTest.java#L376)

bdzzaid:
@casualjim I produce the same issue. I downgrade my swagger-core version to 2.1.3 to resolve my issue.  

My point of view can be wrong, but I think the issue is in this class : 
modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/Schema.java 

When the boolean exampleSetFlag is a primitive field and not a Boolean object, the ObjectMapper will always serialize the boolean value (false if not set, true if the method public void setExample(Object example) is called.

Issue #3643, opened 2020-07-20
[Jaxrs2] Cannot Define Serialisation Rules For Form UrlEncoded Body


BenjaminHetzjelli:
According to the spec, one should be able to define serialisation rules for application/x-www-form-urlencoded body parameters. In practice though, it is not possible at this time through either using @FormParams individually or by combining them into an @BeanParam due to the way they are converted into the request body.

In practice, this means that if one's API does not take parameters in the default way, then requests generated using the spec will not work. This is especially egregious since it seems swagger-js (used by swagger-ui) doesn't serialise form params properly by default (at least when compared to [https://swagger.io/docs/specification/describing-request-body/](this documentation).

frantuma:
Thanks for reporting this! Should be addressed by #3696, closing ticket please reopen if still experiencing issues (see also comment in [this question](https://community.smartbear.com/t5/Swagger-Open-Source-Tools/Specifying-Encoding-For-Form-Data-Parameters/m-p/206091#M1362))

Issue #3282, opened 2019-08-26
ParameterIn.formData / @FormParam?


wickedsp1d3r:
Some example method:
```
@POST
@Path("/testing")
@Consumes(MediaType.APPLICATION_FORM_URLENCODED)
public Response myMethod(
  @Parameter(in = "formData", description = "Testing")
  @FormParam("data") final String data) {
  ...
}
```
Problem:

- `@Parameter` "in" does not allow arbitrary strings
- Not able to use `@FormDataParam` (not available)
- When `@Parameter(in = ParameterIn.X)` is not defined
    - Parameter is resolved as `@RequestBody` (because of `@POST`)
    - For some reason, descriptions and required indicators are not shown in Swagger UI

How do I manage to show the `@FormParam` as formData parameter in Swagger UI?
Thank you in advance.

pioto:
I can confirm this issue.

A combination of `@Consumes(MediaType.APPLICATION_FORM_URLENCODED)` on the method, and `@FormParam("data")` on the parameter, does get Swagger UI to mostly generate thr right sort of thing, but if I add `@Parameter` to that parameter, my description does not appear, and doesn't seem to be copied over in the way that would be expected from https://swagger.io/specification/#support-for-x-www-form-urlencoded-request-bodies

AndrewBoyarsky:
I have the same bug, @Parameter + @FormParam does not show description, provided in the @Parameter

frantuma:
partially addressed by #3696

MichakrawSB:
Closing this issue as part of our https://github.com/swagger-api/swagger-core/issues/5074.
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.
Thank you for your understanding!

PR #3644, opened 2020-07-20
Encoding form parameters via RequestBody on Method Level

BenjaminHetzjelli:
Maybe addresses part of #3282
Definitely addresses #3643

Allow setting encoding for form parameters via the `@RequestBody` annotation on method level. It could be more natural for users to allow using `@Parameter` on each `@FormParam`, but this way introduced a lot less complication to the code, esp in handling the params correctly both wrapped into and unwrapped from an `@BeanParam`.

I've also added a couple tests to make sure it is read properly.


frantuma:
@BenjaminHetzjelli thanks! this is replaced by #3696, see comment there and in original question

BenjaminHetzjelli:
Thanks so much! The solution you've implemented seems great!

PR #3717, opened 2020-10-03
allow override of built-in jackson mappers

pjfanning:
I have a use case where I would like to create an ObjectMapper with a larger type cache (a jackson setting). If I can override the mapper in Json util class, then this would faciltate this.


pjfanning:
@frantuma would it be possible to consider merging this?

pjfanning:
This approach may not be a good idea given the way ObjectMapperFactory already sets a number of its own settings on the Jackson mappers.

pjfanning:
replaced by https://github.com/swagger-api/swagger-core/pull/4053

Issue #3695, opened 2020-09-09
Is it possible to order(sort) request methods (POST,GET,PUT,DELETE) in Swagger-UI using swagger-jaxrs2 with Sring boot + Jersey controller API?


tGuichin:
Currently the swagger-jaxrs2 generated Swagger-UI has no order for each "GET", "PUT", "POST"... etc method.  It randomly spits the endpoints out as it pleases.  Is there a way to sort them?  I tried using ```@JsonPropertyOrder({"POST","GET","PUT","DELETE"})``` , which is the order I would like them in, but to no success.  Also tried ```@JsonPropertyOrder(alphabetic = true)``` .  This annotation was included in the Jersey Controller class. 

I'm using:

```
<dependency>
	<groupId>io.swagger.core.v3</groupId>
	<artifactId>swagger-jaxrs2</artifactId>
	<version>2.1.4</version>
</dependency>
```

swarth100:
I think you need to configure swagger core's Json object mapper

Does this work:

Declare this anywhere
```
/* Mixin to ensure custom ordering of certain properties */
@JsonPropertyOrder(alphabetic = true)
public abstract static class AlphabeticallySorted {}
```

Configure the Json object mapper for API Operation class ordering
```
import io.swagger.v3.core.util.Json;
import io.swagger.v3.oas.models.Operation;

...

Json.mapper().addMixIn(Operation.class, AlphabeticallySorted.class);

```

This way you can configure the ordering of any part of the OAS Schema

Also if you wish for keys in unordered maps to be sorted:

```
Json.mapper().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);
```

tGuichin:
I'm not fully sure how to implement what you mentioned, but I will play around with it to see if I can figure it out when I get some time.  Thank you, @swarth100 .

tGuichin:
@swarth100 So your solution worked, somewhat.  I only added ```Json.mapper().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);`` and it sorted the Swagger-UI endpoints by path (/user/name/...) instead of by request method(GET/PUT/POST/DELETE).  The Json.mapper() piece have no affect on the sorting.  Any ideas on if it's possible to sort by method?  I tried ```Json.mapper().addMixIn(HttpMethod.class, AlphabeticallySorted.class);``` with no luck.   Thanks in advance.

swarth100:
[...]

PR #3740, opened 2020-12-03
refs #3613 - Configurable and deterministic order of JSON and YAML output

frantuma:
Includes and replaces #3613, provides a solution for related issues, including:

- #2400
- #2775
- #2828
- #3266
- #3475
- #3476
- #3549
- #3695
- https://github.com/swagger-api/swagger-parser/issues/728

The solution covers two areas:

1. deterministic input (reflection): methods and fields are sorted consistently, thanks @hiddewie!

2. sorted serialization.

This is achieved in two possible ways:

A. specifying the newly introduced configuration option (and maven/gradle parameter) `sortOutput`, which applies an opinionated sorting , using Jackson `ORDER_MAP_ENTRIES_BY_KEYS` and `SORT_PROPERTIES_ALPHABETICALLY` along with defined order for serialization of `OpenAPI` root and `Schema `instances

B. fully customizing the `ObjectMapper` used in serialization of output, using the existing mechanism of the `ObjectMapperProcessor` which is now defining 2 more methods `processOutputJsonObjectMapper` and `processOutputYamlObjectMapper` to customize the "output mapper" (which is different from the one used while resolving, customizable with the existing method `processJsonObjectMapper`).

Please see tests for an example of usage.

As quick background, many alternatives have been proposed to obtain a deterministic output, including but not limited to usage of sorted Maps and Sets at the POJO level, hard-coding Jackson sort features, providing specific serializers etc. but all options would have affected the deserialization->serialization roundtrip not allowing to maintain the original order. Additionally the order of output fields is and should be opinionated, as the semantic meaning stays the same.

Related tickets will be closed referencing this PR, any further issues in this area should be reported in a new or not closed ticket.


hiddewie:
Thank you @frantuma

PR #3843, opened 2020-12-18
Migrate from maven-bundle-plugin to bnd-maven-plugin for finer OSGi meta…

### LINKED ISSUE #3842: Migrate from maven-bundle-plugin to bnd-maven-plugin for finner OSGi metadata
I'd like to fix a few small issues with the OSGi metadata (as I did in a previous issue) however due to a very old maven-bundle-plugin and for much finer control of the OSGi metadata I'd like to pitch migrating to bnd-maven-plugin (I've got a PR which does this coming right after this.)

rotty3000:
…data

fixes #3842

Signed-off-by: Raymond Augé <raymond.auge@liferay.com>

### DIFF
--- a/modules/swagger-annotations/pom.xml
+++ b/modules/swagger-annotations/pom.xml
@@ -8,7 +8,6 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>swagger-annotations</artifactId>
-    <packaging>bundle</packaging>
     <name>swagger-annotations</name>
     <description>swagger-annotations</description>
     <build>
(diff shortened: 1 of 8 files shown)


rotty3000:
fix unfortunate commit text...

frantuma (commented):
@rotty3000 thanks a lot for your effort in this area. I have looked at your changes and impact on generated artifacts, and there are some possible open issues:

1. The main one: if I am not mistaken the generated JARs manifests do not include anymore an `Automatic-Module-Name` entry, even if this seems to be configured in the new plugin. This would need to be fixed to consistently provide automatic module name to JDK 9+ modules based envs. 

2. There are differences in the outcome related to `Export-Package`, `Import-Package` and possibly `Require-Capability` for all modules; This is probably ok, and even "better" than what previously provided but I cannot at the moment do thorough tests in OSGi env for all modules to verify their correct behaviour. I guess you are more familiar and up to date with OSGi internals, and actively using `swagger-core` in OSGi, I would therefore ask you to confirm that all packages have been tested in OSGi env and that they correctly define imports and exports.

Thanks again for your effort, please let me know what you think

rotty3000:
Hi @frantuma thanks for reviewing :)

> The main one: if I am not mistaken the generated JARs manifests do not include anymore an Automatic-Module-Name entry, even if this seems to be configured in the new plugin. This would need to be fixed to consistently provide automatic module name to JDK 9+ modules based envs.

If you perform a binary comparison of the resulting jars before and after (which I have done for all jars) you will note that the `Automatic-Module-Name` is correctly preserved.

> There are differences in the outcome related to Export-Package, Import-Package and possibly Require-Capability for all modules; This is probably ok, and even "better" than what previously provided but I cannot at the moment do thorough tests in OSGi env for all modules to verify their correct behaviour. I guess you are more familiar and up to date with OSGi internals, and actively using swagger-core in OSGi, I would therefore ask you to confirm that all packages have been tested in OSGi env and that they correctly define imports and exports.

This is expected and desired :) I've tested the result in OSGi using the Apache JAX-RS Whiteboard project for which I wish to update the version of swagger if/when this change gets merged/released.

I'm also prepared to keep the metadata up to date and address issues (as time permits) if someone pings me.

Thanks for you time and effort on swagger.

Issue #2506, opened 2017-11-06
No support for "deprecated" on properties


martypitt:
It appears as though the [`Property`](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/models/properties/Property.java) model doesn't have support for becoming deprecated, although the spec permits it.

webron:
swagger-core 1.5 (the version you point to) is for Swagger/OpenAPI 2.0 which does not support `deprecated` at the property level. That has changed in OAS3 and is supported in the 2.0 branch.

MichakrawSB:
Closing this issue as part of our https://github.com/swagger-api/swagger-core/issues/5074.
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.
Thank you for your understanding!

PR #3844, opened 2020-12-18
Add support for java.lang.Deprecated on classes and fields

joschi:
The `@Deprecated` annotation hasn't been taken into account for setting the `deprecated` attribute of schemas.

This change set adds support for the `@Deprecated` annotation, for example a class with this annotation will result in a schema with the attribute `deprecated = true`.

Refs #2506 (possibly)


MichakrawSB:

Issue #3836, opened 2020-12-16
Supporting jakarta namespace


mzhafez:
Are there any plans to support jakarta namespace (jakarta ee 9) in swagger? With each passing day there are more libraries that have migrated to the jakarta namespace which is incompatible with javax namespace. 

The namespace incompatibility will slowly lead to developers having to choose between using swagger or remaining on older versions of libraries that are still compatible with jakarta. 

For example, the latest version of Jetty supports Jakarta namespace and is now incompatible with the javax namespace which means that one cannot upgrade to the latest version of Jetty and continue using swagger.

frantuma:
This is a quite delicate issue and has not been decided yet, and any comment or contribution in this area are more than welcome, however IMO the best direction and the current plan (no ETA yet) is aiming for support of both `javax` and `jakarta` namespaces at least for the short - medium term (until things stabilize).

The idea is to have parallel releases of original (let's say `javax` based) and modified (`jakarta` based) artifacts, using either direct bytecode manipulation and/or existing specific tools (e.g. 
https://github.com/eclipse/transformer), with minor refactoring beforehand (e.g. hardcoded `javax` etc).

A more comprehensive solution - again depending on how things evolve - would include refactoring the next major version to introduce a compatibility layer, this would however need much more effort than artifact transformation at build/deploy time.

I would say that in any case it's not an option to maintain 2 parallel major release lines for `javax` and `jakarta`, if not based on artifact transformation.

Again and as always contributions are appreciated.

frantuma:
see also https://github.com/swagger-api/swagger-core/issues/3849#issue-779412695

strangelookingnerd:
See my PR #3852

@frantuma I see your reasoning regarding the support of both namespaces and agree that it would be good from a maintainers POV. However I feel that neither transformer nor a compatibility layer are good solutions from a users perspective. 
Looking at Swaggers main dependencies and how they solved this:

JAXB: Clear cut with v3.0.0
Jersey: Clear cut with v3.0.0
Jakarta WS: Clear cut with v3.0.0
Jackson: Supporting both namespaces with shaded JARs (ugly in my opinion). They now provide different JARs seperated by classifier. This is a pain to work with.

I personally would prefer to maintain two major releases for the time being and deprecate the javax version as soon as reasonable.
The future will be jakarta if we like it or not ;)

I'm really willing to support you on this, so please let me know if this is a step towards a solution or not.

frantuma:
[...]

PR #3852, opened 2021-01-08
Enable support for jakarta.* namespace #3836

strangelookingnerd:
This is a rough first try to support the jakarta.* namespace. 
Summary of the changes so far:

- Dependencies changed:
-- jakarta.ws bumped to 3.0.0
-- jakarta.xml.bind-api  bumped to 3.0.0
-- jakarta.activation-api  bumped to 2.0.0
-- jersey bumped to 3.0.0
-- jackson bumped to 2.12.0
- Imports/references changed from javax.* to jakarta.* where applicable
- Testing with rest-assured seems to be impossible unless it is upgraded as well. Had a go with jersey-test-framework using grizzly (jetty requires a JDK updgrade...) however the tests fail during mvn build. Odd since they run just fine from workspace :/


frantuma:
@strangelookingnerd Closing this in favor of #3881 as discussed in #3836. Worth "keeping it around" for future work on definitive jakarta migration

PR #3877, opened 2021-02-11
Fix unstable ordering of urls in gradle resolve task

### LINKED ISSUE #3876: Gradle plugin fails randomly
Hi,

I'm experiencing a strange issue. For some reason the `resolve` task fails for some people but 
not for others. 

We narrowed the issue down that it has to be somehow classpath ordering related, as we have 
two conflicting versions of `swagger-core` in our classpath. This is an issue of its own but we 
never noticed, as the build never failed. 

I guess the reason we experiencing this, is that the resolve task uses an `Set` to filter out 
duplicated urls [here](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-gradle-plugin/src/main/java/io/swagger/v3/plugins/gradle/tasks/ResolveTask.java#L338).
As this set then gets transformed to an array the ordering is not stable. Which means that
for some of my colleagues the wrong version is earlier in the resulting url classpath causing
the build to fail.

patrickpichler:
The use of a set to filter dublicated jars in the resolve task, resulted
in an unstable ordering of the urls. This caused the resolve task to
fail on some machines and not on other, caused by the url paths.

Fixes #3876

### DIFF
--- a/modules/swagger-gradle-plugin/src/main/java/io/swagger/v3/plugins/gradle/tasks/ResolveTask.java
+++ b/modules/swagger-gradle-plugin/src/main/java/io/swagger/v3/plugins/gradle/tasks/ResolveTask.java
@@ -1,5 +1,6 @@
 package io.swagger.v3.plugins.gradle.tasks;
 
+import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


frantuma:
Thanks!

PR #3880, opened 2021-02-12
Fix @JsonUnwrapped does not copy required properties: Resolves #3879

### LINKED ISSUE #3879: @JsonUnwrapped required properties will not be copied
See: https://github.com/springdoc/springdoc-openapi/issues/1060

When a Type is annotated with @JsonUnwrapped and the unwrapped class does have required properties they won't be mapped to the unwrapping Schema.

I've already resolved the issue and created a PR

tobi6112:
Fix @JsonUnwrapped does not copy required properties:

https://github.com/swagger-api/swagger-core/issues/3879

Let me know if you want some changes.


dtanzer:
I ran into this bug too, would need a solution.

Also @tobi6112 , how did you solve it locally? by building swagger-core from source with your fix? Or do you have a workaround for spring boot?

tobi6112:
@dtanzer yes I built it locally by running `mvn install`. This will install swagger-core into your local maven repository. Thema springdoc in my case will use the local dependency.

frantuma:
@tobi6112 Thanks!

toucheqt:
This is not fixed properly yet. If the @JsonUnwrapped annotation contains a prefix or a suffix then the property is not added to the required list.

Issue #3453, opened 2020-02-21
Class with cyclic reference and non property id results in NPE


xardif:
Creating class as below and returning it from an API results in exception (at the bottom of the issue).

```
@JsonIdentityInfo( generator = ObjectIdGenerators.UUIDGenerator.class )
public class ModelWithJsonIdentityCyclicUUID
{
    public ModelWithJsonIdentityCyclicUUID parent;

}
```

I have tried using `@JsonManagedReference`, but this does not result in anything.
The only workaround is to migrate to `ObjectIdGenerators.PropertyGenerator` and create additional property in model as below:

```
@JsonIdentityInfo( generator = ObjectIdGenerators.PropertyGenerator.class, property = "id")
public class ModelWithJsonIdentityCyclicProperty
{
    public String id;
    public ModelWithJsonIdentityCyclicProperty parent;

}
```

The above classes can be used to extend test cases in `io.swagger.v3.jaxrs2.resources.JsonIdentityCyclicResource`.

The exception:
```
java.lang.NullPointerException
	at io.swagger.v3.core.jackson.ModelResolver$GeneratorWrapper.process(ModelResolver.java:1116)
	at io.swagger.v3.core.jackson.ModelResolver$GeneratorWrapper.access$100(ModelResolver.java:981)
	at io.swagger.v3.core.jackson.ModelResolver$GeneratorWrapper$3.processAsProperty(ModelResolver.java:1051)
	at io.swagger.v3.core.jackson.ModelResolver$GeneratorWrapper.processJsonIdentity(ModelResolver.java:1098)
	at io.swagger.v3.core.jackson.ModelResolver.resolve(ModelResolver.java:295)
	at io.swagger.v3.core.converter.ModelConverterContextImpl.resolve(ModelConverterContextImpl.java:90)
	at io.swagger.v3.core.converter.ModelConverters.resolveAsResolvedSchema(ModelConverters.java:112)
	at io.swagger.v3.core.util.ParameterProcessor.applyAnnotations(ParameterProcessor.java:62)
	at io.swagger.v3.jaxrs2.DefaultParameterExtension.extractParameters(DefaultParameterExtension.java:116)
	at io.swagger.v3.jaxrs2.Reader.getParameters(Reader.java:1282)
	at io.swagger.v3.jaxrs2.Reader.read(Reader.java:514)
	at io.swagger.v3.jaxrs2.Reader.read(Reader.java:121)
[...]

PR #3908, opened 2021-03-07
fix NPE with @JsonIdentityInfo at class level

### LINKED ISSUE #3907: NPE with @JsonIdentityInfo at class level
ModelResolver throws a NullPointerException when using the IntSequenceGenerator at class level as shown in the following example.

``` java
public class Pojo {
	public Set<SubPojo> sub = new HashSet<>();
}

@JsonIdentityInfo(generator = ObjectIdGenerators.IntSequenceGenerator.class)
public class SubPojo {
	public Set<SubPojo> sources = new HashSet<>();
}
```

The reason for this is that because of IntSequenceGenerator a property id is generated in order to add it to the model of the class SubPojo, but modelByType at ModelConverterContextImpl does not contain an AnnotatedType for this class at this time.

daniel-schel:
fixes #3907 #3453


frantuma:
Thanks!

PR #3917, opened 2021-03-31
add additional OpenAPIExtension::decorateOperation overload

### LINKED ISSUE #3916: OpenAPIExtension::decorateOperation is missing declaring class type information
Class type information is missing for Resource methods decorated with @Operation defined on a superclass when using OpenAPIExtension::decorateOperation.

`method.getDeclaringClass()` returns the superclass, not the resource class being parsed at the moment.

I created a possible fix for this, please review the linked PR. Thanks!

fabianvo:
fixes #3916

I added a second overload to avoid breaking existing OpenAPIExtension implementations.

Having access to a declaring class allows implementing some resource methods in abstract base classes without needing to override in each subclass just to decorate them with all required annotations.

I included a simple use case in DecoratorExtensionTest.java

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -2443,8 +2443,8 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
     }
 
     protected Boolean resolveNullable(Annotated a, Annotation[] annotations, io.swagger.v3.oas.annotations.media.Schema schema) {
-        if (schema != null && schema.nullable()) {
-            return schema.nullable();
+        if (schema != null && (schema.nullable() || io.swagger.v3.oas.annotations.media.Schema.RequiredMode.NOT_REQUIRED.equals(schema.requiredMode()))) {
+            return true;
         }
         return null;
     }


modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
```diff
@@ -664,6 +660,20 @@ public OpenAPI read(Class<?> cls,
...
+            extension.decorateOperation(operation, method, chain);
+        }
+
+        final Iterator<OpenAPIExtension> chain2 = OpenAPIExtensions.chain();
+        if (chain2.hasNext()) {
+            final OpenAPIExtension extension = chain2.next();
+            extension.decorateOperation(cls, operation, method, chain2);
+        }
+    }
+
```
frantuma, line 676: I would rather use same iterator and invoke in sequence the 2 overloads, is there a reason why two iterations are necessary?

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
```diff
@@ -664,6 +660,20 @@ public OpenAPI read(Class<?> cls,
         return openAPI;
     }
 
+    protected void decorateOperation(Class<?> cls, Operation operation, Method method) {
```
frantuma, line 663: Would possibly just add javadoc `@since 2.1.9`

modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/ext/OpenAPIExtension.java
```diff
@@ -27,4 +27,14 @@ ResolvedParameter extractParameters(List<Annotation> annotations, Type type, Set
      * @param chain     the chain with swagger extensions to process
      */
     void decorateOperation(Operation operation, Method method, Iterator<OpenAPIExtension> chain);
+
+    /**
+     * Decorates operation with additional vendor based extensions.
+     *
```
frantuma, line 33: would add some clarification in javadocs, mentioning that new overload is available `@since 2.1.9` and that the two overloads are both invoked in a given order

frantuma (changes requested):
@fabianvo Thanks for the contribution! I added a couple of comments which I'd rather have addressed before merging

MartinWahnschaffe:
@fabianvo Would be great to have this PR merged. Any chance you have another look at this?

MichakrawSB:

Issue #3911, opened 2021-03-11
YAML errors on config files are shadowed by "false" JSON errors


adagios:
With the `swagger-maven-plugin`, when using either `openapiFilePath` or `configurationFilePath` with YAML files, if there are errors with the YAML a "false" JSON error is shown instead.

Example:
```
[ERROR] Failed to execute goal io.swagger.core.v3:swagger-maven-plugin:2.1.7:resolve (default) on project recordm-rest: com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'openapi': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
[ERROR]  at [Source: (String)"openapi: 3.0.1
[ERROR] info:
[ERROR]   title: RecordM API
[ERROR]   description:
[ERROR]     The RecordM REST API<br/><br/>
[ERROR]         **Note**: To  use the 'Try it out' buttons, you must be logged in.
[ERROR]   version: 1.0.0
```

In this case, the YAML parser is failing because de `description` string is unquoted and has `:` characters.

The problem is that only the exception of the last parser is shown.
The relevante code is in [SwaggerMojo.java](https://github.com/swagger-api/swagger-core/blob/438230d79d4597d6b703437a03c32cdbf8a28ba5/modules/swagger-maven-plugin/src/main/java/io/swagger/v3/plugin/maven/SwaggerMojo.java#L207):
```
            // iterate through mappers and see if one is able to parse
            for (BiFunction<String, Class<T>, T> mapper : mappers) {
                try {
                    instance = mapper.apply(fileContent, outputClass);
                    break;
                } catch (Exception e) {
                    caughtEx = e;
                }
            }

            // if no mapper could read the content correctly, finish with error
            if (instance == null) {
                if (caughtEx == null) {
                    caughtEx = new IllegalStateException("undefined state");
                }
                getLog().error(format("Could not read file '%s' for config %s", pathObj.toString(), configName), caughtEx);
                throw new IllegalStateException(caughtEx.getMessage(), caughtEx);
            }

```

I would be happy to contribute a PR, but I am not sure what the preferred solution would be:

1. Separate the parsers by file extension, instead of only sorting them
2. Keep all the exceptions and:
   2.1. log them all
   2.2 try to log only the one corresponding to the preferred parser

frantuma:
[...]

PR #3940, opened 2021-04-20
ref #3911 - gives preferred status to first mapper errors

adagios:
Implements option 2.1 discussed on the Issue: logs errors for all mappers, giving more importance to the preferred one.

I wrote a test to confirm the behaviour, but it produces a lot of error logging and I don't know if it makes sense to include it. I tend to dislike tests that log a lot of stacktraces.


frantuma:
Thanks!

PR #3975, opened 2021-06-21
addd flag for media type example set value

HugoMario:


modules/swagger-models/src/main/java/io/swagger/v3/oas/models/examples/Example.java
```diff
@@ -78,10 +80,11 @@ public Object getValue() {
 
     public void setValue(Object value) {
         this.value = value;
+        valueSetFlag = true;
```
frantuma, line 83: see how this is done in [Schema](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/Schema.java#L690-L700) to correctly handle all scenarios

modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/MediaType.java
```diff
@@ -93,10 +95,11 @@ public Object getExample() {
 
     public void setExample(Object example) {
         this.example = example;
+        exampleSetFlag = true;
```
frantuma, line 98: see how this is done in [Schema](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/Schema.java#L690-L700) to correctly handle all scenarios

frantuma (changes requested):

frantuma (approved):

PR #4020, opened 2021-09-25
allow arbitrary keywords without x- prefix

HugoMario:


frantuma:
@HugoMario please see comment in https://github.com/swagger-api/oss-planning/issues/386#issuecomment-927601800, closing this

PR #4046, opened 2021-11-10
Fixed example of date formatted string

satoshi-kumano:
# Summary

When format of the string is `date` and example is set to '2021-11-10', expected example string is '2021-11-10'.
Current implementation returns "Wed Nov 10 09:00:00 JST 2021". I afraid this is not example input corresponding to the definition.

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/deserialization/properties/MapPropertyDeserializerTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/deserialization/properties/MapPropertyDeserializerTest.java
@@ -228,4 +228,30 @@ public class MapPropertyDeserializerTest {
         assertTrue(example instanceof String);
         assertEquals(example, "ok");
     }
+
+    @Test(description = "date example format should be yyyy-MM-dd")
+    public void testDateFormatStringExample() throws Exception {
+        Operation operation = Yaml.mapper().readValue(
+                "      responses:\n" +
+                        "        \"200\":\n" +
+                        "          content:\n" +
+                        "            '*/*':\n" +
+                        "              description: OK\n" +
+                        "              schema:\n" +
+                        "                type: object\n" +
+                        "                properties:\n" +
+                        "                  date:\n" +
+                        "                    type: string\n" +
+                        "                    format: date\n" +
+                        "                    example: 2020-03-03\n",
+                        Operation.class);
+
+        ApiResponse response = operation.getResponses().get("200");
+        assertNotNull(response);
+        Schema schema = (Schema)response.getContent().get("*/*").getSchema().getProperties().get("date");
+        Object example = schema.getExample();
+        assertNotNull(example);
+        assertTrue(example instanceof String);
+        assertEquals(example, "2020-03-03");
+    }
 }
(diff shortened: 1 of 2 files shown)


satoshi-kumano:
Hello.

I found this issue when I using OpenAPITools/openapi-generator.
https://github.com/OpenAPITools/openapi-generator/issues/5847

Could not find reported issue in this repository. Please let me know if issue is required.

pjfanning:
If there is an issue with DateSchema, would there also be an equivalent issue with DateTimeSchema?

satoshi-kumano:
@pjfanning 
Thank you for your comment.

DateTimeSchema does not have equivalent issue. I have added test in the commit and confirmed it returns example in expected format.

https://github.com/swagger-api/swagger-core/pull/4046/commits/53286ae2ddfefca03e573729d670eb3cb5475c6f

frantuma:
@satoshi-kumano thanks for reporting this, however the issue seems not to be reproducible, as if I am not missing something serialization is already handled via mixins in [this class](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/mixin/DateSchemaMixin.java)

Following code would print out correctly serialized date example.

```java

DateSchema s = new DateSchema();
s.example(new Date());
Json.prettyPrint(s);

```

Possibly you are serializing using a non configured ObjectMapper instead of the provided one by `Json` or `Yaml` classes; if this is the case please use one of the above or a rightly configured one.

Please close this PR if this solves your issues, otherwise please provide a reproducible test case.

satoshi-kumano:
@frantuma thank you for your review.
I understood that it serializes example in correct format when `Json` or `Yaml` ObjectMapper is used to.

I thought the Schema class is a model represents OAPI schema object and it can be used to generate code from generator programs.
In these cases developers won't use `Json` or `Yaml` ObjectMapper since their goal is not serializing Schema object into JSON or YAML. Usually they use their code templates and expects `getExample` it self  renders example string in correct format.

Example:  `{{model.getExample()}}`

Do I misunderstood usage of swagger-core APIs?

frantuma:
@satoshi-kumano `swagger-models` package (with `Schema` and `DateSchema`) is indeed a java representation of the Schema property of OAI 3.0.0. `DateSchema` is a convenient representation of a schema with `{type: "string", format: "date"}`, casting/converting the example value into the Java `Date` class. `DateShema.getExample()` returns therefore a Date, which can be serialized according to the user scenario, e.g. as mentioned above in swagger-core JSON serialization where format chosen is `yyyy-MM-dd`. In other scenarios other formats might be needed, so this is not mandated by the swagger-models library.

In your case, assuming your `{{model.getExample()}}` is a serialization within a template, you can apply your own format e.g. in handlebars with `{{dateFormat model.getExample() yyyy-MM-dd}}`

Said that, the usage of specialized `Schema` implementations (as `DateSchema`, `ByteArraySchema` etc) has been an opinionated choice to provide users with a Java representation of the related value for `default`, `enum` and `example` (instead to have them "type unaware") and I understand it be somehow puzzling in  other serialization scenarios. For this reason they will be dropped in next major version. 

Just as a side note, such constructs (again `DateSchema`, `ByteArraySchema` etc) are used by Json deserializer (and swagger-parser) while deserializing from a OAI spec as string; if you're e.g. building your own spec programmatically you can just not use them and e.g. use `Schema<String>` with `type=string` and `format=date`.

satoshi-kumano:
@frantuma 

Got it.
My proposal changes public api return type, It is defined as Object but some program might expect Date.
I will wait for next major update and find a solution in application side.
Thank you for sharing the solutions in application side!.

martin-mfg:
> For this reason they will be dropped in next major version.

Hi @frantuma, is there any update on this? Or any idea when the next major version will be released?
As an alternative, would it be possible to just add a new field like `exampleString` to the existing DateSchema class, which stores the original, unparsed example?

htrost:
Why this ticket was closed without a merge to the master?

PR #4055, opened 2021-11-25
fix: Support enums with unbounded wildcards

christopher-cudennec:
This is a fix for enums with unbounded wildcards. We get a NullPointerException otherwise because `enumConstants` is `null`.


frantuma:
Thanks!

PR #4105, opened 2022-01-26
#4103 Support for Jackson reference types

### LINKED ISSUE #4103: Support for additional reference types (other than JDK8 and Guava `Optional`)
I'm using reference types (i.e. AtomicReference<>) in my domain objects to model the three-state semantic of set (value is available), not set (no value available or not of interest) and unset (clear existing value). This is supported out-of-the-box by Jackson and is mapped to `<value>`, `undefined` and `null` in JSON/JavaScript respectively.

Everything work pretty well except for the generated API documentation. I'm using swagger 1.6.2 with spring (not boot) and CXF. The problem is that reference types are not "unwrapped" in the generated documentation.

For example the type MyDomainObject has a single String field description. In Java this value is wrapped by AtomicReference<String> as outlined above:

```java
@JsonInclude(value = Include.NON_NULL)
public class MyDomainObject {
  private AtomicReference<String> description;

  public AtomicReference<String> getDescription() {
    return description;
  }

  public void setDescription(AtomicReference<String> description) {
    this.description = description;
  }
}
```

And the corresponding type definition in the generated swagger looks like this:

```yaml
definitions:
  MyDomainObject:
    type: "object"
    properties:
      description:
        $ref: "#/definitions/AtomicReferenceString"

  AtomicReferenceString:
    type: "object"
    properties:
      opaque:
        type: "string"
      acquire:
        type: "string"
      plain:
        type: "string"
```

But this is not correct as a JSON value for this type would look like this:

```javascript
{
  description: "Some valuable text"
}
```

That is the swagger should simply look like this:

```yaml
MyDomainObject:
  type: "object"
  properties:
    description:
      type: "string"
```

I tried to find a workaround for this problem and found that both JDK8 and Guava's `Optional` are supported by swagger out-of-the-box in the described way. However (IMHO) using `Optional` to model the described 3-state semantic kind of contradicts the intention of `Optional` as you'd have to perform `null`-checks on the `Optional` field - just my opinion. I know that `AtomicReference` was invented with another intention in mind as well but, as this is supported by `jackson-databind` ([without additional modules](https://github.com/FasterXML/jackson-databind/blob/92f809dd161a80d0d5c519f78ea18443715970ad/src/main/java/com/fasterxml/jackson/databind/type/TypeFactory.java#L1553)), I think this is currently the best way to implement this semantic.

Browsing through the swagger code I found that support for additional reference types would be quite easy to add ([OptionalUtils](https://github.com/swagger-api/swagger-core/blob/438230d79d4597d6b703437a03c32cdbf8a28ba5/modules/swagger-core/src/main/java/io/swagger/v3/core/util/OptionalUtils.java) in 2.X and [ModelResolver._isOptionalType](https://github.com/swagger-api/swagger-core/blob/8144616d59ae159bd8fc386022c5b374acd638e3/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java#L194) in 1.6).

I'd happy to provide PRs for this enhancement but want to make sure this change is going to be accepted for both 1.6 and 2.X branches of swagger.

---

For completeness: 

I posted related questions on [stackoverflow](https://stackoverflow.com/questions/70739570/swagger-openapi-support-for-reference-types-e-g-atomicreferencestring) and [smartbear](https://community.smartbear.com/t5/Swagger-Open-Source-Tools/Support-for-reference-types-e-g-AtomicReference-lt-String-gt/td-p/228737) already but got no feedback, as of now

dpeger:
In addition to the hard coded `Optional` reference types the type information from jackson is used to unwrap additional reference types (e.g. `AtomicReference`)

Fixes #4103 for 1.5 branch (swagger 1.6.X)

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/jackson/ModelResolver.java
@@ -170,7 +170,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
             if (propType.isEnumType()) {
                 property = new StringProperty();
                 _addEnumProps(propType.getRawClass(), (StringProperty) property);
-            } else if (_isOptionalType(propType)) {
+            } else if (_isReferenceType(propType)) {
                 property = context.resolveProperty(propType.containedType(0), null);
             } else {
                 // complex type
@@ -191,9 +191,9 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
         return property;
     }
 
-    private boolean _isOptionalType(JavaType propType) {
+    private boolean _isReferenceType(JavaType propType) {
         return Arrays.asList("com.google.common.base.Optional", "java.util.Optional")
-                .contains(propType.getRawClass().getCanonicalName());
+                .contains(propType.getRawClass().getCanonicalName()) || propType.isReferenceType();
     }
 
     @Override
--- a/modules/swagger-core/src/test/java/io/swagger/jackson/ModelResolverTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/jackson/ModelResolverTest.java
@@ -1,16 +1,25 @@
 package io.swagger.jackson;
 
+import com.fasterxml.jackson.databind.BeanDescription;
 import com.fasterxml.jackson.databind.JavaType;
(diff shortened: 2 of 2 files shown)


frantuma:
Thanks!

PR #4109, opened 2022-01-27
add item support in ModelImpl

gracekarina:
This PR adds support for items all Schema types.
refs: [swagger-api/swagger-parser/1636](https://github.com/swagger-api/swagger-parser/issues/1636)

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/models/ModelImpl.java
+++ b/modules/swagger-models/src/main/java/io/swagger/models/ModelImpl.java
@@ -23,12 +23,26 @@ public class ModelImpl extends AbstractModel {
     private String description;
     private Object example;
     private Property additionalProperties;
+    private Property items;
     private String discriminator;
     @JsonProperty("default")
     private String defaultValue;
     private List<String> _enum;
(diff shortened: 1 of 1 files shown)


frantuma (changes requested):
Possibly there is a better way to handle this as adding `items` to `ModelImpl` leaving also the field in `ArrayModel` will possibly break things and be source of confusion.

As this is meant to fix https://github.com/swagger-api/swagger-parser/issues/1636, the best course of action would be probably to update [deserializer](https://github.com/swagger-api/swagger-core/blob/8d4e2cb87f2bb7a624f24c6cb5c8e1c0a5fdb6e4/modules/swagger-core/src/main/java/io/swagger/util/ModelDeserializer.java#L56) to also deserialize to `ArrayModel` when `type` is empty AND `items` field is present. On the parser side, a similar change can be implemented in [SwaggerDeserializer](https://github.com/swagger-api/swagger-parser/blob/9143c220ddc9d7f6810ad4dab80152c8d09696d2/modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java#L779)

PR #4113, opened 2022-02-04
Json deserialization process should not remove extension for Security Scheme

vgalloy:
Motivation
According to the specification the "Security Scheme Object" may be extended with Specification Extensions.

Modification
* Add test to highlight the bug
* SecuritySchemeDeserializer now parse other field


frantuma:
Thanks!

PR #4128, opened 2022-02-22
hide unannotated properties in JsonView context

### LINKED ISSUE #4127: Missing JsonView annotation interpreted wrong
The `JsonView` annotation is interpreted slightly wrong: fields which are not annotated with it are actually **not** serialized if a view is specified, they are only included in the default serialization. This is described in [this blogpost](https://www.baeldung.com/jackson-json-view-annotation) and I'm using this behaviour in my own code. The original MR for this (#2662) stated it differently, namely that it's always included.

The affected method is here:
https://github.com/swagger-api/swagger-core/blob/bae7c5cad477887888d97c8c6081a850d535b397/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L2183

Plunts:
Closes #4127


frantuma:
see comment in [related ticket](https://github.com/swagger-api/swagger-core/issues/4127#issuecomment-1299879491)

Issue #3853, opened 2021-01-11
allOf refs are broken when using fully qualified model names


haakonmt:
I am using Swagger (2.1.6) through `springdoc-openapi` (1.5.2) to generate json schemas for my Spring Boot services. Due to having name clashes, I have to set the `TypeNameResolver#useFqn` flag to true. This leads to allOf-refs (subtypes) being broken, as they are not prefixed with `#/components/schemas/`. It looks very similar to [this issue](https://github.com/swagger-api/swagger-core/issues/3727). [This line](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L1352) looks like it _might_ be the culprit, but haven't had the time do dig into the code properly.

frantuma:
Thanks for reporting this. Can you possibly add a test code reproducing the issue?

frantuma:
Thanks for reporting and investigating this! fixed in #4146

PR #4141, opened 2022-03-19
Fix for issue #3853. AllOf ref is incorrect for FQ class names

fennekit:


fennekit:
Fix for ticket 3853

frantuma:
Thanks! replaced by #4146 which includes this PR's test. Went for explicit setting of full ref as a safer option

fennekit:
Indeed a better solution. Thanks for using the test.

Issue #4049, opened 2021-11-19
Pattern regex is not being picked by swagger for list of String


router19:
I have added a below pattern validation for List of String. Though the pattern is picked up by swagger for String type variables while swagger does not pick a pattern for a list of strings. 

My class looks like this: 
```
public class Sample {

@Pattern(regexp="[a-zA-Z-]*")
private String key;

private List<@Pattern(regexp="[a-zA-Z]*") String values;

//getters and setters are written
```

}

The swagger output looks like : 
```
"Sample":{
     "type" : "object"
     "properties" :{
         "key" :{
              "type" : "string",
             "pattern": "[a-zA-Z]*"
         },
         "values": {
                "type":"array",
                "items":{
                     "type":"string"
              }
        }
  }
}
```

Any help on how can we get Pattern added for List of String
I am using io.swagger version 1.5.19
validation API version -> validation-api-2.0.1

gcatanese:
I have submitted a PR #4151 to address this (I was indeed able to reproduce it with a unit test)

frantuma:
fixed in #4151, thanks!

PR #4151, opened 2022-03-27
Set regex pattern on array

gcatanese:
Addressing issue #4049 where pattern is not picked up for `List<String>`

Added tested to verify


frantuma:
Thanks!

PR #4166, opened 2022-04-14
Applied on demand conversion to OAS 3.1

HugoMario:


frantuma:
@HugoMario 
included and replaced by #4180 which :

* squashes into single commit
* fixes duplicated Oas3.1 filter calls in different classes
* fixes alternative usage of user filter OR oas3.1
* fixes wrong target branch
* minor fixes

PR #4168, opened 2022-04-20
fix #3998 model schema enum values of jackson fields and private methods

### LINKED ISSUE #3998: Jackson enum value is not used if @JsonValue is on an enum field or nonpublic method
When Jackson's [`@JsonValue`](http://fasterxml.github.io/jackson-annotations/javadoc/2.12/com/fasterxml/jackson/annotation/JsonValue.html) is used on a public enum method, the Swagger enum list properly uses this value (thanks to #3553).  However, when it is used on an enum field instead, or on a non-public enum method, the Swagger enum list does not.  This causes a discrepancy between what the REST API actually accepts (as defined by Jackson) and what the Swagger definition states it does.

A real quick way to verify this is to update the [`JacksonValueEnum`](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/test/java/io/swagger/v3/core/oas/models/JacksonValueEnum.java) in the `swagger-core` test.  The [`EnumPropertyTest.testExtractJacksonEnumFields`](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/EnumPropertyTest.java#L204) test will then fail.

```java
public enum JacksonValueEnum {
    FIRST("one"),
    SECOND("two"),
    THIRD("three");

    @JsonValue
    private final String value;

    JacksonValueEnum(String value) {
        this.value = value;
    }
}
```

OR

```java
public enum JacksonValueEnum {
    FIRST("one"),
    SECOND("two"),
    THIRD("three");

    private final String value;

    JacksonValueEnum(String value) {
        this.value = value;
    }

    @JsonValue
    private String getValue() {
        return value;
    }
}
```

JasonVoss:
This fixes issue 3998. If there exist jackson fields, or private methods, then jackson will serialize these as the values for an enum. This enables looking for those values first before using the enum field name.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -77,6 +77,7 @@ import javax.xml.bind.annotation.XmlRootElement;
 import javax.xml.bind.annotation.XmlSchema;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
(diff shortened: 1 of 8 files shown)


frantuma:
Thanks!

PR #4187, opened 2022-05-25
Added items to Schema equals and hashcode

jeff-miller-cfa:
Fixed a problem where `Schema` was not using the `items` property in the `equals()` or `hashcode()` method. This appears to have been missed when moving the `items` property off of the `ArraySchema` class and onto the `Schema` class.


frantuma:
@jeff-miller-cfa Thanks a lot!

PR #4193, opened 2022-06-03
schema name null fix

kenzhang-mark43:
This is to fix the issue mentioned here
https://github.com/springdoc/springdoc-openapi/issues/1266


kenzhang-mark43:
@frantuma @char0n @gracekarinina @HugoMario @ponelat 
Could any one of you please approve the build workflow? 
This is a very easy fix.

PR #4195, opened 2022-06-03
refs #4167 - move classes from models to annotations, add swagger-ann…

### LINKED ISSUE #4167: module xxx reads package io.swagger.v3.oas.annotations from both io.swagger.v3.oas.annotations and io.swagger.v3.oas.models
Hi. My Project is based on Java 9 with named module. When I upgrade to swagger 2.2.0, I catched `module xxx reads package io.swagger.v3.oas.annotations from both io.swagger.v3.oas.annotations and io.swagger.v3.oas.models` when I compile it . I think it may cased by [split package](https://nipafx.dev/java-9-migration-guide/#Split-Packages).  Is there any way to solve it. Thinks.

Here is the [demo](https://github.com/lWoHvYe/split-package-demo) of this

rherrick:
…-models to swagger-annotations, added dependency on swagger-annotations to swagger-models

The other option to fix #4167 is just moving the OpenAPI30 and OpenAPI31 annotations to a package that doesn't begin with `io.swagger.v3.oas.annotations` but instead is under `io.swagger.v3.oas.models`, e.g. `io.swagger.v3.oas.models.annotations`. The question is whether it's more important to reduce transitive dependencies or have consistent packaging for annotations across the core libraries.


frantuma:
@rherrick thanks a lot for your reporting, analysis and contribution. As mentioned in #4167 we would go for package moving as you suggest as alternative above. Closing therefore this is favor of #4200

Issue #4212, opened 2022-07-14
IllegalAccessError: class ModelResolver tried to access protected method AnnotatedMember.getAllAnnotations()


CidTori:
In my project I have swagger-core:2.2.0 and jackson-databind:2.8.11.3, and I have an exception with the following root cause:
```
Caused by: java.lang.IllegalAccessError: class io.swagger.v3.core.jackson.ModelResolver tried to access protected method 'com.fasterxml.jackson.databind.introspect.AnnotationMap com.fasterxml.jackson.databind.introspect.AnnotatedMember.getAllAnnotations()' (io.swagger.v3.core.jackson.ModelResolver and com.fasterxml.jackson.databind.introspect.AnnotatedMember are in unnamed module of loader 'app')
	at io.swagger.v3.core.jackson.ModelResolver.resolve(ModelResolver.java:579)
	at org.springdoc.core.converters.PageableOpenAPIConverter.resolve(PageableOpenAPIConverter.java:91)
	at org.springdoc.core.converters.AdditionalModelsConverter.resolve(AdditionalModelsConverter.java:138)
	at org.springdoc.core.converters.PropertyCustomizingConverter.resolve(PropertyCustomizingConverter.java:58)
	at org.springdoc.core.converters.FileSupportConverter.resolve(FileSupportConverter.java:66)
	at org.springdoc.core.converters.ResponseSupportConverter.resolve(ResponseSupportConverter.java:77)
	at org.springdoc.core.converters.SchemaPropertyDeprecatingConverter.resolve(SchemaPropertyDeprecatingConverter.java:58)
	at org.springdoc.core.converters.PolymorphicModelConverter.resolve(PolymorphicModelConverter.java:63)
	at io.swagger.v3.core.converter.ModelConverterContextImpl.resolve(ModelConverterContextImpl.java:97)
	at io.swagger.v3.core.jackson.ModelResolver.resolve(ModelResolver.java:440)
	at org.springdoc.core.converters.PageableOpenAPIConverter.resolve(PageableOpenAPIConverter.java:91)
	at org.springdoc.core.converters.AdditionalModelsConverter.resolve(AdditionalModelsConverter.java:138)
	at org.springdoc.core.converters.PropertyCustomizingConverter.resolve(PropertyCustomizingConverter.java:58)
	at org.springdoc.core.converters.FileSupportConverter.resolve(FileSupportConverter.java:66)
[...]

PR #4214, opened 2022-07-19
refs #4212 - update jackson API usage to support Jackson < 2.9

frantuma:
refs https://github.com/swagger-api/swagger-core/issues/4212 - update jackson API usage to support Jackson < 2.9


wolfch-elsevier:
So which Maven release do I need to specify to get this fix?

Issue #4213, opened 2022-07-19
Ignore @NotNull annotations


nvervelle:
In [ModelResolver](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L105), the list `NOT_NULL_ANNOTATIONS` is defined and contains the annotations that implies that the field is required.

Is there a way to disable this behavior, or remove some annotations from the list ? Or to ignore the annotation on some attributes ?

The assumption that a field annotated `@NotNull` should always be declared as `required = true` is incorrect and we should be able to disable this behavior. It can cause problem both for incoming data structure and for outgoing data structure.

### Example of problems for incoming data structure

```java
    @Schema(description = "Additional properties")
    @JsonProperty("additionalProperties")
    @JsonSetter(nulls = Nulls.AS_EMPTY)
    @NotNull
    Map<String, String> additionalProperties
```
The parameter `additionalProperties` should be marked as `required = false`, even with the `@NotNull` annotation : when the `additionalProperties` attribute is not provided, the `@JsonSetter` annotation will ensure that the parameter is properly initialized with an empty Map. So on the Java side, it's clearly a non-nullable attribute, but on the openapi spec side it should be `required = false`.

### Example of problems for outgoing data structure

```java
    @Schema(description = "Additional properties")
    @JsonProperty("additionalProperties")
    @JsonInclude(JsonInclude.Include.NON_EMPTY)
    @NotNull
    Map<String, String> additionalProperties
```
The parameter `additionalProperties` should be marked as `required = false`, even with the `@NotNull` annotation : when the `additionalProperties` parameter is an empty map, the `@JsonInclude` annotation will ensure that the attribute is not added to the payload. So on the Java side, it's a non-nullable attribute, but on the openapi spec side it should be `required = false`.

nvervelle:
One idea to allow disabling the `NOT_NULL_ANNOTATIONS` on a given attribute : 
* Add a attribute `forceUseRequired` on `@Schema` annotation, with a default value as `false`
* Ignore the `NOT_NULL_ANNOTATIONS` when `forceUseRequired` is equal to `true`

I can submit a simple PR for this proposal.

frantuma:
Thanks for looking into it, and yes a PR is welcome

[...]

PR #4216, opened 2022-07-20
feat: allow disabling heuristics on annotations like @NotNull

nvervelle:
Add a property `forceUseRequired` on `@Schema` annotations that allows disabling the heuristics on annotations like `@NotNull` which otherwise may force the required status to true even if `@Schema(required = false)`

Proposal for issue #4213


frantuma:
replaced by #4286

PR #4276, opened 2022-10-18
Update pom.xml

avatar21:
Attempt to fix vulnerabilities from dependencies:
CVE-2022-42004
CVE-2022-42003

reference: https://mvnrepository.com/artifact/io.swagger.core.v3/swagger-core/2.2.3


frantuma:
applied version of Jackson 2.13.4 and databind 2.13.4.2 already address mentioned vulnerabilities (see https://github.com/FasterXML/jackson-databind/issues/3590 and https://nvd.nist.gov/vuln/detail/CVE-2022-42003#range-8429091)

PR #4278, opened 2022-10-22
Add support for XmlSchema annotation during model generation

RCBiczok:
The goal of this PR is to enable Model resolver of swagger-core to extract the XML namespace from XmlRootElement if no namespace attribute has been set, but an XmlSchema annotation is present in the java package where the class annotated with XmlRootElement exists

I had to add these changes, as we utilize have a XSD schema file -> JAXB Generator -> OpenAPI Model generation flow.
Changing the annotation directly is unfortunately not an option for us, hence this PR.

Please let me know if there are further questions / requests.


ponelat:
@frantuma your thoughts on this PR?

frantuma:
Thanks!

PR #4285, opened 2022-11-01
allow returning 3.1 schema types in 'getType'

frantuma:


modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/Schema.java
```diff
@@ -1110,6 +1113,10 @@ public Schema addRequiredItem(String requiredItem) {
      **/
 
     public String getType() {
+        boolean bindTypes = Boolean.valueOf(System.getProperty(BIND_TYPE_AND_TYPES, "false"));
```
andrei-ivanov, line 1116: `parseBoolean` would have been better than `valueOf`

Issue #4334, opened 2022-12-21
Springboot + Jersey 3 + Swagger-core - Nested @Beanparam not rendering correctly


leolimas:
I'm in a journey to upgrade swagger-jaxrs2-jakarta 1.6.8 to 2.2.7, I got almost everything working except some object parameters that should be exploded as inputs in the Swagger-ui and them still be interpreted as JSON input... and not as 2 distinct inputs

- Java: 17 
- Springboot: 3.0.0
- Jersey: 3.1.0
- swagger-jaxrs2-jakarta: 2.2.7

Resource Interface
```
@Tag(name = "MyResource", description = "MyResource enpoint")
@Path("/")
@RequestMapping
public interface MyResourceAPI {

  @GET
  @Path("/get/{name}/something")
  @Produces(MediaType.APPLICATION_JSON)
  @Operation(summary = "MyResource")
  @GetMapping(value = "/get/{name}/something")
  @ApiResponses(value = {
      @ApiResponse(responseCode = "404", description = "Not found"),
      @ApiResponse(responseCode = "400", description = "Bad request"),
      @ApiResponse(responseCode = "200", description = "Sucesso", content = @Content(schema = @Schema(implementation = MyResourcehResponse.class)))
  })
  Response search(@Context HttpServletRequest servletRequest, @BeanParam MyCustomRequest myRequest);
}
```
Resource Implementation
```
@Component
public class MyResourceAPIImpl extends implements MyResourceAPI {

  @Override
  public Response search(HttpServletRequest servletRequest, MyCustomRequest myRequest) {
    #hidden logic
    return Response.ok().entity(myResponse).build();
  }
}
```
Request classes
```
public class MyCustomRequest extends Request {

}

public class Request {
  @BeanParam
  @Builder.Default
  private Pagination pagination = new Pagination();
}

public class Pagination {

  @QueryParam("limit")
  @DefaultValue("200")
  @Schema(defaultValue = "200", type = "integer", format = "int32")
  private Integer limit = 200;

  @QueryParam("offset")
  @DefaultValue("0")
  @Schema(defaultValue = "0", type = "integer", format = "int32")
  private Integer offset = 0;
}
```

Swagger-ui using 1.6.8
![image](https://user-images.githubusercontent.com/332219/208974527-77d316c6-b355-4d66-b5c2-04eb14db3225.png)

Swagger-ui using 2.2.7
[...]

Issue #3973, opened 2021-06-16
Invalid schema for nested BeanParam (query) parameters


DanPetras:
Hi, there seems to be issue with processing nested @javax.ws.rs.BeanParam objects / parameters.

I'm using io.swagger.core.v3.swagger-gradle-plugin version 2.1.9
Jersey version 2.27
Jackson version 2.9.9

gradle **resolve** task generates:

```
/api/nested:
  get:
    requestBody:
      content:
        '*/*':
          schema:
            $ref: '#/components/schemas/NestedParams'
```
with 
```
components:
  schemas:
    NestedParams:
      type: object
```
instead of
```
/api/normal:
  get:
    parameters:
      in: query
      name: page
      schema:
        type: integer
        format: int32
        default: -1
      in: query
      name: size
      schema:
        type: integer
        format: int32
      ...
```

You can see whole output as [yaml](https://github.com/DanPetras/NestedBeanParams/blob/master/openapi.yaml) or [json](https://github.com/DanPetras/NestedBeanParams/blob/master/openapi.json).

I created example project at [DanPetras/NestedBeanParams](https://github.com/DanPetras/NestedBeanParams).

I also tried older versions of swagger-gradle-plugin but with no success.
Is this a bug or am I doing something wrong here?

The "old swagger" (io.swagger:swagger-jersey2-jaxrs:1.6.2) is working as expected.

leolimas:
Hi @DanPetras!!

Did you get any resolution for this problem? I just upgraded from 1.6.8 to 2.2.7 and now I'm getting the same problem my Pagination object do not resolve like 2 QueryParam (limit, offset)

DanPetras:
> Hi @DanPetras!!
> 
> Did you get any resolution for this problem? I just upgraded from 1.6.8 to 2.2.7 and now I'm getting the same problem my Pagination object do not resolve like 2 QueryParam (limit, offset)

No, I didn't find any.

ewaostrowska:
The issue has been fixed with: https://github.com/swagger-api/swagger-core/pull/4343. This will be included with the next release

PR #4343, opened 2023-01-04
fix: stops reprocessing parameters extracted from nested BeanParam #4334

vitormelo95:
#4334 #3973

As described by the above issues, when using a nested @BeanParam, the defined parameters are shown as the bean type through the schema ref. This is not the expected behavior as we should see the schema of the inner parameters.

Debugging the `DefaultParameterExtension` we discovered that this is caused by applying the parent field annotations in the extracted parameters. The solution was remove this reprocessing part as we couldn't find a situation where this is necessary.


leolimas:
@frantuma @HugoMario hey guys can check this?

plesingr-aeb:
Hi, the problem is still there - tested with 2.2.40.
Could you please proceed with the fix ASAP? I have recently invested several days refactoring into bean params and now it seems I'll have to discard everything.

ewaostrowska:
@vitormelo95 thank you for your contribution in the project. It will be included in the upcoming release

PR #4344, opened 2023-01-05
Fix Schema annotation with nullable and "null" example or defaultValue

### LINKED ISSUE #4339: Schema annotation processing isn’t working with nullable and "null" example
## Expected Behavior

According to the OpenAPI specification I should be able to write an OpenAPI document that contains `null` values.

See for reference:
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#data-types
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#schemaNullable

## Current Behavior

If I write the OpenAPI document completely by hand (programmatically or the text document itself) it works without any problems but when I use POJO classes that contains fields annotated with `@Schema` e.g.: `@Schema(nullable = true, example = "null")` it doesn’t always work correctly.

After serialization in the OpenAPI document some fields don’t have an example value at all, and others have converted `null` values. For the following types from a "null" value it produces:
- integer, number
  - No example
- string
  - "null" string value
- boolean
  - False
- object, array:
  - `null`, It works as expected

I’ve noticed that regardless of the "nullable" property only the object and array types work correctly.

## Possible Solution

Annotation processing should respect the "nullable" property and if that is true and the example is "null”, and the field has no constraints: it should produce in the OpenAPI document an example with a `null` value regardless of the annotated data type.

## Steps to Reproduce

1. Mark a field with any data type (excluding object, array) with `@Schema(nullable = true, example = "null")`.
2. After serialization in the OpenAPI document it should produce incorrect or missing example fields.

I’ve created an example web project to test this:
https://github.com/bodzso/swagger-nullable-test

## Context (Environment)

I’m trying to generate an OpenAPI document from POJOs and endpoints. Currently I cannot produce working requests that works with swagger-ui because of the errors described above. As far as I know it isn’t possible to disable default value generation in swagger-ui and with the above errors combined I didn’t find a possible workaround. Writing the entire OpenAPI document by hand isn’t feasible because the generated one is over a thousand lines.

I’d like to achieve a simple thing: don’t send some optional (nullable) fields in certain examples and don't hide (`@Schema(hidden = true)`) these fields from the Schema.

bodzso:
This PR does the following:
- When nullable is set to `true` in a `@Schema` annotation we can provide `"null"` to `example` and/or `defaultValue`

Fixes #4339

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -2191,6 +2191,10 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
         if (StringUtils.isNotBlank(format) && StringUtils.isBlank(schema.getFormat())) {
             schema.format(format);
         }
+        Boolean nullable = resolveNullable(a, annotations, schemaAnnotation);
+        if (nullable != null) {
+            schema.nullable(nullable);
+        }
         Object defaultValue = resolveDefaultValue(a, annotations, schemaAnnotation);
         if (defaultValue != null) {
             schema.setDefault(defaultValue);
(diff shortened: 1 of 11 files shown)


TotallyMehis:
👍 Waiting for this to be merged!

th9118:
Will this be merged?  Would help use out greatly.

MichakrawSB:
Related issue should be fixed with https://github.com/swagger-api/swagger-core/pull/5060, if PR is still valid, please refresh PR, create issue with description and link new PR to this issue. Thanks for contribution!

PR #4350, opened 2023-01-13
Fix to "Backslashes are added in long texts" issue.

HugoMario:
The problem is related to a quoating check feature added in jackson.
https://github.com/FasterXML/jackson-dataformats-text/blob/2.15/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java#L664-L668

Basically any text that contains a comma `,` will be quoted on serialization.
https://github.com/FasterXML/jackson-dataformats-text/blob/2.15/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/util/StringQuotingChecker.java#L132-L165

On long text this makes that backslashes be added when the text is wrapped.

So, this PR is filed to ignore the comma `,` on quoting checking to keep the text without quotes and avoid the present of backslashes.


MichakrawSB:

PR #4353, opened 2023-01-23
port support to exclude and override mappings to primitive types to 1.5

skel2007:
Default mapping for java.time.Instant is DateTime in master but Long in 1.5, and there is no way to override this behaviour. 

See my comment: https://github.com/swagger-api/swagger-core/pull/3477#issuecomment-1398563765


frantuma:
Thanks!

PR #4373, opened 2023-03-03
added support for generic type bean validation annotations

### LINKED ISSUE #4359: Support Bean Validation for generic type arguments
Original issue is [here](https://github.com/springdoc/springdoc-openapi/issues/2042) and I've been redirected to report it here.

https://github.com/springdoc/springdoc-openapi/issues/2042#issuecomment-1417665098

> @lpandzic,
This logic is handled in the swagger-core library.
Please submit this Enhancement in the correct Github 
https://github.com/swagger-api/swagger-core

The original issue is as follows:

Given the controller:

```java
import jakarta.validation.constraints.NotNull;
import jakarta.validation.constraints.Size;
import lombok.Value;
import org.springframework.web.bind.annotation.*;

import java.util.Optional;

@RestController
public class FooController {

    @PostMapping("fooRecord")
    public void fooRecord(@RequestBody FooRecord foo) {

    }

    @PostMapping("fooClass")
    public void fooClass(@RequestBody FooClass foo) {

    }

    record FooRecord(@NotNull Optional<@Size(max = 1000) String> value) {

    }

    @Value
    static class FooClass {
        @NotNull
        @Size(max = 1000)
        String value;

        public Optional<String> getValue() {
            return Optional.of(value);
        }
    }
}
```

currently with latest springdoc generated openapi looks like:

```json
      "FooRecord":{
        "required":[
          "value"
        ],
        "type":"object",
        "properties":{
          "value":{
            "type":"string"
          }
        }
      },
      "FooClass":{
        "required":[
          "value"
        ],
        "type":"object",
        "properties":{
          "value":{
            "maxLength":1000,
            "minLength":0,
            "type":"string"
          }
        }
      },
```
under components schemas.
The problematic part is the FooRecord component.

The result I'd expect in ideal case would be

```json
      "FooRecord":{
        "type":"object",
        "properties":{
          "value":{
            "maxLength":1000,
            "minLength":0,
            "type":"string"
          }
        }
      },
```

EDIT: I've assumed this doesn't work only for Optional types, but after implementing it turns out it doesn't work at all for generic type arguments on any type.

lpandzic:
Fixes #4359


frantuma:
Thanks!

lpandzic:
Quick question @frantuma , when is next release scheduled?

frantuma:
we don't have a fixed release schedule, but next one should be out within 1-2 days

lpandzic:
@frantuma soon? :)

PR #4391, opened 2023-03-24
add support of visibility of properties not annotated with jsonview b…

### LINKED ISSUE #4127: Missing JsonView annotation interpreted wrong

uc4w6c:
fix #4127.

In jackson, the DEFAULT_VIEW_INCLUSION feature changes whether or not to display properties that do not have the jsonview annotation.
This is a pr to deal with it.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/AnnotatedType.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/AnnotatedType.java
@@ -22,6 +22,7 @@ public class AnnotatedType {
     private Annotation[] ctxAnnotations;
     private boolean resolveAsRef;
     private JsonView jsonViewAnnotation;
+    private boolean includePropertiesWithoutJSONView = true;
     private boolean skipSchemaName;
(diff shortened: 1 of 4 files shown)


uc4w6c:
I would like to complete this issue.
What should I do?

cfecherolle:
Hi, I'm also interested in seeing this new setting integrated to Swagger, is there anyway we could help with this PR?

frantuma:
Merging this, please notice that this is will only impact "direct" usage of `ModelResolver`

PR #4394, opened 2023-04-14
fix #3581: recognize Hidden annotation on enum values.

### LINKED ISSUE #3581: Can not hide an enum value using @Hidden
Here is an explicit example where all 3 enum values are still rendered in the swagger output. Using the latest swagger core version with Java 8.

````
public enum Foo {

    Value1,

    @Hidden
    Value2,

    Value3,

}

````

Expected output is to only render Value1 and Value3 in the output, not all 3.

SimY4:
This fixes #3581

If you place Hidden annotation on enum constant this value will be excluded from swagger model.


GroovySmoothie (approved):
Nifty. This would be handy for deprecations.

gsdatta:
This would be awesome to have – anything we can do to help get it merged?

PR #4396, opened 2023-04-24
Added possibility to provide JSON/YAML Factory to initialise ObjectMapper

MateuszPol:
This PR add possibility to provide JsonFactory and YamlFactory when creating ObjectMapper using ObjectMapperFactory.

Change is pretty generic, but it's mainly related to SnakeYaml limit in parsing object size which is by default set on 3MB.

In this case we can provide YamlFactory with limit changed to any value user want.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/Json.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/Json.java
@@ -1,5 +1,6 @@
 package io.swagger.v3.core.util;
 
+import com.fasterxml.jackson.core.JsonFactory;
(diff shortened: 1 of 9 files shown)


frantuma (approved):

pjfanning:
This is also useful for Jackson 2.15.0 users as that version brings in [StreamReadConstraints](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.15.0/com/fasterxml/jackson/core/StreamReadConstraints.html) and users who don't like the defaults, can provide a JsonFactory with their own configured StreamReadConstraints instance.

PR #4417, opened 2023-05-24
Update Jackson to 2.15.1 to be consistent with SnakeYAML 2.0

yurkom:


frantuma (changes requested):
Thanks!, this should be updated also in https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-project-jakarta/pom.xml#L576-L577

frantuma:
Thanks!

Issue #4416, opened 2023-05-23
Enhancement: @JsonIgnoreProperties( allowGetters = true) properties are ignored when they shouldn't be.


HaruhiSuzumiyah:
As previously stated in https://github.com/swagger-api/swagger-core/issues/3410 allow getters should let the properties through. 

From https://github.com/swagger-api/swagger-core/pull/3649 and https://github.com/swagger-api/swagger-core/commit/8e4b8090e5a56fca364f3e2461a289799de0c82a we can see that the current behavior is to ignore everything. 

Instead the behavior should be to ignore properties where allowGetters is not true.

frantuma:
addressed in https://github.com/swagger-api/swagger-core/pull/4418

Issue #3410, opened 2020-01-13
Question: Is there a way to allow @JsonIgnoreProperties( allowGetters = true) to allow the variable to be present in response example values , instead of avoiding it ?


Andrew3431:
The @JsonIgnoreProperties hides the variables from appearing in the swagger document, which is expected. But in cases where the variable has allowGetters = true property set, shouldn't the variable be displayed in response example values ?
NOTE: @JsonIgnoreProperties was added not for swagger purposes in my code , but for functional use (to avoid deserialization) , therefore I cant use @JSONVIEW instead of @JsonIgnoreProperties..

Can this feature be supported ?

HaruhiSuzumiyah:
I know it's been a few years, but I'm also experiencing this issue.

frantuma:
this has been addressed in https://github.com/swagger-api/swagger-core/pull/4418

PR #4418, opened 2023-05-24
Change ModelResolver to better handle JsonIgnoreProperties

HaruhiSuzumiyah:
Rather than ignoring all properties with @JsonIgnoreProperties, we only ignore if it doesn't have the "allowGetters" = true

Since if it allowsGetters, you are going to receive it as part of the response

https://github.com/swagger-api/swagger-core/issues/3410

https://github.com/swagger-api/swagger-core/issues/4416

For example, the following line would cause var1 and var2 to be excluded:

`@JsonIgnoreProperties(value = { "var1" , "var2" })`

However if it has allowGetters, it isn't excluded
`@JsonIgnoreProperties(value = { "var1" , "var2" }, allowGetters = true`

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -2150,6 +2150,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
         		propertiesToIgnore.addAll(Arrays.asList(ignoreProperties.value()));
         	}
         }
+        propertiesToIgnore.addAll(resolveIgnoredProperties(annotations));
         return propertiesToIgnore;
     }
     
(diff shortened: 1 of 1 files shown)


modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
```diff
@@ -2146,20 +2146,23 @@ protected Set<String> resolveIgnoredProperties(Annotations a, Annotation[] annot
         Set<String> propertiesToIgnore = new HashSet<>();
         JsonIgnoreProperties ignoreProperties = a.get(JsonIgnoreProperties.class);
         if (ignoreProperties != null) {
-            propertiesToIgnore.addAll(Arrays.asList(ignoreProperties.value()));
+        	if(!ignoreProperties.allowGetters()) {
+        		propertiesToIgnore.addAll(Arrays.asList(ignoreProperties.value()));
+        	}
         }
-        propertiesToIgnore.addAll(resolveIgnoredProperties(annotations));
```
frantuma, line 2151: this shouldn't be removed  I believe

frantuma:
Thanks!

PR #4434, opened 2023-06-23
fix: makes populating instance variables accessible to subclasses

frantuma:


frantuma:
includes and replaces https://github.com/swagger-api/swagger-core/pull/4431

PR #4480, opened 2023-09-18
fix: compare string using equals instead of == for Reader class

gssbzn:
found a case where there's a string comparison using `==` instead of equals, took also the oportunity to try and reduce some warning on the class

### DIFF
--- a/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
+++ b/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
@@ -165,7 +165,7 @@ public class Reader implements OpenApiReader {
         for (Class<?> cls : sortedClasses) {
             if (ReaderListener.class.isAssignableFrom(cls) && !listeners.containsKey(cls)) {
                 try {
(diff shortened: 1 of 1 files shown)


modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/Reader.java
```diff
@@ -730,10 +728,10 @@ protected void applyPathParamsPatterns(Operation operation, Map<String, String>
                 .filter(p -> p.getSchema() != null)
                 .filter(p -> StringUtils.isBlank(p.getSchema().getPattern()))
                 .filter(p -> !Parameter.StyleEnum.MATRIX.equals(p.getStyle()))
-                .filter(p -> "string" == p.getSchema().getType() || (p.getSchema().getTypes() != null && p.getSchema().getTypes().contains("string")))
+                .filter(p -> "string".equals(p.getSchema().getType()) || (p.getSchema().getTypes() != null && p.getSchema().getTypes().contains("string")))
```
gssbzn, line 731: this is the main fix

frantuma:
Thanks!

PR #4560, opened 2023-11-29
Move SecurityScheme annotation to model mapping to core

bgK:
This moves the `SecurityParser` class responsible for populating `SecurityScheme` models from `SecurityScheme` annotations to the `core` module.
The move allows using it without taking a dependency on JAX-RS and to use it for building integrations for other technologies.

I'm opening this very simple pull request as a probe. Is moving the code for mapping annotations to models out of the jaxrs module desirable for the project? Some related code is already in the `AnnotationsUtils` class in the `core` module.


MichakrawSB:

Issue #2965, opened 2018-09-22
@Header annotation does not parse @Schema(implementation)


timothybasanov:
I've added something like this to my JAX-RS-based server:
```java
 @ApiResponse(responseCode = "204",
      description = "Operation was successful",
      headers = @Header(name = Headers.EXAMPLE_HEADER,
          schema = @Schema(implementation = ExampleTypes.ExampleHeader.class))
```

Expected: To see this header in my openapi generated spec.
Actual: Swagger core ignored a header annotation

Additional details:
  * `io.swagger.v3.oas.annotations.responses.ApiResponse#headers` is seemingly only read from `io.swagger.v3.jaxrs2.OperationParser#getApiResponses`
  * It calls `io.swagger.v3.core.util.AnnotationsUtils#getHeader` which in turn does this:
```java
if (header.schema() != null) {
            if (header.schema().implementation().equals(Void.class)) {
                AnnotationsUtils.getSchemaFromAnnotation(
```
  * Looks like if schema is specified, but it has implementation, then header is silently discarded

Recommended solutions (one of):
  * Ignore `@Schema` annotation entirely, but still add a header even if only a name is specified
  * Parse "implementation" part of header spec correctly
  * Throw exception when non-Void implementation is detected
  * Log when non-Void implementation is detected

ans1genie:
Any updates when this bug will be fixed? any workaround?
I would like to see this suggestion from @timothybasanov previous comment: 
> Parse "implementation" part of header spec correctly

ans1genie:
A possible workaround, (it may not look clean 🤷🏻), is to implement the OperationCustomizer interface, intercept/find the header and then patch the schema.

micryc:
Fixed by #4608

Issue #4196, opened 2022-06-04
Header annotation doesn't fully support OAS 3.0


christophejan:
Currently, all  3.x.x versions of OpenAPI Specification define [Header object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#headerObject) as :

> The Header Object follows the structure of the **Parameter Object** with the following changes:
> 
> 1. `name` MUST NOT be specified, it is given in the corresponding `headers` map.
> 1. `in` MUST NOT be specified, it is implicitly in `header`.
> 1. All traits that are affected by the location MUST be applicable to a location of `header` (for example, [`style`](#parameterStyle)).

But current swagger-core [Header annotation](https://github.com/swagger-api/swagger-core/blob/v2.2.0/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/headers/Header.java) seems to have a lot of missing attributes when I compare it to [Parameter annotation](https://github.com/swagger-api/swagger-core/blob/v2.2.0/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/Parameter.java) :
- [allowEmptyValue](https://github.com/swagger-api/swagger-core/blob/v2.2.0/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/Parameter.java#L83)
- [style](https://github.com/swagger-api/swagger-core/blob/2e98a903d841ae7a7afffe0b76ae34d5c4c4dc55/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/Parameter.java#L90)
- [explode](https://github.com/swagger-api/swagger-core/blob/v2.2.0/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/Parameter.java#L97)
- [allowReserved](https://github.com/swagger-api/swagger-core/blob/v2.2.0/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/Parameter.java#L104)
- [array](https://github.com/swagger-api/swagger-core/blob/v2.2.0/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/Parameter.java#L118)
- [content](https://github.com/swagger-api/swagger-core/blob/v2.2.0/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/Parameter.java#L125)

[...]

PR #4608, opened 2024-01-29
Add Header Object missing attributes

micryc:
Refs #4196

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
@@ -1287,6 +1287,9 @@ public abstract class AnnotationsUtils {
         return linkParametersMap;
     }
 
+    public static Optional<Map<String, Header>> getHeaders(io.swagger.v3.oas.annotations.headers.Header[] annotationHeaders, JsonView jsonViewAnnotation) {
+        return getHeaders(annotationHeaders, null, jsonViewAnnotation);
+    }
     public static Optional<Map<String, Header>> getHeaders(io.swagger.v3.oas.annotations.headers.Header[] annotationHeaders, Components components, JsonView jsonViewAnnotation) {
         return getHeaders(annotationHeaders, components, jsonViewAnnotation, false);
     }
@@ -1298,7 +1301,7 @@ public abstract class AnnotationsUtils {
 
         Map<String, Header> headers = new HashMap<>();
         for (io.swagger.v3.oas.annotations.headers.Header header : annotationHeaders) {
-            getHeader(header, components, jsonViewAnnotation).ifPresent(headerResult -> headers.put(header.name(), headerResult));
+            getHeader(header, components, jsonViewAnnotation, openapi31).ifPresent(headerResult -> headers.put(header.name(), headerResult));
         }
 
         if (headers.size() == 0) {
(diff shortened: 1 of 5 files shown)


modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
```diff
@@ -33,6 +34,7 @@
 import io.swagger.v3.oas.models.media.JsonSchema;
 import io.swagger.v3.oas.models.media.MediaType;
 import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.oas.models.parameters.Parameter;
```
frantuma, line 37: not used import?

modules/swagger-jaxrs2/src/test/java/io/swagger/v3/jaxrs2/annotations/operations/AnnotatedOperationMethodTest.java
```diff
@@ -359,6 +362,139 @@ static class GetOperationWithResponseMultipleHeaders {
         public void simpleGet() {
         }
     }
+    @Test
```
frantuma, line 365: I would add a new line (applies to all added methods)

frantuma (changes requested):
Seems great, only missing thing on top of unused import and new lines in testwould be - as you mentioned - support for schema.implementation / components.

As a test you could use something like:

```java
    @Test
    public void testOperationWithResponseArraySchemaImplementation() {
        Reader reader = new Reader(new OpenAPI());
        OpenAPI openAPI = reader.read(GetOperationResponseHeaderWithArraySchemaImplementation.class);
        Yaml.prettyPrint(openAPI);
        // would expect
        String expectedYAML = "openapi: 3.0.1\n" +
                "paths:\n" +
                "  /path:\n" +
                "    get:\n" +
                "      summary: Simple get operation\n" +
                "      description: Defines a simple get operation with no inputs and a complex output\n" +
                "      operationId: getWithPayloadResponse\n" +
                "      responses:\n" +
                "        \"200\":\n" +
                "          description: voila!\n" +
                "          headers:\n" +
                "            Rate-Limit-Limit:\n" +
                "              description: The number of allowed requests in the current period\n" +
                "              style: simple\n" +
                "              schema:\n" +
                "                maxItems: 10\n" +
                "                minItems: 1\n" +
                "                type: array\n" +
                "                items:\n" +
                "                  $ref: '#/components/schemas/SampleResponseSchema'\n" +
                "      deprecated: true\n" +
                "components:\n" +
                "  schemas:\n" +
                "    SampleResponseSchema:\n" +
                "      type: object\n" +
                "      properties:\n" +
                "        id:\n" +
                "          type: string\n" +
                "          description: the user id";
        assertEquals(expectedYAML, Yaml.pretty(openAPI));
    }

    static class GetOperationResponseHeaderWithArraySchemaImplementation {
        @Operation(
                summary = "Simple get operation",
                description = "Defines a simple get operation with no inputs and a complex output",
                operationId = "getWithPayloadResponse",
                deprecated = true,
                responses = {
                        @ApiResponse(
                                responseCode = "200",
                                description = "voila!",
                                headers = {@Header(
                                        name = "Rate-Limit-Limit",
                                        description = "The number of allowed requests in the current period",
                                        array = @ArraySchema(maxItems = 10, minItems = 1,schema = @Schema(implementation = SampleResponseSchema.class)))})})
        @GET
        @Path("/path")
        public void simpleGet() {
        }
    }
```

this would mean passing components from [OperationParser](https://github.com/swagger-api/swagger-core/blob/fab5000141536a5a6e9e068211fde1ec1603492b/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/OperationParser.java#L95) to the AnnotationUtils methods pipeline, as done in other cases. We must maintain backward compatibility, therefore new methods with new `components` param must be invoked from existing ones (passing null as param)

micryc:
Fixes #2965 also.

PR #4619, opened 2024-02-13
Fix #4618: process array schema in OpenAPI 3.1 same as in OpenAPI 3.0

### LINKED ISSUE #4618: OpenAPI 3.1 SpecFilter.removeBrokenReferenceDefinitions brokes ArraySchema processing
This problem is related to springdoc-openapi bug reported here: https://github.com/springdoc/springdoc-openapi/issues/2448

I attached minimalistic project to reproduce this problem.

Seems like change from ArraySchema to JSONSchema broke construction of reference tree

image

This is working when using 3.0.1 OpenAPI version.

GooDer:
Attempt to fix #4618


frantuma:
Thanks a lot for reporting the issue and for the fix!

PR #4626, opened 2024-03-07
fix: Add class loader to OpenAPIExtensions ServiceLoader

### LINKED ISSUE #3726: OpenAPIExtension's are not loaded by swagger-gradle-plugin
swagger-gradle-plugin does not load `io.swagger.v3.jaxrs2.ext.OpenAPIExtension` classes.  

The code in `io.swagger.v3.jaxrs2.ext.OpenAPIExtensions` uses `ServiceLoader.load(OpenAPIExtension.class)` to load extensions. The problem is that `ServiceLoader` uses `Thread.currentThread().getContextClassLoader()` when a `ClassLoader` is not specified which cannot see the `META-INF\services` files from within a gradle plugin. 

Using `ServiceLoader.load(OpenAPIExtension.class, OpenAPIExtensions.class.getClassLoader())` should load services correctly.

thefirstofthe300:
Fixes #3726.

I tested this locally on my own project. When the appropriate file exists in META-INF/services, the OpenAPIExtension is loaded properly by the Gradle plugin after this change.

### DIFF
--- a/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/ext/OpenAPIExtensions.java
+++ b/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/ext/OpenAPIExtensions.java
@@ -28,7 +28,7 @@ public class OpenAPIExtensions {
 
     static {
         extensions = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


frantuma:
thanks!

PR #4637, opened 2024-03-21
Fix #4634: unite processing of allOf, anyOf, oneOf relation so it wil…

### LINKED ISSUE #4634: OpenAPI 3.1 broken reference to child class with oneOf reference
Hi, I encountered another bug related to new OpenAPI 3.1 version.

When we have some object which's attribute is another object which is abstract and has at lease one implementation class it will not be rendered and error is displayed in swagger UI.

Version 3.0 output:
image

Version 3.1 output:
image

I found out that it is caused again by change in element class from ComposedSchema to JsonSchema.

I will try to provide fix also tomorrow.

Minimalistic project to reproduce:
[swagger-core-bug.zip](https://github.com/swagger-api/swagger-core/files/14668059/swagger-core-bug.zip)

GooDer:
…l work same in OpenAPI 3.1 and OpenAPI 3.0


micryc:
Fixes #4634 
Thank you @GooDer for your work and provided fix

Issue #3944, opened 2021-04-27
Default mapping of java byte array


lscorcia:
Hi, I got bounced here from CXF Jira so I apologize in advance if it's not the right place to open this issue.

Basic repro project is here: https://github.com/apache/cxf/tree/cxf-3.4.3/distribution/src/main/release/samples/jax_rs/description_openapi_v3_web . Apply the attached patch file to expose the issue:
[repro.txt](https://github.com/swagger-api/swagger-core/files/6384051/repro.txt).

I have a CXF/JAX-RS service that uses OpenAPIFeature to publish an OpenAPIv3 descriptor. As it's basically a duplicate of a SOAP Web Service, it does not have any annotation placed on its data types.
I am having troubles with fields of type `byte[]`. They are being represented in the descriptor as follows (see property called `binary`):

```
"components" : {
  "schemas" : {
    "Item" : {
      "type" : "object",
      "properties" : {
         "name" : {
            "type" : "string"
         },
         "value" : {
            "type" : "string"
         },
         "binary" : {
           "type" : "array",
           "items" : {
             "type" : "string",
             "format" : "byte"
           }
         }
       }
     }
   }
}
```

The byte array is wrapped in another array, which codegen tools parse as `List` of `byte[]` .
The correct representation for the parameter would be: 

```
"binary": {     
  "type": "string",     
  "format": "byte" 
} 
```

Now, I understand I can annotate the hell out of these objects, but since the API perimeter is huge, it would mean a lot of work for something that should probably be working out of the box.

ilopezv:
Any update on this?

PR #3955, opened 2021-05-13
Support proper mapping of the byte array

reta:
An attempt to properly map `byte[]` to the `ByteArraySchema` or `BinarySchema` depending on the format specified (if any). The idea here is to introduce the support in `PrimitiveType` of types which could be mapped to multiple schemas, depending on the format, with the first one being the default in case format is omitted. Closes https://github.com/swagger-api/swagger-core/issues/3944 and https://github.com/swagger-api/swagger-core/issues/3511.

Specification snippets generated:
```
        bytes:
          type: string
          format: byte
```

```
        binary:
          type: string
          format: binary
```

@webron would really appreciate the feedback if this change make sense, test cases are include, thanks a lot.

Issue #3511, opened 2020-04-10
Swagger generates ArraySchema instead of StringSchema for byte array.


Birthright50:
I use latest swagger 2.1.2, springdoc 1.3.1, spring boot 2.2.6 (MVC stack)
For example, one endpoint from my controller:
```
    @PostMapping(Routes.Item.External.ITEM_BY_ID_PHOTO)
    public ActionResult<Long> uploadImage(
            @PathVariable("id") @Min(1) long itemId,
            @AuthenticationPrincipal User principal,
            @RequestBody @Validated Base64ImageRequest req
    ) {
        final Long sellerId = principal.getCurrentAgent().isSeller() ? principal.getCurrentSellerId() : null;
        return ActionResult.ok(this.imageService.createImage(itemId, sellerId, req));
    }
```
Request model `Base64ImageRequest`:
```
public class Base64ImageRequest {
    private short order;
    @NotNull
    private byte[] image;
}
```
An example of a valid request:
```
{
"order": 5,
"image":  "someBase64Image"
}
```

It was expected that the generated specification with such a structure for this request:
```
      "Base64ImageRequest": {
        "required": [
          "image"
        ],
        "type": "object",
        "properties": {
          "order": {
            "type": "integer",
            "format": "int32"
          },
          "image": {
            "type": "string"
          }
        }
      }
```
But at the output, it turns out that the image field is interpreted as an array of strings.

```
      "Base64ImageRequest": {
        "required": [
          "image"
        ],
        "type": "object",
        "properties": {
          "order": {
            "type": "integer",
            "format": "int32"
          },
          "image": {
            "type": "array",
            "items": {
              "type": "string",
              "format": "byte"
            }
          }
        }
      }
```

[The problem is here](https://github.com/swagger-api/swagger-core/blob/a78875c8e24db082105d3dc3b34cb40d6513d471/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L655)
This line returns for field `image` `io.swagger.v3.oas.models.media.ArraySchema` instead of `io.swagger.v3.oas.models.media.StringSchema`

webron:
If you're using springddoc, you need to file an issue with them. They may use our annotations, but the way they parse the annotations is up to them.

Birthright50:
> If you're using springddoc, you need to file an issue with them. They may use our annotations, but the way they parse the annotations is up to them.


[...]

PR #4659, opened 2024-04-19
Support proper mapping of the byte array

frantuma:
Includes and replaces https://github.com/swagger-api/swagger-core/pull/3955

Closes https://github.com/swagger-api/swagger-core/issues/3944 and https://github.com/swagger-api/swagger-core/issues/3511.

See also https://github.com/swagger-api/swagger-parser/issues/1630

It introduces [binary-string-conversion](https://github.com/swagger-api/swagger-core/pull/4659/files#diff-cb2ef0df3091032171d05fded8d3f06c1679574101437806921706be09cad764R30) env or system variable, allowing to define behavior for resolution and conversion of byte arrays:

binary-string-conversion = base64 : resolves into ByteArraySchema/BinarySchema using Base64 for casting/conversion
binary-string-conversion = string-schema : resolves into StringSchema

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -309,10 +309,20 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
             isPrimitive = true;
         }
         if (model == null) {
-            PrimitiveType primitiveType = PrimitiveType.fromType(type);
-            if (primitiveType != null) {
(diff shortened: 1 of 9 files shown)


RInverid:
@frantuma Cool! one question: my current workaround for this issue is to annotate all endpoints returning byte[] as returning byte. Is this change compatible with this workaround? or will I need to pay attention when upgrading to the next version?

thanks a lot

frantuma:
@RInverid it should be ok

spulci:
@frantuma When we can expect this PR to be included in the next release?

Thanks

Simone

frantuma:
@spulci released 2.2.22

Issue #4662, opened 2024-04-25
If a property has no annotation, but only a getter, it should be read-only.


beyond-seunghyun:
If a property has no annotation, but only a getter, it should be read-only.

For example.
```java
public class Cap {

    public String getBrand() {
        return "no brand";
    }
}
```

A Cap's property "brand" should be read-only. Because there's no way to write "brand".

PR #4663, opened 2024-04-25
refs #4662 - If a property has no annotation, but only a getter, it should be read-only.

beyond-seunghyun:
See [issues#4662](https://github.com/swagger-api/swagger-core/issues/4662)


T3rm1:
@fehguy What can be done to get pull requests, such as this one, get merged in more quickly?

PR #4665, opened 2024-04-30
Set parameter schema base on schema or array of the @Parameter annotation

coder-hugo:
The current implementation ignores the content of `schema` and `array` of the `@Parameter` annotation, when generating the schema for a parameter. So it's impossible to specify the allowed values for a parameter of type `String` by using an `enum`. This PR will fix this.


coder-hugo:
Any news on this?

PR #4686, opened 2024-06-04
Make Jackson ObjectMapper initialization thread safe

Philzen:
Using the https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom here for thread safety.


frantuma:
included and replaced by #4739

PR #4687, opened 2024-06-04
Include source in Jackson exceptions

Philzen:
The default behavior for JsonParser.Feature.INCLUDE_SOURCE_IN_LOCATION was changed from `true` to `false` in Jackson 2.16:

https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.16#streaming-jackson-core

As swagger / OAI specs pretty much by definition are open and thus should never include sensitive data, it's much more helpful to the DX having such details in the exception messages.


MichakrawSB:

PR #4691, opened 2024-06-04
fix #4690 - Using builder setter and setting JsonProperty Access will not get property types correctly

### LINKED ISSUE #4690: [BUG] Using builder setter and setting JsonProperty Access will not get property types correctly
Using builder setter and setting JsonProperty Access will not get property types correctly.

```java
public class ResourceSearchVO implements Serializable {
	private String name;
    @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
    private ResourceEnum type;

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }

    public ResourceEnum getType() {
        return type;
    }

    public ResourceSearchVO setType(ResourceEnum type) {
        this.type = type;
        return this;
    }
}
```
image

image

fangzhengjin:


fangzhengjin:
@frantuma plz review

Issue #4641, opened 2024-03-27
Missing array type when using @ArraySchema


paulrutter:
Since we upgraded from 2.2.11 to 2.2.21, the output when using `@ArraySchema` is no longer correct.
We're outputting 3.1 via `openAPI31: true`.

The output in JSON misses `"type": "array"`.
See example on https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Annotations#arrayschema

Relevant annotations used:
```
content = @Content(
        mediaType = APPLICATION_JSON,
        array = @ArraySchema(schema =  @Schema(implementation = BulkResultBean.class)),
        examples = {
                @ExampleObject(
                    name = "Example bulk response",
                    description = "Example bulk response",
                    value = "description...."
                )
       }
})
```
Left is with 2.2.11, right with 2.2.21.
![image](https://github.com/swagger-api/swagger-core/assets/11535991/0ca12200-23c8-407f-8267-797c8a98a064)

I see several mentions of fixes in array schema in the changelog, but i doubt this output is now correct.

micryc:
Hi @paulrutter Are you still experiencing this issue ? I tried to reproduce with core 2.2.22 version and it looks like type shows up in specification. For now I am gonna close this ticket, if issue still appears for you, feel free to reopen

paulrutter:
I will check it out @micryc and reopen if needed.

taylortails:
I'm seeing this behaviour with 2.2.22 here:
https://github.com/springdoc/springdoc-openapi-demos/issues/68

It's not clear to me yet whether it's a spring or swagger issue

taylortails:
This looks to be the cause because a JsonSchema is created without a type:

https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java#L510

paulrutter:
@micryc we should reopen, as it apperantly is not fixed (in all cases at least)

owen4gh:
Changing line 510 in modles/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java from

`arraySchemaObject = new JsonSchema();`

to

`arraySchemaObject = new ArraySchema();`

appears to fix the issue.

paulrutter:
Can you create a pr for it @owen4gh?

owen4gh:
Created https://github.com/swagger-api/swagger-core/pull/4700

[...]

PR #4692, opened 2024-06-06
Missing array type when using @ArraySchema

taylortails:
Can I have some guidance on where to add a junit too?


taylortails:
#4641

taylortails:
testing suggests this is at best an incomplete fix

Issue #3960, opened 2021-05-28
[Platform][API] Remove JSonBackReferences from API Models


sb-yb:
![image](https://user-images.githubusercontent.com/73500297/119917375-49e99880-bf1b-11eb-97c8-32bb7cc0ce8a.png)

MichakrawSB:
Closing this issue as part of our [housekeeping & cleanup](https://github.com/swagger-api/swagger-core/issues/5074).
This issue is outdated and/or inactive for a long time, and we’re not able to revisit it meaningfully.

If this problem is still relevant, please open a new issue and reference this one for context.
Thank you for your understanding!

PR #4699, opened 2024-06-14
Ignore @JsonBackReference when resolving the model

### LINKED ISSUE #3682: Support for @JsonBackReference
I have a field annotated with `@JsonBackReference`, which is the proper way to handle circular references in your Java objects using Jackson. Traditionally, you would normally annotate with `@JsonIgnore` but this is not recommended by Jackson. In springdoc, it is recommended to use `@Schema(hidden = true)` but this is ignored when a Jackson annotation is present. [See this issue for more details](https://github.com/springdoc/springdoc-openapi/issues/390).

**Steps To Reproduce**
Declare the following annotations at the field level:
- `@JsonBackReference`
- `@Schema(hidden = true)`

**Expected behavior**
The field should not be present when viewing models in the swagger UI.

**Additional context**
It might be worth noting that Jackson recommends using `@JsonManagedReference` and `@JsonBackReference` to handle circular references, and thus should be supported by swagger.

JBodkin-Amphora:
Removes properties annotated with `@JsonBackReference` from the model, as Jackson automatically populates this with the parent. Without this, you must annotation the model with `@Schema(hidden = true)`

fixes #3682 and #3960

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -1,5 +1,6 @@
 package io.swagger.v3.core.jackson;
 
+import com.fasterxml.jackson.annotation.JsonBackReference;
 import com.fasterxml.jackson.annotation.JsonIdentityInfo;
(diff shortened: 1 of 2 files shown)


MichakrawSB:

Issue #4722, opened 2024-08-28
Missing schema name for enum


bennelli:
https://github.com/swagger-api/swagger-core/blob/5a92e3a6b8e8218aa3243bc443e91c8a27af4401/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L307

The name of the model is resolved already in previous lines of code, but not set into the StringSchema in case of type enum.

micryc:
Hi @bennelli when `enumAsRef` is not applied, which is the default behavior, name should not be set, and the enum should not be part of components/schemas, if you set enumAsRef to true, enum should appear as a component in generated openapi spec, is it solving your issue ? If not, please provide some details about your scenario, for now I am closing the ticket, feel free to reopen if you have any questions

PR #4727, opened 2024-08-30
set schema name for EnumTypes

sam0r040:
At Springwolf we use the parsed schemas to generate Examples. Thereby we use the field `Schema.getName()` for all schemas including enums. 

When enumAsRef is used, the name of the enum is missing in its class schema. Also, when serialized, the name of a schema is ignored because of the `@JsonIgnore` annotation.

This PR adds the schema name for enums. We found the following comment:
https://github.com/swagger-api/swagger-core/blob/5ee9cefe9455bf07d2671d9fd93fce2bd7c7b94f/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L379 
And therefore believe that all enum classes should appear as models, resulting in the adaption of the tests.

Relates to #4722

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -305,6 +305,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
 
         if (model == null && type.isEnumType()) {
             model = new StringSchema();
+            model.setName(name);
             _addEnumProps(type.getRawClass(), model);
             isPrimitive = true;
         }
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/ModelConverterTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/ModelConverterTest.java
@@ -3,6 +3,7 @@ package io.swagger.v3.core.converting;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 import io.swagger.v3.core.converter.ModelConverters;
 import io.swagger.v3.core.matchers.SerializationMatchers;
 import io.swagger.v3.core.oas.models.Cat;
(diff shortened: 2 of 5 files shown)


micryc:
Hi @sam0r040 could you share some details on how you used `enumAsRef` annotation ? I think it should work as intended

sam0r040:
Hi @micryc,
sorry for the delay.

At Springwolf, we create AsyncAPI documentation for events, similar to OpenAPI/Swagger. @pdalfarr uses enumAsRef as shown in the following code example:

```java
@io.swagger.v3.oas.annotations.media.Schema(enumAsRef = true)
    public enum MyEnumObject {
        VALUE1,
        VALUE2
    }

public record MyRootObject(MyEnumObject myEnumObjectField) {}
```

Our expectation of `enumAsRef` is that the `enum schema` is not embedded into the `MyRootObject` directly, but instead a ref to new `MyEnumObject` for the enum is created - including name, values etc. So, instead of inline, a `ref` is used.

We debugged the swagger-code and found the following two schemas when `enumAsRef` is used:
Using: `final Map<String, Schema> models = converters.readAll(MyRootObject.class);`

First, `MyRootObject`
```
Schema
- name=MyRootObject
- type=object
- properties:
  - myEnumObjectField
    - Schema(name=myEnumObjectField)
      - ref=#/components/schemas/MyEnumObject
```

Second, `MyEnumObject`
```
Schema
- name=null <-- missing name, expected: name="MyEnumObject"
- type=string
- _enum=VALUE1,VALUE2
```

Original issue at Springwolf: https://github.com/springwolf/springwolf-core/issues/829#issuecomment-2263377913

micryc:
Hi @sam0r040, I have tried to reproduce scenario with examples you provided and here is outcome:
```yaml
MyEnumObject:
  type: string
  enum:
  - VALUE1
  - VALUE2
MyRootObject:
  type: object
  properties:
    myEnumObject:
      $ref: '#/components/schemas/MyEnumObject'
```
I don't see any unexpected behaviour in that case (I used latest swagger core 2.2.26)

sam0r040:
Hi @micryc,
we're not looking at the rendered specification but at the pojos generated by `io.swagger.v3.core.converter.ModelConverters.resolveAsResolvedSchema(...)`. For those, the name attribute will be null on the referenced enum object.

PR #4728, opened 2024-09-01
Fix: Avoid caching AnnotationIntrospector to support custom module loading

sahil-ramagiri:
Do not cache the value of '_intr' because users can load custom modules later after swagger is configured, and we want to utilize their annotation inspection, just like jackson would do.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/AbstractModelConverter.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/AbstractModelConverter.java
@@ -21,7 +21,6 @@ import java.util.concurrent.ConcurrentHashMap;
 
 public abstract class AbstractModelConverter implements ModelConverter {
     protected final ObjectMapper _mapper;
-    protected final AnnotationIntrospector _intr;
     protected final TypeNameResolver _typeNameResolver;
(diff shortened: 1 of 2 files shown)


sahil-ramagiri:
Hi @frantuma,

I’d like to draw your attention to an issue regarding the binding of the annotation introspector.

**Issue:** Currently, binding the annotation introspector to _intr causes a divergence in functionality between Swagger's annotation introspector and Jackson's annotation introspector. This could lead to inconsistencies, especially when users extend Jackson with custom modules that perform custom annotation introspection.

**Solution:** To address this, I propose that we avoid binding the introspector directly and instead retrieve the current annotation introspector from _mapper. This approach ensures that both Swagger and Jackson utilize the same introspector, maintaining functionality and consistency across the system.

**Use Case:** This change is particularly beneficial for users implementing custom annotation introspection

I have referenced an implementation that demonstrates adding sealed class support without @JsonSubtype annotations.

Please let me know your thoughts or if you need any additional information.

Best regards.

sahil-ramagiri:
Hello @micryc, can I bring your attention to this PR. I believe this PR bring net positive value to the project, thoughts?

frantuma (approved):

Issue #4715, opened 2024-07-30
swagger-maven-plugin w/convertToOpenAPI31 sets dialect rejected by the swagger ui


shawnsmith:
If you configure the `swagger-maven-plugin` as follows:
```
<plugin>
    <groupId>io.swagger.core.v3</groupId>
    <artifactId>swagger-maven-plugin-jakarta</artifactId>
    <configuration>
        <convertToOpenAPI31>true</convertToOpenAPI31>
        ...
    </configuration>
</plugin>
```
Then the generated `openapi.yaml` will contain the following due to https://github.com/swagger-api/swagger-core/blob/v2.2.22/modules/swagger-core/src/main/java/io/swagger/v3/core/util/OpenAPI30To31.java#L13:
```
jsonSchemaDialect: https://json-schema.org/draft/2020-12/schema
```
Unfortunately this causes the Swagger UI to display an unfriendly warning due to https://github.com/swagger-api/swagger-ui/issues/8491. The UI expects this instead:
```
jsonSchemaDialect: https://spec.openapis.org/oas/3.1/dialect/base
```

One possible workaround would be to use the `openapi31` config setting instead, which skips setting `jsonSchemaDialect`:
```
<plugin>
    <groupId>io.swagger.core.v3</groupId>
    <artifactId>swagger-maven-plugin-jakarta</artifactId>
    <configuration>
        <openapi31>true</openapi31>
        ...
    </configuration>
</plugin>
```
but then the generated `openapi.yaml` will be missing top-level `type: object` fields in the component schemas because the generator calls `schema.type("object")` instead of `schema.types(Set.of("object"))` and skips `OpenAPI31SpecFilter`/`OpenAPISchema2JsonSchema`.

micryc:
@shawnsmith 
Fixed by: #4729 
Just to mention, the openapi31 flag is the best and correct way to resolve into an OpenAPI 3.1 specification, as it understands 3.1 annotations and correctly uses 3.1 model. convertToOpenAPI31 is an early option to get out a 3.1 doc, still usable but less "powerful"
I would suggest to use openapi31 flag, as issue with missing type: object is already fixed also
I am closing this ticket for now, if there will be still any issues, feel free to reopen

PR #4729, opened 2024-09-03
refs #4715 - change JsonSchemaDialect to one accepted by Swagger-UI

micryc:
refs issue #4715, related to [swagger-ui openapi 3.1 support]( https://github.com/swagger-api/swagger-ui/issues/8491)

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/Ticket4679Test.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/Ticket4679Test.java
@@ -15,6 +15,7 @@ public class Ticket4679Test extends SwaggerTestBase{
     public void testCustomSchemaImplementation() {
 
         String expectedYaml = "ModelWithCustomSchemaImplementationInProperty:\n" +
+                "  type: object\n" +
                 "  properties:\n" +
                 "    exampleField:\n" +
                 "      type: integer\n" +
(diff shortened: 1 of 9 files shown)


frantuma (changes requested):
The change to the `jsonSchemaDialect` is ok.

The issue of missing `type` in schema serialization when using `openapi31` is instead more a bug in `ModelResolver`

Changing [these lines](https://github.com/swagger-api/swagger-core/blob/c37b72cc653c41dc1ab90a4e10d2c957c84d52d4/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L539-L550) to the following should do the trick:

```java
            model = new ComposedSchema().name(name);
            if (openapi31) {
                model.addType("object");
            } else {
                model.type("object");
            }
        } else {
            AnnotatedType aType = ReferenceTypeUtils.unwrapReference(annotatedType);
            if (aType != null) {
                model = context.resolve(aType);
                return model;
            } else {
                model = new Schema().name(name);
                if (openapi31) {
                    model.addType("object");
                } else {
                    model.type("object");
                }
```


Also tests need to be added

Also to be mentioned in ticket: the `openapi31` flag is the best and correct way to resolve into an OpenAPI 3.1 specification, as it understands 3.1 annotations and correctly uses 3.1 model. `convertToOpenAPI31` is an early option to get out a 3.1 doc, still usable but less "powerful"

Issue #4703, opened 2024-06-18
BUG @Pattern on type parameter of properties using collections  is ignored when using oas 3.1.0


Lorenzo-Bracci:
## Problem
When generating OpenAPI 3.1.0 spec the `@Pattern` annotation on a type parameter of properties with a collection type are ignored. The expected behaviour is that the generated property has the `pattern` property set for the `items` schema but no `pattern` is found in the generated open api. 
This issue is similar to #4702 

## Expected behaviour 
`@Pattern#regexp` is used to generate `pattern` property for the `items` schema of the property

## Reproducer
The issues is reproduced in the following test:

```
    @Test(description = "Shows that @Pattern is not correctly handled in type parameters of properties using collections when using oas 3.1.0")
    public void testModelUsingCollectionTypePropertyDoesNotHandlePatternAnnotationForOas31() {

        String expectedYaml = "DtoUsingPatternOnCollection:\n" +
                "  type: object\n" +
                "  properties:\n" +
                "    myField:\n" +
                "      type: array\n" +
                "      items:\n" +
                "        pattern: myPattern\n" +
                "        type: string";

        Map<String, io.swagger.v3.oas.models.media.Schema> stringSchemaMap = ModelConverters.getInstance(true).readAll(DtoUsingPatternOnCollection.class);
        // fails as pattern will not be added to the items property of myField for oas 3.1.0
        SerializationMatchers.assertEqualsToYaml31(stringSchemaMap, expectedYaml);
    }

    public class DtoUsingPatternOnCollection {
        private List<@Pattern(regexp = "myPattern") String> myField;

        public List<String> getMyField() {
            return myField;
        }

        public void setMyField(List<String> myField) {
            this.myField = myField;
        }
    }
```

## Investigation with proposed solution

Looking at the `ModelResolver.applyBeanValidatorAnnotations` code (which is where information from bean validation annotations are processed), the `@Pattern` annotation is processed only if the `property` input parameter of type `Schema` is an instance of `StringSchema` or if `property.getItems()` is an instance of  `StringSchema`. If the check succeeds then the information from `@Pattern` is propagated to the generated schema (using the `pattern` property). Nevertheless for open api 3.1.0 `property.getItems()` will be of type `JsonSchema` instead of `StringSchema` and therefore the information from `@Pattern` will not be used. A possible solution would be that instead of using doing an `instanceOf` check against `StringSchema` a check is performed on the `type`/`types` fields to ensure that the property is a  string schema.

[...]

Issue #4702, opened 2024-06-18
BUG @Size is ignored in properties using collections when using oas 3.1.0


Lorenzo-Bracci:
## Problem
When generating OpenAPI 3.1.0 spec the `@Size` annotation on property with a collection type is ignored. The expected behaviour is that the generated property has both `maxItems` and `minItems` set but these properties are not found in the generated open api. 

## Expected behaviour 
`@Size#min` is used to generate `minItems` and `@Size#max` is used to generate `maxItems` for the generated property schema

## Reproducer
The issues is reproduced in the following test:

```
    @Test(description = "Shows that @Size is not correctly handled in properties using collections when using oas 3.1.0")
    public void testModelUsingCollectionTypePropertyDoesNotHandleSizeAnnotationForOas31() {

        String expectedYaml = "DtoUsingSizeOnCollection:\n" +
                "  type: object\n" +
                "  properties:\n" +
                "    myField:\n" +
                "      maxItems: 100\n" +
                "      minItems: 1\n" +
                "      type: array\n" +
                "      items:\n" +
                "        type: string";

        Map<String, io.swagger.v3.oas.models.media.Schema> stringSchemaMap = ModelConverters.getInstance(true).readAll(DtoUsingSizeOnCollection.class);
        // fails as the maxItems/minItems will not be added to the schema of myField for oas 3.1.0
        SerializationMatchers.assertEqualsToYaml31(stringSchemaMap, expectedYaml);
    }

    public class DtoUsingSizeOnCollection {
        @Size(min = 1, max = 100)
        private List<String> myField;

        public List<String> getMyField() {
            return myField;
        }

        public void setMyField(List<String> myField) {
            this.myField = myField;
        }
    }
```

## Investigation with proposed solution

Looking at the `ModelResolver.applyBeanValidatorAnnotations` code (which is where information from bean validation annotations are processed), there is a check that the `property` input parameter of type `Schema` is an instance of `ArraySchema`. If the check succeeds then the information from `@Size` is propagated to the generated schema (using the `minItems`/`maxItems` properties). Nevertheless for open api 3.1.0 `property` will be of type `JsonSchema` instead of `ArraySchema` and therefore the information from `@Size` will not be used. A possible solution would be that instead of using `(property instanceof ArraySchema)` a check is performed on the `type`/`types` fields of `property` to ensure that the property is an array schema.

micryc:
[...]

PR #4734, opened 2024-09-16
refs-#4703/#4702-@Pattern/@Size annotations handling on collections

micryc:
Refs:
- https://github.com/swagger-api/swagger-core/issues/4702
- https://github.com/swagger-api/swagger-core/issues/4703

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -1511,43 +1511,38 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
             if ("integer".equals(property.getType()) || "number".equals(property.getType())) {
                 property.setMinimum(new BigDecimal(size.min()));
                 property.setMaximum(new BigDecimal(size.max()));
-            } else if (property instanceof StringSchema) {
-                StringSchema sp = (StringSchema) property;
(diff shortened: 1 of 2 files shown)


modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
```diff
@@ -1519,6 +1519,11 @@ protected void applyBeanValidatorAnnotations(Schema property, Annotation[] annot
                 ArraySchema sp = (ArraySchema) property;
                 sp.setMinItems(size.min());
                 sp.setMaxItems(size.max());
+            }else if(openapi31 && property instanceof JsonSchema){
```
frantuma, line 1522: This is not sufficient I believe, we need to basically replace the `instanceof` checks with a test on the type and types. It would probably make sense to have some helper method working for both 3.0 and 3.1

so something like

```
boolean isStringSchema(Schema schema) {
 return "string".equals(schema.getType()) ||" (schema.getTypes() != null && schema.getTypes().contains("string")
}
```

and so on.
you would then use that instead e.g. of `else if (property instanceof StringSchema) {`

the same goes for array, number etc.

Also it would be probably a series of `if`, not `else`, as in oas 3.1 we could have type = "integer, string" so both can be applied

minor: please add spaces after and before `{` and `}`

The same would apply below for pattern

frantuma (changes requested):

frantuma (approved):

PR #4749, opened 2024-10-02
Implement Tests to check basic Java Record processing result

micryc:


frantuma (changes requested):
CI files should be updated to include the new module files while preparing and post release, managing versions

frantuma (approved):

Issue #4755, opened 2024-10-04
Swagger Core is not compatible with Jackson 2.18.x


jamesnetherton:
`ModelResolver.findJsonVauleType` references Jackson [`BeanDescription.findJsonVauleMethod`](https://github.com/swagger-api/swagger-core/blob/6c14ceb262d019bc4eec170b30d163c81f1b51b5/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L1127) which was [deprecated](https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.17.2/src/main/java/com/fasterxml/jackson/databind/BeanDescription.java#L259-L260) since Jackson 2.9. It got removed in Jackson 2.18.0, hence there's the potential for runtime errors if Jackson 2.18.x is on the classpath. Or build time errors if GraalVM is used to create a native application.

micryc:
Hi @jamesnetherton thank you for reporting, possible problem should be solved by: https://github.com/swagger-api/swagger-core/pull/4756, Closing ticket for now, If you will experience any problem, feel free to reopen

frantuma:
fixed in #4806 4806

PR #4756, opened 2024-10-08
Update findJsonValueType method to provide better support for Jackson…

micryc:
… < 2.9
refs: #4755


modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
```diff
@@ -1116,18 +1116,18 @@ protected Type findJsonValueType(final BeanDescription beanDesc) {
...
-            if (jsonValueMethod != null) {
-                return jsonValueMethod.getType();
-            }
+        }
+
+        jsonValueMember = invokeMethod(beanDesc, "findJsonValueMethod");
+        if (jsonValueMember != null) {
+            return jsonValueMember.getType();
+        } else {
+            LOGGER.error("jackson BeanDescription.findJsonValueMethod not found, this could lead to inaccurate result, please update jackson to 2.9+");
```
frantuma, line 1130: message is wrong

frantuma (changes requested):

frantuma (approved):

PR #4761, opened 2024-10-17
feat(gradle-plugin): enable lazy evaluation, implicit task dependencies

nbrugger-tgm:
## Why
- lazy configuration: The lack  cause pain when working in a remotely up to date gradle enviroment and 
- fixed immutable task dependencies: this is a blocker since you have no modify theese task dependencies (resolve depending compileTestJava for example) 

## Changes
  - enable lazy evaluation by
    - replacing all Task configurations with gradles Property types
    - using `register` instead of `create` to register the task
  - remove wrongly set task dependencies that can cause circular dependencies. Task dependencies are now inferred by the set classpath. When gradle detects that an output of a task (like compileJava) is used an automatic dependency is formed and `resolve` will always execute compileJava first. The advantage is that this is dynamic. If one does NOT want `compileJava` to be executed it won't be as soon as nothing of compileJava is on the configured classpath

### DIFF
--- a/modules/swagger-gradle-plugin/src/main/java/io/swagger/v3/plugins/gradle/SwaggerPlugin.java
+++ b/modules/swagger-gradle-plugin/src/main/java/io/swagger/v3/plugins/gradle/SwaggerPlugin.java
@@ -4,9 +4,10 @@ import io.swagger.v3.plugins.gradle.tasks.ResolveTask;
 import org.gradle.api.Action;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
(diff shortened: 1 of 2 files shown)


nbrugger-tgm:
Sorry for the force pushes, i had to fix the commit it had the wrong commiter was not signed and i messed up the commit message

ponelat:
Thanks @nbrugger-tgm , we'll try to take a look as soon as we can, we appreciate the code!

frantuma (approved):

frantuma:
Thanks!

PR #4783, opened 2024-11-15
fixing flaky test testFormatOfBigDecimal

xinying-7:
The test is failing NonDex since it's directly comparing JSON Strings which can have different key order, so I converted the actual and expected JSON String into JSON objects and compare with JSONAssert.assertEquals, which ensures correct equality checks regardless of key order.


xinying-7:
The previous comment is for the majority of test cases. Here's some specialized tests which I used different approach to fix:

For test case `deserialization.JsonDeserializationTest.testNullExampleAndValues`, it was comparing serialized strings , which caused it to fail NonDex due to potential variations in key order. So I used map-based comparison to ensure logical equality of the YAML structure, regardless of the key order in the serialized output.

For test case `converting.ModelConverterTest.serializeParameterizedType`, it fails NonDex due to the non-deterministic key order in class schema,  the iterator `et.next()` doesn't always return `id` as the first key and the following keys in the same order.
Since the main prupose is to check if each key has the correct schema type, I fixed the flakiness by directly access them by name instead of iterating, for example, using `props.get("id")`.

For test case `resolving.JsonPropertyTest.testTicket2845`, it fails NonDex because of variations in key order as well. So I split each lines in the YAML string and store them into HashSet objects, and directly comparing the HashSet objects, which would make sure the comparison passes regardless of order.

MichakrawSB:

PR #4791, opened 2024-11-19
Move applySchemaResolution check to a method

YousefHaggy:
Simply moves the applySchemaResolution check to a method on ModelResolver.
This way we can configure behavior in a custom ModelResolver subclass by overriding the method.


frantuma:
Thanks!

PR #4795, opened 2024-11-22
Fix for record component annotations on fields with JsonProperty annotations

richard-lindsay:
Add support for record classes with `@JsonProperty` annotations that do not match the field names when getting the component annotations. 

When these two differ a NoSuchMethodException is throw. Instead access the declared method using the name of the primary member.


frantuma (approved):

frantuma:
Thanks!

PR #4246, opened 2022-09-05
Defining grouping in swagger 

S66D:
Defining grouping in swagger endpoints to reduce the challenges brought in creating a per class request models in huge system for better maintainability and for less class redundancy.

PR #4797, opened 2024-11-22
fix: Fields with not null constraint annotations are considered optional if groups attribute is set

T3rm1:
This PR keeps properties as optional when they are annotated with one of the not-null annoatations and have a value in their groups attribute.
The groups attribute lets you define under which circumstances the annotation is evaluated. For `NotNull` this means, the field can be optional in some cases and mandatory in other cases.
Since we don't know what should apply, the parameter should be considered optional.

@frantuma It would greatly help me if we could merge this asap. I tried a few workarounds but none are fully working.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -102,6 +102,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.Objects;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


mc1arke:
For reference: I've duplicate the concept in https://github.com/swagger-api/swagger-core/pull/4805 as I'd only checked for open issues, not PRs. My solution allows the calling code to override the filtering so that certain groups can be treated as being mandatory where required, and I think this solution PR is missing annotation on method parameters for JAXRS - see `io.swagger.v3.core.util.ParameterProcessor` making a call to `io.swagger.v3.core.jackson.ModelResolver#NOT_NULL_ANNOTATIONS`.

T3rm1:
> I think this solution PR is missing annotation on method parameters for JAXRS

You might be right. I'll take a look.

didjoman:
Hi, 
I think it would be a nice improvement as the ``@NotNull`` is not well handled with ``groups`` currently.

But just so that you know, there is also this pull-request that seemed to have implemented groups 2 years ago:
https://github.com/swagger-api/swagger-core/pull/4246 
However I don't know what it is worse now, as it was not updated... 

Anyway, I think any fix on that subject would be very much appreciated. It would fix a lot of issues/questions raised by the community as we can see in the other PR.

@frantuma What is your opinion on this issue ?

PR #4803, opened 2024-12-07
fix: Replace $ in the FQN of the class

lipiridi:
Fixes swagger semantic errors:
1. Component names can only contain the characters A-Z a-z 0-9 - . _
2. $ref values must be RFC3986-compliant percent-encoded URIs


frantuma:
Thanks!

Issue #4804, opened 2024-12-07
Incorrect inclusion of javax/jakarta validation annotations with `groups` property set


mc1arke:
The javax/jakarta annotations contain a `groups` property that can be used to enable different combinations of validation on a model depending on which group is currently active - https://jakarta.ee/learn/docs/jakartaee-tutorial/current/beanvalidation/bean-validation-advanced/bean-validation-advanced.html#_grouping_constraints.

When executed against a model, swagger-core currently treats any of the supported javax/jakarta annotations as applying a constraint to the schema, even where the annotated field/attribute specifies groups and is therefore is not mandatory on all requests. By default, swagger-core should only set fields as mandatory/constrained based on annotations where no groups are specified, but provide a way of the end-user adding other validation groups into the constraints where they want to treat certain groups as always being enabled.

### Replication Steps
```java
	@Test
	public void shouldIncludeOnlyNonGroupedJakartaValidatedFieldsAsMandatoryByDefault() {
		ResolvedSchema schema = ModelConverters.getInstance(false).resolveAsResolvedSchema(new AnnotatedType().type(CustomClass.class));
		String expectedJson = "{\"schema\":{\"required\":[\"nonGroupValidatedField\"],\"type\":\"object\",\"properties\":{\"nonGroupValidatedField\":{\"type\":\"string\"},\"singleGroupValidatedField\":{\"type\":\"integer\",\"format\":\"int32\"},\"multipleGroupValidatedField\":{\"type\":\"number\"},\"otherGroupValidatedField\":{\"type\":\"string\"},\"singleGroupValidatedField2\":{\"type\":\"string\"}}},\"referencedSchemas\":{\"CustomClass\":{\"required\":[\"nonGroupValidatedField\"],\"type\":\"object\",\"properties\":{\"nonGroupValidatedField\":{\"type\":\"string\"},\"singleGroupValidatedField\":{\"type\":\"integer\",\"format\":\"int32\"},\"multipleGroupValidatedField\":{\"type\":\"number\"},\"otherGroupValidatedField\":{\"type\":\"string\"},\"singleGroupValidatedField2\":{\"type\":\"string\"}}}}}";
		SerializationMatchers.assertEqualsToJson(schema, expectedJson);
	}


	private interface ValidationGroup {

	}

	private interface OtherValidationGroup {

	}


        private static class CustomClass {

		@NotNull
		public String nonGroupValidatedField;
		@Min(value = 1, groups = ValidationGroup.class)
[...]

PR #4805, opened 2024-12-07
#4804: Only include validation constraints with no groups by default

mc1arke:
The current process for adding constraints into the schema presumes that all javax or jakarta validation annotations enforce a constraint, and doesn't take into account the 'groups' property of the annotation. This results in constrains being added to the schema that would only be applied for certain request, so incorrectly marks fields as required in the resulting schema where they may only be mandatory on a subset of requests. To overcome this, a new `ValidationAnnotationFilter` is being introduced which only treats a validation annotation as constraining the schema where the conditions on the filter are met be the annotation. The default implementation of this filter only treats constraints from annotations with no groups as being enforced to bring it inline with how the default Java Beans validation operates.


frantuma:
This has been replaced by https://github.com/swagger-api/swagger-core/pull/4853

PR #4818, opened 2025-01-07
fix patternproperties handling in SpecFilter

### LINKED ISSUE #4817: RemovingUnreferencedDefinitions functionality ignores schema refs from patternProperties
When using the 'patternProperties' functionality in conjunction with the removingUnreferencedDefinitions functionality the references placed in the 'patternProperties' attribute are written in the spec file but the reference can not be resolved because definition of the reference was removed.

Opharion:
- add support for patternProperties in SpecFilter
- fix #4817


frantuma:
Thanks!

PR #4853, opened 2025-03-10
add better support for validation groups

frantuma:


modules/swagger-gradle-plugin/src/main/java/io/swagger/v3/plugins/gradle/tasks/ResolveTask.java
```diff
@@ -510,6 +549,16 @@ public void resolve() throws GradleException {
                 method.invoke(swaggerLoader, defaultResponseCode.get());
             }
 
+            if (validatorProcessorClass.isPresent() && StringUtils.isNotBlank(validatorProcessorClass.get())) {
+                method = swaggerLoaderClass.getDeclaredMethod("setValidatorProcessorClass", String.class);
+                method.invoke(swaggerLoader, validatorProcessorClass.get());
+            }
+
+            if (defaultResponseCode.isPresent() && StringUtils.isNotBlank(groupsValidationStrategy.get())) {
```
erdi, line 557: A bug has been introduced on this line, see https://github.com/swagger-api/swagger-core/issues/4858.

PR #4861, opened 2025-03-13
Added ability to set name for ApiResponse and RequestBody annotations

altro3:
It is now possible to specify the name of any component using an annotation, except for RequestBody and ApiResponse. The name of the components is required not only for the linear description of the specification, but also so that these components can be taken out to the #/components/ block.

The idea is to add a new annotation in which the user can describe common components and then refer to these components.

In other words, the RequestBody and ApiResponse annotations currently have a `ref` field to reference the component, but there is no way to add this component to the #/components/ block

My idea is to add an OpenAPIComponents annotation something like this:

{AE43898C-05D9-4203-ADE6-08EC89BF8E8F}

To allow the code to describe common components, and add annotations with references to common components to methods / fields / parameters


altro3:
@frantuma Look to this, please. If there is anything that needs to be improved, let me know.

PR #4879, opened 2025-04-25
Consistent usage of TestNG

strangelookingnerd:
This PR aims to _remove_ usage of JUnit and consistently use TestNG across the modules instead.

Changes include:

* Use TestNG annotations over JUnit counterparts
* Use TestNG assertions over JUnit counterparts (also simplify some assertions in the process)
* Consistently use static imports for TestNG assertions
* Remove JUnit depencency if not required
* Add TestNG dependency if required

There is one exception in `swagger-maven-plugin` where JUnit is still required by the underlying maven test suite and thus tests can not be migrated.

### DIFF
--- a/modules/swagger-core/pom.xml
+++ b/modules/swagger-core/pom.xml
@@ -147,11 +147,6 @@
             <version>1.0.5</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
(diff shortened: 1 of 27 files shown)


frantuma:
Thanks! Included and replaced by #4885

PR #4880, opened 2025-04-28
issue-4850: fix extensions 'x-' prefix decoration for oas 3.1

djankows:


frantuma (changes requested):
The code changes do update the logic to use the prefix also for 3.1, but this is applied regardless of the "target". It includes scenario where the extension is set to a Schema (JSON Schema), for which the prefix is not needed and would therefore keep the logic for openapi31 before the chance

Possibly adding a parameter to the method like `usePrefix` (applied to openapi31 only, current method kept for backward compatibility) is the way to go, and see in callers which one to call
image

frantuma (approved):
@djankows Added [commit](https://github.com/swagger-api/swagger-core/pull/4880/commits/582cb3b68913485bdf73891c2ce0daf7373d930a) with a couple of additional `AnnotationsUtils.getExtensions` invocations updates that were missing.

Please review and if ok merge, please Squash commits before merging (with significant message)

PR #4929, opened 2025-07-14
BUGFIX: Bump commons lang3 version to mitigate cve

sregucki-dt:


daniel-kmiecik:
Thank you for the contribution!

We've reviewed the CVE (CVE-2025-48924) mentioned in commons-lang3 3.17 and confirmed that the affected class, ClassUtils, is not used anywhere in the project. Therefore, this vulnerability does not impact swagger-core.

We generally avoid upgrading dependencies solely due to generic CVEs unless:

- There is a proven impact on swagger-parser users, or

- A security advisory or policy explicitly requires it.

Otherwise, dependency upgrades are reviewed and batched with regular maintenance.

For future dependency update requests, could you please provide more context on why this update is needed? For example:

- Is there a known vulnerability (CVE) that is related to the code?

- Is the current version causing compatibility or build issues?

- Are there specific new features or bugfixes you need from the newer version?

Thanks again for taking the time to open this PR!

sregucki-dt:
@daniel-kmiecik 
Hey, 
as active users of the swagger-core library in our production systems, we greatly appreciate the value it brings to our API documentation workflows. 
However, we’ve recently encountered a significant concern related to dependency org.apache.commons:commons-lang3:3.17.0:
- https://nvd.nist.gov/vuln/detail/CVE-2025-48924
- https://security.snyk.io/vuln/SNYK-JAVA-ORGAPACHECOMMONS-10734078

Our internal security scanning systems have flagged a high-severity on this transitive dependency. While we have a few mitigation options on our side - such as overriding the vulnerable dependency version - we believe the most sustainable solution would be to have this dependency updated directly in swagger-core. We kindly ask you to consider bumping the affected library to a safe version (3.18.0).

Best regards, 
Sebastian
SDE, Dynatrace

daniel-kmiecik:
@sregucki-dt this will be updated in the next version with the following PR #4938

PR #4931, opened 2025-07-15
Create a failing test presenting open api 3.1 - enum property schema enumAsRef error

damiankaplon:
The test from the PR fails.

However, after changing the model resolver to generate open api 3.0 with ⬇️ works as expected
```
final ModelResolver modelResolver = new ModelResolver(mapper()).openapi31(true);
```


damiankaplon:
@frantuma Could you please have a look? Here is the [issue](https://github.com/swagger-api/swagger-core/issues/4932). I created the test on my own but the fix was generated with agentic 🤖

ewaostrowska:
Hi! Thanks for preparing the test to illustrate the issue! 
The proposed solution helped to understand what the problem is, however it did not fix the issue entirely because the Enum properties other than $ref  were not propagated to the Enum schema. For this reason I have created another PR [here](https://github.com/swagger-api/swagger-core/pull/4969)

PR #4954, opened 2025-08-14
fix: Java Byte to OAS integer data type

### LINKED ISSUE #3052: Setting `type` in @Schema has no effect
**Describe the bug**

The `type` property of `@Schema` has no effect for objects in the "Schemas" section of the Swagger page.

**To Reproduce**
Can be reproduced by cloning [this](https://github.com/sam-hieken/springdocs-schema-has-no-effect-repro/tree/master) repository, running `mvn spring-boot:run`, and navigating to [http://localhost:8081/swagger-ui.html](http://localhost:8081/swagger-ui.html).

The `UserModel` class is explicitly defined as `type = "integer"`:

```java
public class UserModel {
	...

	/** User's age */
	@Schema(type = "integer", format = "int8")
	private byte age;
}
```

 yet the default type (`string` for `byte`s in Java) is not overriden:

Image

I've tested this with `type = "integer"` on both Java `String`s and `byte`s, both with the same result.

**Expected behavior**

The `integer` type being applied to the field in Swagger.

**Screenshots**
See above

**Additional context**
Using SpringDocs 2.8.9 and Spring Boot 3.5.4 (see POM of linked repo reproducing issue for other version info).

jongminchung:
## Description

Fixes: https://github.com/springdoc/springdoc-openapi/issues/3052

This PR fixes an incorrect type mapping for Byte class in OpenAPI 3.1 schema generation.

**Problem**: The createProperty31() method in the BYTE enum was incorrectly returning a JsonSchema with type "string" and format "byte", which is not compliant with OpenAPI 3.1 specification.

**Solution**: Changed the type from "string" to "integer" for proper byte representation in OpenAPI 3.1. The byte type in OpenAPI 3.1 should be represented as an integer type, not a string with byte format.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

- [ ] I have added/updated tests as needed

image

The image above is the image referenced by [Micronut Docs](https://micronaut-projects.github.io/micronaut-openapi/latest/guide/#:~:text=Byte%20(type%3A%20integer%2C%20x%2Dformat%3A%20int8)%2C).

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -173,6 +173,13 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
             type = _mapper.constructType(annotatedType.getType());
         }
 
+        // EARLY special handling for OpenAPI 3.1 and byte[]: treat as array of primitive items
+        if (openapi31 && type.getRawClass().isArray() && type.getRawClass().getComponentType() == byte.class) {
+            ArraySchema arraySchema = new ArraySchema().items(PrimitiveType.BYTE.createProperty31());
+            arraySchema.specVersion(SpecVersion.V31);
+            return arraySchema;
+        }
+
         final Annotation resolvedSchemaOrArrayAnnotation = AnnotationsUtils.mergeSchemaAnnotations(annotatedType.getCtxAnnotations(), type);
         final io.swagger.v3.oas.annotations.media.Schema resolvedSchemaAnnotation =
                 resolvedSchemaOrArrayAnnotation == null ?
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
@@ -63,7 +63,7 @@ public enum PrimitiveType {
         }
         @Override
         public Schema createProperty31() {
-            return new JsonSchema().typesItem("string").format("byte");
+            return new JsonSchema().typesItem("integer").format("int8");
         }
     },
     BINARY(Byte.class, "binary") {
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
@@ -63,7 +63,7 @@ public enum PrimitiveType {
         }
         @Override
         public Schema createProperty31() {
-            return new JsonSchema().typesItem("integer").format("int8");
+            return new JsonSchema().typesItem("integer");
         }
     },
     BINARY(Byte.class, "binary") {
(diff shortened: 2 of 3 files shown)


ewaostrowska:
Hi @jongmin-chung!
Thank you for pointing this out. It does seem like mapping a numeric byte to integer/int8 would be conceptually correct for OAS 3.1 and is a change we should consider. 

I have added this issue to our backlog to do some further investigations. 

Some aspects to be checked:
- The PR’s test asserts that @Schema(type="string", format="byte") on a byte field still yields integer/int8. If teams relied on forcing a base64 string for historical reasons, this is a behavior change.
- We need to consider interactions with tools such as linters, codegens and validators which may warn on int8 format. int8 it’s not one of the core formats from the spec text.
- We need to test the behavior of byte arrays

jongminchung:
Hi @ewaostrowska 

> [!NOTE]
> It's a library that a lot of people use, so it took me a while to understand the code with my research. 
> Please understand. I wanted to comment a faster review. 

<br/>

> The PR’s test asserts that @Schema(type="string", format="byte") on a byte field still yields integer/int8. If teams relied on forcing a base64 string for historical reasons, this is a behavior change.

Thanks for the feedback. Referring to the OpenAPI 3.1 specification ([link](https://swagger.io/specification/#:~:text=type%3A%20string%0AcontentMediaType%3A%20image/png%0AcontentEncoding%3A%20base64)), I noticed that contentEncoding: base64 is applied on top of a type: string.

From that perspective, it seems more consistent that a Java primitive byte should map to integer, rather than being represented as a base64 string. Personally, I think base64 semantics belong to String types only.

Could you please confirm if mapping Java byte → base64 string was intended as a team rule, or if moving to integer/int8 for OAS 3.1 would be acceptable? Happy to adjust the PR once the direction is clear.

<br/> <br/>

> We need to consider interactions with tools such as linters, codegens and validators which may warn on int8 format. int8 it’s not one of the core formats from the spec text.

It was something that I also thought about. — int8 is not one of the core formats listed in the OAS 3.1 specification, and using it may trigger warnings in linters, codegens, or validators. To avoid introducing a non-standard format, I will remove the format: int8 and simply use type: integer. ([7e4544a](https://github.com/swagger-api/swagger-core/pull/4954/commits/7e4544a4cf8b8c4a757e12560aaed4d2a23f1bb6))

For cases where we want to reflect Java’s byte range (-128 to 127), OAS 3.1 allows expressing that via minimum and maximum constraints on the integer, rather than a dedicated format.

```java
# Alternative
public Schema createProperty31() {
    // OAS 3.1: use integer with explicit range for Java byte (-128..127), no non-standard format
    return new JsonSchema()
            .typesItem("integer")
            .minimum(java.math.BigDecimal.valueOf(-128))
            .maximum(java.math.BigDecimal.valueOf(127));
}
```

<br/> <br/>

> We need to test the behavior of byte arrays

As you said, when I tested the byte arrays, they changed in the same way that byte is converted to OAS. ([Link](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java#L345-L346))

I will raise PR to revise the test code and code as below change. I would appreciate it if you could check about the team background. Please also check the official OAS3.1 document I referred to.

I tried to match the I add test, but OAS2, OAS3_1 backward compatibility, so it's not a code I like 😂, but I wrote the results and tests I wanted. ([eab93d7](https://github.com/swagger-api/swagger-core/pull/4954/commits/eab93d7eb5f5c9b5d70dd956fa38550ac6b68bd2))

```yaml
# byte
byte:
  type: integer

# byte[]
arrayOfByte:
  type: array
    items: 
      type: integer
```

Issue #4967, opened 2025-09-09
[Bug]: RequiredMode.NOT_REQUIRED is not respected by properties annotated with @NotBlank and @NotEmpty


AnderData:
## Description of the problem/issue

Using `RequiredMode.NOT_REQUIRED` does not correctly override the inferred requirement of fields annotationed with `@NotBlank` or `@NotEmpty`.

For example, these two fields are currently incorrectly marked as required in the generated OpenAPI file:

public class Sample {
  @Schema(requiredMode = RequiredMode.NOT_REQUIRED)
  @NotBlank String exampleString;

  @Schema(requiredMode = RequiredMode.NOT_REQUIRED)
  @NotEmpty List<String> exampleList;
}

Before https://github.com/swagger-api/swagger-core/issues/4862 these fields were correctly being overridden to be not required.

I was able to confirm the before/after behavior by adding those two fields to the file:
`swagger-core/src/test/java/io/swagger/v3/core/oas/models/RequiredFields.java`
and the associated unit test `testRequiredProperty()` in:
`modules/swagger-core/src/test/java/io/swagger/v3/core/converting/override/ModelPropertyOverrideTest.java`

## Affected Version

Started in 2.2.30
Introduced by #4862 

## Steps to Reproduce

Can be completely reproduced by updating the `swagger-core/src/test/java/io/swagger/v3/core/oas/models/RequiredFields.java` and add two additional fields as described above in the `Sample` class.

## Expected Behavior

Fields annotated by `@NotBlank` or `@NotEmpty` should be not required in the openapi doc if also annotated by `@Schema(requiredMode = RequiredMode.NOT_REQUIRED)`.

## Actual Behavior

Fields annotated by `@NotBlank` or `@NotEmpty` are always required in the openapi regardless of `@Schema` annotation configuration.

## Checklist

- [X] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) and this is not a duplicate.
- [X] I have provided sufficient information for maintainers to reproduce the issue.

PR #4968, opened 2025-09-09
fix: NotBlank and NotEmpty annotations follow NotNull behavior

AnderData:
## Description
Fixes: #4967 

## Type of Change
- [X] 🐛 Bug fix

## Checklist
- [X] I have added/updated tests as needed
- [X] I have added/updated documentation where applicable
- [X] The PR title is descriptive
- [X] The code builds and passes tests locally
- [X] I have linked related issues (if any)

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -1804,7 +1804,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
             }
         }
 
-        if (annos.containsKey("javax.validation.constraints.NotEmpty")) {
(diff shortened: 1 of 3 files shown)


daniel-kmiecik:
@AnderData The added tests look good overall, but they don’t seem to cover all relevant combinations of annotations and types. It would be helpful to include additional cases to make sure the new behavior is fully validated.

Specifically, please consider adding tests for:

- @NotNull combined with @NotBlank / @NotEmpty (to confirm they still produce required = true);

- Collection types (List, Set) annotated with @NotEmpty, verifying that they generate appropriate schema constraints (e.g. minItems);

- String fields annotated with both @NotBlank and @Size, ensuring that both restrictions are reflected in the resulting schema.

This would improve confidence that the change doesn’t introduce regressions or subtle differences in how these annotations interact.

ewaostrowska:
@AnderData thank you for the contribution!

Issue #4965, opened 2025-09-02
[Bug]: [swagger-core-jakarta] ModelConverterContextImpl does not properly cache AnnotatedType


anteperic:
## Description of the problem/issue

Based on analysis, `swagger-core-jakarta` does not properly cache (store) already processed `AnnotatedType`s. 

For instance, Java `String` type is always recalculated as it is not found in:
`private final Set<AnnotatedType> processedTypes;`

Root cause is probably faulty `equals()` or `hashcode()` implementation in the `AnnotatedType.

## Affected Version
Discovered in 2.2.27

## Steps to Reproduce

A simple code modification of the `io.swagger.v3.core.converter.ModelConverterContextImpl` can showcase that the same types are "recalculated" See "CONVERTERS" part.

```java
    @Override
    public Schema resolve(AnnotatedType type) {

        AnnotatedType aType = ReferenceTypeUtils.unwrapReference(type);
        if (aType != null) {
            return resolve(aType);
        }

        if (processedTypes.contains(type)) {
            return modelByType.get(type);
        } else {
            processedTypes.add(type);
        }
        Iterator<ModelConverter> converters = this.getConverters();
        Schema resolved = null;
        if (converters.hasNext()) {
            var start = System.currentTimeMillis();
            ModelConverter converter = converters.next();
            LOGGER.info("[CONVERTERS] {}", type.getType().getTypeName());
            resolved = converter.resolve(type, this, converters);
            var duration = System.currentTimeMillis() - start;
            LOGGER.info("[CONVERTERS] {} took {}ms", type.getType().getTypeName(), duration);
        }

        if (resolved != null) {
            modelByType.put(type, resolved);

            Schema resolvedImpl = resolved;
            if (resolvedImpl.getName() != null) {
                modelByName.put(resolvedImpl.getName(), resolved);
            }
        } else {
            processedTypes.remove(type);
        }

        return resolved;
    }

```

## Expected Behavior & Actual Behavior

Swagger `Schema` for repeating `AnnotatedType`s should not be resolved by converter. Instead, value should be fetched from existing `modelByType`. 

## Logs / Stack Traces

This is an example from our machine. Note: some specifics are omitted. 

2025-09-02T17:43:21.124 INFO qtp1695504428-376 io.swagger.v3.core.converter.ModelConverterContextImpl - [CONVERTERS] [simple type, class java.lang.String]
[...]

PR #4975, opened 2025-09-18
Fix(core): Prevent redundant schema resolution by fixing AnnotatedType equality

juntae6942:
…e equality

# Pull Request

---

Description
This pull request resolves a performance issue where the ModelConverterContext fails to cache AnnotatedType instances, leading to redundant schema resolutions.

The root cause was a faulty equals() and hashCode() implementation in the AnnotatedType class. These methods incorrectly included contextual fields (e.g., propertyName), causing logically identical types to be treated as different objects by the cache (processedTypes Set). This bug resulted in the same types being processed multiple times during schema generation, causing significant performance degradation.

This commit corrects the equals() and hashCode() implementations to only consider fields that define a type's intrinsic identity, ensuring the cache functions as expected.

Closes: #4965

## Type of Change

- [ x] 🐛 Bug fix
- [ ] ✨ New feature
- [ x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

Test Case Explanation:

A new test, AnnotatedTypeCachingTest.java, has been added to reproduce the bug and verify the fix.

Why a DummyModelConverter is used: The standard ModelResolver has internal optimizations for simple types like String, which prevents it from recursively calling the ModelConverterContext. This interferes with testing the context's caching behavior. To solve this, the test uses a DummyModelConverter that simulates the recursive resolution of a model in a controlled environment, allowing us to test the caching logic in isolation.

How it Works:

Before the fix: The test fails because the faulty equals() implementation causes multiple AnnotatedType instances for the String type to be added to the cache. The assertion assertEquals(stringTypeCount, 1) fails because the actual count is greater than 1.

After the fix: The test passes because the corrected equals() implementation ensures that only one AnnotatedType instance for the String type is stored in the cache. The assertion assertEquals(stringTypeCount, 1) succeeds.

This test provides clear, automated proof that the caching issue is resolved.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/AnnotatedType.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/AnnotatedType.java
@@ -243,47 +243,28 @@ public class AnnotatedType {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) {
-            return true;
-        }
-        if (!(o instanceof AnnotatedType)) {
-            return false;
-        }
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
         AnnotatedType that = (AnnotatedType) o;
-
-        if ((type == null && that.type != null) || (type != null && that.type == null)) {
-            return false;
-        }
-
-        if (type != null && that.type != null && !type.equals(that.type)) {
-            return false;
-        }
-        return Arrays.equals(this.ctxAnnotations, that.ctxAnnotations);
+        return skipOverride == that.skipOverride &&
+                schemaProperty == that.schemaProperty &&
+                resolveAsRef == that.resolveAsRef &&
+                resolveEnumAsRef == that.resolveEnumAsRef &&
+                includePropertiesWithoutJSONView == that.includePropertiesWithoutJSONView &&
+                skipSchemaName == that.skipSchemaName &&
+                skipJsonIdentity == that.skipJsonIdentity &&
+                Objects.equals(type, that.type) &&
+                Objects.equals(name, that.name) &&
+                Arrays.equals(ctxAnnotations, that.ctxAnnotations) &&
+                Objects.equals(jsonViewAnnotation, that.jsonViewAnnotation);
     }
-
 
     @Override
     public int hashCode() {
-        if (ctxAnnotations == null || ctxAnnotations.length == 0) {
-            return Objects.hash(type, "fixed");
-        }
-        List<Annotation> meaningfulAnnotations = new ArrayList<>();
-
-        boolean hasDifference = false;
-        for (Annotation a: ctxAnnotations) {
-            if(!a.annotationType().getName().startsWith("sun") && !a.annotationType().getName().startsWith("jdk")) {
-                meaningfulAnnotations.add(a);
-            } else {
-                hasDifference = true;
-            }
-        }
-        int result = 1;
-        result = 31 * result + (type == null ? 0 : Objects.hash(type, "fixed"));
-        if (hasDifference) {
-            result = 31 * result + meaningfulAnnotations.hashCode();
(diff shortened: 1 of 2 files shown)


juntae6942:
Hi [ewaostrowska](https://github.com/ewaostrowska), thank you for your time and effort maintaining this project.
I’d really appreciate it if you could take a look at this PR whenever you have a chance 🙏

ewaostrowska:
Hi @juntae6942, 
thanks a ton for opening the PR and digging into this! 
Your change does fix the redundant schema resolutions and there are less random cache misses. Common types (such as Strings) won't be re-resolved over and over again. The provided tests are a great documentation of the problem. 

There are however few cases that could be addressed in this area:

1. **Order-sensitive annotations**
`ctx = {@A, @B}  vs  {@B, @A}   // NOT equal → extra resolution`
equals uses Arrays.equals(ctxAnnotations, …), which is order-sensitive. If callers produce the same effective set of annotations in different orders, there will still be missed cache hits.

2. **JDK/internal annotation noise**
If one call site picks up a jdk.*/sun.* annotation and another doesn’t, the arrays differ → not equal (even if hashCode tries to ignore them).
hashCode tries to ignore sun.*/jdk.*, but equals still compares the raw arrays, would be nice to have these two aligned

3. **Mutability hazard**
ctxAnnotations is used in equality but is mutable; changing it after insertion into a Set can corrupt lookups.
ctxAnnotations participates in equality, and it’s an array stored on the object; if it’s mutated after insertion into a Set, lookups can break. 

4. **getClass() != o.getClass() disallows subclass equality**. 
At the moment we don not inherit but it’s a behavioral change vs instanceof and would be good to have a final keyword on the AnnotatedType if we make a change. Currently I think it's better to leave it with instanceOf

5. Adding few tests to the cases above and toggling fields excluded from equality (name, resolveAsRef, etc.)


We really appreciate your contribution, and we’d be happy to see an updated PR with these extras. Are you interested in taking it further?

juntae6942:
[ewaostrowska](https://github.com/ewaostrowska) Thanks for the detailed feedback! Yes, I'm happy to take this further. I'll look into the points you raised and update the PR.

juntae6942:
Hi [ewaostrowska](https://github.com/ewaostrowska),
I would be grateful if you could review my changes and let me know if the approach I took is appropriate.

ewaostrowska:
hi @juntae6942, thanks for taking it further
I think the approach should fix cache fragmentation from the contextual fields, subclass equality, jdk annotations and addresses the mutability hazard which is awesome :) It does still need some work on aligning the hash and equal mismatch. Also some clean up would be necessary. Thanks for adding more tests as well

juntae6942:
Hi [ewaostrowska](https://github.com/ewaostrowska), 
Thanks again for your feedback.

I've pushed a new commit that refactors the equals() and hashCode() methods in AnnotatedType. Based on your comments, I've simplified them to only include fields that define the type's core identity for caching: type, ctxAnnotations, and jsonViewAnnotation.

Contextual fields like name and resolveAsRef have been removed from the equality check to prevent cache fragmentation.

Could you please let me know if this direction aligns with what you had in mind? If I've misinterpreted your feedback, I would appreciate any further guidance on the correct approach.

Thanks for your help!

ewaostrowska:
Hi @juntae6942, 
Thanks for providing further improved PR and your patience! I have approved it and it will be applied with the next release

benjaminknauer:
Unfortunately this PR introduces a regression for polymorphic schemas; reverting dd5ce5445d74e2173446f5b6238f7f3d42bf58e6 on top of `2.2.39` or the current master commit at 24ade8e678032a3b7f433a278e2b65fbf927f9e6 (`2.2.40-SNAPSHOT`) brings back the expected behavior.

I’ve documented the details and a small reproduction in issue #5003.

PR #4983, opened 2025-10-04
feat: Infer discriminator mapping

### LINKED ISSUE #3411: Consider the @JsonSubTypes for generating the discriminator mapping
To generate the appropriate mapping for a discriminator of a parent class, we are required to add the discriminatorMapping property to @Schema, which is almost always redundant if the class already has a @JsonSubTypes annotation.
As the discriminatorProperty is read from the @JsonTypeInfo.property can the discriminatorMappings also be read from the JsonSubTypes if available?

As a workaround, I have registered a CustomModelResolver and extended the method resolveDiscriminator as follows,

```java
@Override
protected Discriminator resolveDiscriminator(JavaType type, ModelConverterContext context) {
	Discriminator discriminator = super.resolveDiscriminator(type, context);
	if (discriminator != null && discriminator.getPropertyName() != null &&
			(discriminator.getMapping() == null || discriminator.getMapping().isEmpty())) {
		JsonSubTypes jsonSubTypes = type.getRawClass().getDeclaredAnnotation(JsonSubTypes.class);
		if (jsonSubTypes != null) {
			Arrays.stream(jsonSubTypes.value()).forEach(subtype -> {
				discriminator.mapping(subtype.name(), RefUtils.constructRef(
						context.resolve(new AnnotatedType().type(subtype.value())).getName()));
			});
		}
	}
	return discriminator;
}
```

**Additional Info**: Using the swagger-maven-plugin v2.1.1 for generating the OAS files during the compile phase.

PatrickFeiring:
## Description

This PR adds inference of discriminator mappings when an explicit mapping is not present, which should greatly reduce the need for explicit `@DiscriminatorMapping` annotations and make it easier to keep the mapping in sync.

The basic theme of the approach is to leave as much resolving to Jackson as possible. This is done by instantiating a type serializer for the class, and letting Jackson compute the ids used in the mapping. This means that type ids from all the following sources are supported

 - name field in `@JsonSubTypes.Type` annotations
 - `@JsonTypeName` annotations
 - Mechanisms given by `JsonTypeInfo.Id` in `@JsonTypeInfo` annotations

It also means we support things such as default property names when no property value is set on the `@JsonTypeInfo` annotation (`@type`) and the fallback to `JsonTypeInfo.As.PROPERTY` when `JsonTypeInfo.As.EXTERNAL_PROPERTY` is used on a class.

It tries to be conservative and add only a discriminator mapping when it is applicable, e.g. not for `JsonTypeInfo.As.WRAPPER_ARRAY`. It keeps the old behaviour of using an explicit discriminator mapping if present.

The main `resolveDiscriminator` do have a new signature, which is a breaking change for those subclassing and overriding. The new signature should be sensible as it matches that of `resolveSubtypes`, and resolving subtypes is similar to determining the discriminator mappings. It is tricky to keep this backwards compatible while still making it handle the more general cases, but migration path is super easy and hopefully most overriding is not needed any more. The beanDesc can be recomputed, but getting the jsonViewAnnotation is trickier.

I've also tested that it works on the Jackson 2.8 version. Despite the readme mentioning support from 2.4 and newer, the model resolver uses functionality in Jackson only available from 2.8, so I assumed that sufficed.

Closes: https://github.com/swagger-api/swagger-core/issues/3411

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -16,8 +16,10 @@ import com.fasterxml.jackson.databind.AnnotationIntrospector;
 import com.fasterxml.jackson.databind.BeanDescription;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.PropertyMetadata;
+import com.fasterxml.jackson.databind.SerializationConfig;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.introspect.Annotated;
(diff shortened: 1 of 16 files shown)


ldt-fweichert:
Does this implementation also automatically generate `oneOf` with the jackson models subtypes?

Issue #4966, opened 2025-09-05
[Feature]: Provide a bom module for managing dependencies


fengwuxp:
## Feature Description
It is convenient for developers to manage dependencies uniformly through BOM, especially in scenarios with dependency conflicts in complex projects.

## Use Case
example
```xml
<dependency>
         <groupId>io.swagger.core.v3</groupId>
         <artifactId>swagger-bom</artifactId>
         <version>${swagger-openapiv3.version}</version>
         <type>pom</type>
         <scope>import</scope>
</dependency>
```

jochenberger:
See #4824

ewaostrowska:
The BOM module has been just merged :) https://github.com/swagger-api/swagger-core/pull/4987. Hopefully it will allow us to avoid issues like the recent one with the #defaultSentinel

PR #4987, opened 2025-10-07
[Feature]: Provide a BOM module for managing dependencies (#4966)

kuntal1461:
Fixes: https://github.com/swagger-api/swagger-core/issues/4966

- **`modules/swagger-bom/pom.xml`** *(new)* — Standalone BOM POM (not parented by the reactor) that lists all `io.swagger.core.v3` runtime artifacts for both the `javax` and Jakarta families under `<dependencyManagement>`, pinned to `${project.version}`. Includes a `release` profile with GPG signing and Sonatype Central publishing.

- **`modules/swagger-bom-integration-test/pom.xml`** *(new)* — Reactor module that imports `swagger-bom` and declares every managed artifact *without* explicit versions. The `maven-dependency-plugin:resolve` goal runs at `validate` phase and fails the build if any artifact cannot be resolved, giving an automated regression test for the BOM.

- **`pom.xml`** — Registers both new modules in the reactor `<modules>` list.

- **`.github/workflows/maven.yml` / `maven-pulls.yml`** — Added a pre-build step that extracts the root reactor version and the BOM version and fails CI if they differ (drift-guard).

- **`CI/prepare-release.sh` / `post-release.sh`** — Added `mvn versions:set` calls for `modules/swagger-bom` so the BOM version is bumped in lock-step with the rest of the project during releases.

- **`README.md`** — New "Maven BOM" section with Maven `<dependencyManagement>` and Gradle `platform()` usage snippets, and an explanation of why build plugins are intentionally excluded from the BOM.


daniel-kmiecik:
@kuntal1461 we really appreciate your work and commitment!
Please check your PR and remove unnecessary changes. Leave only the part concerning BOM module.

kuntal1461:
Hi @daniel-kmiecik  Please check once . and please let know anything I need to updated

kuntal1461:
Hi @daniel-kmiecik 
Please check once and let me know if there’s anything I need to update.
I haven’t received any review comments yet, so just wanted to confirm if any changes are pending from my side.

mymx2:
align all versions, include `swagger-annotations-jakarta`.

see: https://search.maven.org/search?q=io.swagger.core.v3

ewaostrowska:
Hi @kuntal1461, thank you for your input in the topic! I introduced few changes here as I feel like this PR did not cover the whole issue:
  - `modules/swagger-bom/pom.xml `- Added 7 Jakarta artifacts to <dependencyManagement>: swagger-annotations-jakarta, swagger-models-jakarta, swagger-core-jakarta, swagger-integration-jakarta, swagger-jaxrs2-jakarta,
  swagger-jaxrs2-servlet-initializer-jakarta, swagger-jaxrs2-servlet-initializer-v2-jakarta. - as pointed out by @mymx2 (thanks!)
- CI version-drift guard.` github/workflows/maven-pulls.yml`, `.github/workflows/maven.yml` , `CI/prepare-release.sh`, `CI/post-release.sh` - Added a step that extracts the root version and the BOM version via the exec plugin and fails if they differ. `.github/workflows/maven.yml`

michaldo:
It seems swagger-bom is not published to Maven central in release 2.2.47

ewaostrowska:
Hi @michaldo,
I double checked it and it seems that the bom for both [2.2.47](https://repo1.maven.org/maven2/io/swagger/core/v3/swagger-bom/2.2.47/) and [2.2.48 ](https://repo1.maven.org/maven2/io/swagger/core/v3/swagger-bom/2.2.48/) is available in the maven central. The bom is published separately from the parent, which from what I have read is the standard way of publishing the bom to avoid inheriting anything it shouldn't. 

Please let me know if this answers your question.

michaldo:
You are right, my fault

I verified with
`mvn dependency:get -Dartifact=io.swagger.core.v3:swagger-bom:2.2.47`
instead of
`mvn dependency:get -Dartifact=io.swagger.core.v3:swagger-bom:2.2.47:pom`

ewaostrowska:
Thanks for the confirmation!

PR #4989, opened 2025-10-07
Fix: primitive boxed type schemas incorrectly cast to "string" in OAS 3.1 #4963

### LINKED ISSUE #4963: [Bug]: Primitive boxed type schemas used in @ApiResponse content schema implementations are incorrectly cast to string

When generating the specs for OpenAPI 3.1 for a JAX-RS resource, there is an issue with `@ApiResponse` object that uses primitive boxed types for the content schema implementation. When resolving the schema, although the `types` field in the schema object (`io.swagger.v3.oas.models.media.Schema`) is correct, `type` field is always `string`.

## Affected Version
2.2.31 and above

Earliest version the bug appears in:  
2.2.31 (introduced in [PR 4868](https://github.com/swagger-api/swagger-core/pull/4868))

## Steps to Reproduce

```
@Test
public void testApiResponses31() {
    SwaggerConfiguration config = new SwaggerConfiguration().openAPI31(true);
    Reader reader = new Reader(config);
    OpenAPI openAPI = reader.read(ApiResponsesResource.class);

    Operation postOperation = openAPI.getPaths().get("/mypath").getPost();
    Schema responseSchema = postOperation.getResponses().get("200").getContent().get("*/*").getSchema();

    assertEquals(responseSchema.getType(), "string"); // Incorrect type
    assertEquals(responseSchema.getTypes().iterator().next(), "boolean"); // Actual type
}
```
with `ApiResponsesResource` class being:
```
@Path("mypath")
public class ApiResponsesResource {

    @POST
    @ApiResponses({
            @ApiResponse(responseCode = "200", content = @Content(schema = @Schema(implementation = Boolean.class)))
    })
    public void myMethod() { }
}
```

## Expected Behavior

The `type` value should either be `null` or it should match the value in `types`.

## Actual Behavior

`type` is always set to `string` (in `AnnotationsUtils#947`).

## Proposed solution

The if condition surrounding the setting of the type to string in `AnnotationsUtils#947` should also take `types` into consideration, so something like:
```
        if (StringUtils.isBlank(existingSchemaObject.get$ref()) && StringUtils.isBlank(existingSchemaObject.getType()) && 
                (existingSchemaObject.getTypes() == null || existingSchemaObject.getTypes().isEmpty())) {
```

## Checklist

K5qu4r3d:
# Pull Request

Fixes: https://github.com/swagger-api/swagger-core/issues/4963

When generating OpenAPI 3.1 specs, `@ApiResponse` annotations that reference primitive boxed types (e.g. `Boolean.class`) as schema implementations produce a schema where the OAS 3.1 `types` set is correctly populated but `type` is then overwritten to `"string"`. This causes invalid specs and incorrect client generation.

---

## What was broken

`AnnotationsUtils` falls back to `setType("string")` when a schema has no `$ref` and no `type`. It did not check `getTypes()`, so OAS 3.1 schemas that already had the correct `types` set were silently clobbered.

## What changed

- `AnnotationsUtils.java`: added a `doesSchemaHaveNullOrEmptyTypes` guard to the `setType("string")` fallback so schemas already carrying an OAS 3.1 `types` value are not overwritten.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

- Added files:
    - `modules/swagger-jaxrs2/src/test/java/io/swagger/v3/jaxrs2/resources/APIResponsesResource.java`
    - `modules/swagger-jaxrs2/src/test/java/io/swagger/v3/jaxrs2/APIResponsesResourceTest.java`

- Affected files:
    - `modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java`
    - `modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java`
    - `modules/swagger-java17-support/src/test/java/io/swagger/v3/java17/resolving/v31/ModelResolverOAS31Test.java`
    - `modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/Ticket2992Test.java`


modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
```diff
@@ -374,6 +375,9 @@ public Schema createProperty31() {
         addKeys(names, OBJECT, "object");
         NAMES = Collections.unmodifiableMap(names);
 
+        // Add "LocalTime" class mappings
+        enablePartialTime();
```
ewaostrowska: I think this is not a part of the solution to the https://github.com/swagger-api/swagger-core/issues/4963. Please consider creating a feature request for this change and describing why this change should be introduced. Some analysis would be necessary to confirm it's align with JSON Schema.

For now please remove it

modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
```diff
@@ -951,7 +951,10 @@ public static Schema resolveSchemaFromType(Class<?> schemaImplementation, Compon
                 existingSchemaObject = reResolvedSchema.get();
             }
         }
-        if (StringUtils.isBlank(existingSchemaObject.get$ref()) && StringUtils.isBlank(existingSchemaObject.getType())) {
+        boolean doesSchemaHaveTypes = existingSchemaObject.getTypes() == null || existingSchemaObject.getTypes().isEmpty();
```
ewaostrowska: The name of the method here is misleading. It returns true when empty/ null. Please rename it

ewaostrowska (changes requested):

ewaostrowska (approved):

ewaostrowska:
There are still some failing tests in `APIResponsesResourceTest.java`. Please adjust tests :)

ewaostrowska:
Hi @K5qu4r3d!
The fix in the AnnotationsUtils.java seems to be minimal and correct - the check on the types which is OAS 3.1 specific has been missing and adding it should restore setting the right types for this version of specification. 
 I have rolled back all the changes in PrimitiveType.class since : 
 - it would cause regression on https://github.com/swagger-api/swagger-core/pull/5034
 - we do not want to set the type, which is OAS 3.0 specific, on schema for OAS 3.1. It is supposed to use types.

Issue #4852, opened 2025-03-06
StackOverflowError with recursive data model introduced with version 2.2.28


Yaytay:
My project builds correctly using 
  io.swagger.core.v3:swagger-jaxrs2-jakarta:2.2.27
  io.swagger.core.v3:swagger-annotations-jakarta:2.2.27
as can be seen here: https://github.com/Yaytay/query-engine/actions/runs/13693736285

Bumping to 
  io.swagger.core.v3:swagger-jaxrs2-jakarta:2.2.28
  io.swagger.core.v3:swagger-annotations-jakarta:2.2.28

causes a StackOverflowError:
https://github.com/Yaytay/query-engine/actions/runs/13693747582/job/38291540555#step:8:93660
```
uk.co.spudsoft.query.main.OpenApiModelConverter.resolve@69
io.swagger.v3.core.converter.ModelConverterContextImpl.resolve@97
io.swagger.v3.core.jackson.ModelResolver.resolve@745
uk.co.spudsoft.query.main.OpenApiModelConverter.resolve@70
io.swagger.v3.core.converter.ModelConverterContextImpl.resolve@97
io.swagger.v3.core.jackson.ModelResolver.resolve@493
uk.co.spudsoft.query.main.OpenApiModelConverter.resolve@70
io.swagger.v3.core.converter.ModelConverterContextImpl.resolve@97
io.swagger.v3.core.jackson.ModelResolver.resolve@745
uk.co.spudsoft.query.main.OpenApiModelConverter.resolve@70
io.swagger.v3.core.converter.ModelConverterContextImpl.resolve@97
io.swagger.v3.core.jackson.ModelResolver.resolve@745
uk.co.spudsoft.query.main.OpenApiModelConverter.resolve@70
io.swagger.v3.core.converter.ModelConverterContextImpl.resolve@97
io.swagger.v3.core.jackson.ModelResolver.resolveSubtypes@1665
io.swagger.v3.core.jackson.ModelResolver.resolve@906
uk.co.spudsoft.query.main.OpenApiModelConverter.resolve@70
io.swagger.v3.core.converter.ModelConverterContextImpl.resolve@97
io.swagger.v3.core.jackson.ModelResolver.resolve@745
uk.co.spudsoft.query.main.OpenApiModelConverter.resolve@70
io.swagger.v3.core.converter.ModelConverterContextImpl.resolve@97
io.swagger.v3.core.jackson.ModelResolver.resolve@745
[...]

PR #5004, opened 2025-10-24
#4852 Unit test and fix to allow recursive models to process without StackOverflow 

Yaytay:
# Pull Request

## Description

Bug fix ton undo a tiny part of https://github.com/swagger-api/swagger-core/compare/v2.2.27...v2.2.28#diff-bed6aab5e29942b1a575bded333604f9dfaeb2ef9475f1433266ece4e49f4fcb that introduced a StackOverflow exception on recursive data models with OpenAPI 3.1.

Fixes: #4852

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,8 @@ The OpenAPI Specification has undergone several revisions since initial creation
 
 Swagger core Version      | Release Date | OpenAPI Spec compatibility | Notes | Status
 ------------------------- | ------------ | -------------------------- | ----- | ----
-2.2.39 (**current stable**)| 2025-10-13   | 3.x           | [tag v2.2.39](https://github.com/swagger-api/swagger-core/tree/v2.2.39) | Supported
+2.2.40 (**current stable**)| 2025-10-28   | 3.x           | [tag v2.2.40](https://github.com/swagger-api/swagger-core/tree/v2.2.40) | Supported
+2.2.39                     | 2025-10-13   | 3.x           | [tag v2.2.39](https://github.com/swagger-api/swagger-core/tree/v2.2.39) | Supported
 2.2.38                     | 2025-09-29   | 3.x           | [tag v2.2.38](https://github.com/swagger-api/swagger-core/tree/v2.2.38) | Supported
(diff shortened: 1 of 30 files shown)


ewaostrowska:
hi @Yaytay ,
Thanks for providing this PR. It does help with the issue reported in the https://github.com/swagger-api/swagger-core/issues/4852 and pointing this spot is already quite helpful for us. Unfortunately we cannot merge this PR yet. 

One issue that I see here is that the test ReaderTest.testArraySchemaImplementation is failing with 
```
expected [openapi: 3.1.0
paths:
      items:
        type: number
]
```
This indicates that items are not resolved correctly after the change. 
We need to investigate further on why this boolean property has been added to AnnotationsUtils.getArraySchema. 

That extra true argument in 2.2.28 wasn't random — it was almost certainly introduced to fix some other array-handling edge case in the OpenAPI 3.1 path (for example, making sure annotated @ArraySchema metadata such as minItems, uniqueItems, description, proper items resolution, etc., is actually reflected in the generated schema for arrays of complex/polymorphic types). 

We need to verify as well how does it influence the OAS 3.x schema as well. 

Thanks again for your contribution! We're happy to see any updates from you on this topic :)

Yaytay:
My apologies for making this a PR when the unit tests fail - I didn't realise the tests in the other module were affected, entirely my bad.

What I don't understand is why there would be any circumstances in which you'd want to reset the ModelConverterContext while processing a single model.
Actually there's a lot I don't understand about swagger-core :)

I've updated the PR so that it tracks the AnnotatedTypes that it is currently resolving.
This enables it to explicitly pass "false" in my circumstances, but otherwise retain the "true" argument.
At least locally this makes all tests in swagger-project pass.

ewaostrowska:
@Yaytay, 
Thanks for introducing the changes to the PR! The solution looks better now and it also the test `ReaderTest.testArraySchemaImplementation `is not failing anymore. I have done some clean up both in tests and ModelResolver, but the core of the logic was correct and solves the issue without reintroducing [regression](https://github.com/swagger-api/swagger-core/pull/4823). And yes, logic of swagger-core and especially ModelResolver can be quite challenging to understand :)

Thanks for your contribution!

Issue #5003, opened 2025-10-23
[Bug]: Regression: AnnotatedType equality change drops polymorphic subtype allOf entries


benjaminknauer:
## Description of the problem/issue

After upgrading to `swagger-core` `2.2.39`, polymorphic schemas that rely on subtype resolution produce incomplete `allOf` definitions.

### Expected Schema

```json
"ConcretionResponseA": {
    "type": "object",
    "allOf": [
        { "$ref": "#/components/schemas/BaseResponse" },
        {
            "type": "object",
            "properties": {
                "concretionAField": { "type": "string" }
            }
        }
    ]
}
```

### Actual Schema (Buggy)

```json
"ConcretionResponseA": {
    "allOf": [
        { "$ref": "#/components/schemas/BaseResponse" }
    ]
}
```

### Possible cause

The regression stems from commit dd5ce5445d74e2173446f5b6238f7f3d42bf58e6, which normalizes context annotations inside `AnnotatedType.equals()`/`hashCode()` but still ignores flags such as schemaProperty and propertyName. As a result, property and subtype resolutions now collapse into the same cache entry and the subtype-specific fragment is skipped.

In the attached example Spring Boot project the type `ConcretionResponseA` is first resolved as a property and later as a subtype; the second pass now reuses the cached property schema, so the subtype-specific properties never reach the generated OpenAPI document.

## Affected Version

`>=  2.2.39`

Earliest version the bug appears in: `2.2.39` (regression introduced by commit dd5ce5445d74e2173446f5b6238f7f3d42bf58e6)

## Steps to Reproduce
Example Project: https://github.com/benjaminknauer/swagger-2-2-39-polymorphism-bug-example

1. clone the example project
2. `mvn clean test`
3. Inspect the failure of `BuggySchemaIntegrationTest` (compares `/api-docs` response with expected schema). 

The serialized schema will show `ConcretionResponseA` with only a single `allOf` entry.

## Expected Behavior

`ConcretionResponseA` should contain two `allOf` entries: the `$ref` to `BaseResponse` and a second object with the subtype’s own property (`concretionAField`).

## Actual Behavior

`ConcretionResponseA` only contains the `$ref` to `BaseResponse`; the object with `concretionAField` is missing because subtype resolution returns the cached result from the earlier property resolution.

juntae6942:
[benjaminknauer](https://github.com/benjaminknauer)
Thanks for reporting this issue!
I’ll take a look and submit a fix soon.

[...]

PR #5005, opened 2025-10-24
Fix/annotated type caching 5003

juntae6942:
Closes: #5003 

Description
This PR fixes a regression bug that causes an AnnotatedType cache collision when a polymorphic type is resolved in two different contexts: once as a 'property' (e.g., direct return type) and once as a 'subtype' (e.g., from @JsonSubTypes).

The regression was introduced when equals()/hashCode() began normalizing context annotations but did not properly account for flags like schemaProperty or propertyName. This caused both 'property' and 'subtype' resolutions to merge into the same cache entry, leading to an incomplete schema (the subtype-specific fragment was skipped).

This fix addresses the issue in two parts:

AnnotatedType.java: equals() and hashCode() are updated to correctly compare the schemaProperty flag. The propertyName is now also conditionally compared only when schemaProperty is true, aligning equals and hashCode logic.

ModelResolver.java: resolveSubtypes() is updated to explicitly set schemaProperty(true) when creating an AnnotatedType for a class discovered via @JsonSubTypes.

This combination ensures that the 'property' context (schemaProperty=false) and 'subtype' context (schemaProperty=true) generate distinct cache keys, resolving the collision.

Fixes: ## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

Checklist

Screenshots / Additional Context
This fix can be verified using the reproduction project provided in the bug report: https://github.com/benjaminknauer/swagger-2-2-39-polymorphism-bug-example

With this patch applied, the BuggySchemaIntegrationTest in that project passes successfully.

The new unit tests added to AnnotatedTypeTest.java specifically validate the new equals/hashCode logic for all schemaProperty and propertyName scenarios to prevent future regressions.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/AnnotatedType.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/AnnotatedType.java
@@ -29,6 +29,7 @@ public class AnnotatedType {
     private boolean skipSchemaName;
     private boolean skipJsonIdentity;
     private String propertyName;
+    private boolean isSubtype;
 
     private Components components;
 
@@ -243,6 +244,19 @@ public class AnnotatedType {
         return this;
     }
 
+    public boolean isSubtype() {
+        return isSubtype;
+    }
+
+    public void setSubtype(boolean isSubtype) {
+        this.isSubtype = isSubtype;
+    }
+
+    public AnnotatedType subtype(boolean isSubtype) {
+        this.isSubtype = isSubtype;
+        return this;
+    }
+
     private List<Annotation> getProcessedAnnotations(Annotation[] annotations) {
         if (annotations == null || annotations.length == 0) {
             return new ArrayList<>();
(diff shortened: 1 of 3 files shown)


ewaostrowska:
Hi @juntae6942, 

Thank you for your quick response to this issue :) we do appreciate it a lot. 

This should fix the problem of polymorphic schemas caching. 
However, for now I cannot accept this PR because of the side effects it may cause. Setting schemaProperty(true) in the ModelResolver is the part that can be troublesome.  That flag may influence not only equality/hashCode, but also conditional logic elsewhere that checks isSchemaProperty() to decide how to decorate the resolved schema. There’s a risk of formatting differences in the composed schema. I have added a test to illustrate this issue. 

There are few approaches that could be taken here, but I need decision from the team. Please hold on for now with the topic. I will inform you whenever there's any decision.

Thanks again for your contribution!

ewaostrowska:
@juntae6942, 
I think introducing a isSubtype variable here instead of setting schemaProperty(true) for subtypes should resolve this issue.

juntae6942:
Hi [ewaostrowska](https://github.com/ewaostrowska),

Ah, I see you've already pushed a commit using the isSubtype flag. That makes perfect sense, and the approach looks good! 👍

Thanks for taking care of the fix!

Is there anything else needed from my end for this?

ewaostrowska:
@juntae6942 no, I don't think anything anything more from your side is needed :) thanks for contributing!

juntae6942:
[ewaostrowska](https://github.com/ewaostrowska), Thanks for the quick feedback! I really appreciate it.
I’d love to continue contributing to Swagger in the future as well. 😊

Issue #4993, opened 2025-10-11
[Bug]: Regression: class specified in @ArraySchema.schema.implementation not rendered in schemas


arungitan:
## Description of the problem/issue

we are using Springdoc with the swagger-core dependency. 
we were using `@ArraySchema` successfully in Springdoc 2.6 (default 3.0 openapi output from swagger) but once we moved to springdoc 2.8 (default 3.1 openapi output from swagger), the implementation class specified in `@ArraySchema.schema` is not rendered at all.  
So the swagger UI fails with errors like: 
```
"Resolver error at responses.200.content.*/*.schema.items.$ref
Could not resolve reference: Invalid object key "ThingAssignment" at position 2 in "/components/schemas/ThingAssignment": key not found in object"
```

## Affected Version

2.2.36 

Earliest version the bug appears in (if known):   not known  

## Steps to Reproduce

See attached zip file with app and notes. (The sample uses 2.2.36 but I also tried 2.2.38 with the same failure)

## Expected Behavior

Should render the ThingAssignment in the Schemas section and emit the reference to array of ThingAssignment for the consuming class (User.java in the sample).

## Actual Behavior

(see notes from zip)
Emitted the array reference but does not render the implementation class schema (ThingAssignment)
In prior versions with 3.0 output format, it was working fine. 

## Logs / Stack Traces

no logs but see notes in zip for the emitted output. 

## Additional Context

- By overriding the output format from the default 3.1 to 3.0, the problem goes away. `springdoc.api-docs.version=openapi_3_0`
- This swagger-core issue looks related but not sure if it fixes the problem for this issue as well, and if so, when will the fix be released? https://github.com/swagger-api/swagger-core/issues/4679 
- issue originally raised against springdoc:  
   https://github.com/springdoc/springdoc-openapi/issues/3097

## Checklist

- [ x] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) and this is not a duplicate.
- [ x] I have provided sufficient information for maintainers to reproduce the issue.

arungitan:
Sample app and notes:
[swagger-core-bug-sample.zip](https://github.com/user-attachments/files/22860474/swagger-core-bug-sample.zip)

Mattias-Sehlstedt:
[...]

Issue #4800, opened 2024-11-28
BUG @Schema annotation is propagated from a property to the class defined in @Schema#implementation 


OllieKosh:
### Problem
When `Schema#implementation` is defined on a property, all information specified through `@Schema` is propagated to the schema created for the type defined in `Schema#implementation`. However this information should only be used when generating the property schema. I am using Open API 3.1. This problem was introduced after upgrading from 2.2.22 to 2.2.25.

### Expected behavior
`@Schema` annotation is processed on the property level, but not used in schema generation for the class specified in `Schema#implementation` of that property.

For example if I have the following DTO
```
public class Example {
    @Schema(implementation = MyEnum.class, description = "My description", defaultValue = "yes")
    private String whyYes;
    //getters, setters
}
``` 
and MyEnum class looks like this
```
enum MyEnum {
    yes, no
}
```
the generated contract for the two components is expected to look like this
```
  schemas:
    Example:
      type: object
      properties:
        whyYes:
          description: My description
          $ref: "#/components/schemas/MyEnum"
          default: "yes"
    MyEnum:
      type: `string`
      enum:
      - "yes"
      - "no"
```
However current behavior propagates the schema annotation (and the properties from that schema annotation, in this case `description` and `defaultValue`) to the underlying object, which is incorrect
```
  schemas:
    Example:
      type: object
      properties:
        whyYes:
          description: My description
          $ref: "#/components/schemas/MyEnum"
          default: "yes"
    MyEnum:
      type: string
      description: My description
      default: "yes"
      enum:
      - "yes"
      - "no"
```
Note that if `@Schema#implementation` was not specified this problem would not occur.

### Reproducer
Add something similar to the provided example above (if using enum, set enums to be resolved as references) to this [test class](https://github.com/swagger-api/swagger-core/blob/c37b72cc653c41dc1ab90a4e10d2c957c84d52d4/modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/Ticket4679Test.java) (which was created to test the solutions for this [issue](https://github.com/swagger-api/swagger-core/issues/4679)).

### Investigation
[...]

PR #5006, opened 2025-10-27
Issue 4993

ewaostrowska:
Fixes: #4993 

The gap in resolving types of arraySchema implementation was caused in fix for prevent leaking properties to enum schema when Schema#implementation is set. (Issue #4800). To prevent leaking properties to enum schema the logic which copies the ctxAnnotations when type is overriden with implementation was deleted. Instead a more narrow solution of assigning type from schema had been implemented. However, this solution did not address assigning type from implementation for ArraySchema.


ewaostrowska:
Needs further analysis

Issue #4999, opened 2025-10-18
[Bug]: Example value starting with a number is parsed incorrectly


Sagr-gupta:
**Description**

When an example in an OpenAPI annotation (e.g., @Schema(example = "5 lacs per annum")) starts with a number, it appears that Swagger Core interprets the value as a numeric literal because the example begins with a digit. This results in the example being serialized as a number instead of a string.

For instance : 

`@Schema(example = "5 lacs per annum")
private String salary;`

In the generated OpenAPI schema, the example becomes: 

`"example": 5`

instead of : 

`"example": "5 lacs per annum"`

**Where the issue occurs**

The issue originates from the following code in io.swagger.v3.core.util.AnnotationsUtils#getSchemaFromAnnotation:

<img width="866" height="301" alt="Image" src="https://github.com/user-attachments/assets/14b37517-267c-4c7c-8f17-70f45187af22" />

Because Json31.mapper().readTree(...) interprets any string starting with a digit as JSON numeric content,
the portion "lacs per annum" is ignored and only the number 5 is retained.

**Affected Version**

swagger-core-jakarta: 2.2.36

springdoc-openapi: 2.8.13

For reproducing the issue below I have attached a demo project. Kindly take reference.

[openapi.zip](https://github.com/user-attachments/files/22986552/openapi.zip)

zero2top:
Hi! 👋 I'd like to work on this issue if it's not already being handled.
It seems the problem comes from `AnnotationsUtils#getSchemaFromAnnotation`, 
where `Json31.mapper().readTree(...)` parses a numeric string as a number.

I plan to:
- Add a conditional check to treat examples starting with a digit as strings.
- Add a test case in `AnnotationsUtilsTest` to verify this behavior.

Please let me know if it's okay for me to proceed. Thanks!

daniel-kmiecik:
@zero2top We encourage you to prepare a fix for this issue. Once you have a pull request ready, we will review it and work with you on any feedback and final merge.

Thanks for your contribution!

zero2top:
Hi! 👋

I’ve completed the fix for this issue.

### ✔️ What I changed
- Updated `AnnotationsUtils#getSchemaFromAnnotation` to ensure example values starting with digits are always treated as strings.
- Added a test case `testExampleStartingWithNumberShouldBeString` in `AnnotationsUtilsTest` to validate this behavior.

### 📌 Status
The fix is implemented and tested locally — I’m preparing the pull request now.

[...]

PR #5021, opened 2025-11-21
fix: Treat example values starting with numbers as strings (#4999)

zero2top:
Pull Request

Thank you for contributing to swagger-core!

Description

This PR fixes an issue where example values that begin with a digit (e.g., "5 lacs per annum") are incorrectly parsed as a numeric literal instead of a string.

The root cause was that Json31.mapper().readTree(...) interprets any string that starts with a digit as JSON numeric content, resulting in the loss of the remaining string.

What this PR changes

Updated AnnotationsUtils#getSchemaFromAnnotation to ensure that example values beginning with digits are always treated as strings.

Added a test case testExampleStartingWithNumberShouldBeString to verify the expected behavior.

Fixes: #4999

Type of Change

-   Bug fix

-   Tests

Checklist

1.  I have added/updated tests as needed
2. The PR title is descriptive
3. The code builds and passes tests locally
4. I have linked the related issue (#4999)
 
 
 Example before the fix:

"example": 5

After the fix:

"example": "5 lacs per annum"

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/override/CustomResolverTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/override/CustomResolverTest.java
@@ -48,12 +48,12 @@ public class CustomResolverTest {
         Schema model = models.get("SuperFoo");
         assertNotNull(model);
         assertEquals(model.getProperties().size(), 4);
-        assertEquals(model.getRequired(), (Collections.singletonList("bar")));
+        assertEquals(model.getRequired(), Collections.singletonList("bar"));
 
         final Schema fooProperty = (Schema) model.getProperties().get("foo");
         assertEquals(fooProperty.get$ref(), "#/components/schemas/Foo");
         Schema fooModel = models.get("Foo");
-        assertEquals(fooModel.getRequired(), (Collections.singletonList("bar")));
+        assertEquals(fooModel.getRequired(), Collections.singletonList("bar"));
 
         final Schema optionalBarProperty = (Schema) model.getProperties().get("optionalBar");
         assertEquals(optionalBarProperty.get$ref(), "#/components/schemas/Bar");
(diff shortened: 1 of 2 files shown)


zero2top:
Hi! 👋  
The PR is ready for review.  
Could you please approve and run the workflows when you have a moment?  
The required checks are currently waiting for maintainer approval.  
Thanks!

modules/swagger-core/src/test/java/io/swagger/v3/core/util/AnnotationsUtilsTest.java
```diff
@@ -96,6 +100,108 @@ private void emailType() {
...
+        };
+
+        io.swagger.v3.oas.models.media.Schema<?> schema =
+                AnnotationsUtils.getSchemaFromAnnotation(annotation, null, null, false, null).get();
+
+        assertNotNull(schema);
+        assertEquals(schema.getExample(), "5 lacs per annum");
+    }
 
     class DummyClass implements Serializable {}
```
daniel-kmiecik: Please do not split existing code into specific tests.

modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
```diff
@@ -765,16 +765,24 @@ public static Optional<Schema> getSchemaFromAnnotation(
...
         if (StringUtils.isNotBlank(schema.example())) {
-            try {
-                if (openapi31) {
-                    schemaObject.setExample(Json31.mapper().readTree(schema.example()));
-                } else {
-                    schemaObject.setExample(Json.mapper().readTree(schema.example()));
+            String exampleValue = schema.example().trim();
+
+            // Fix: Prevent numeric-starting example strings (e.g. "5 lacs per annum") from being parsed as numbers
+            if (exampleValue.matches("^[0-9].*") && !exampleValue.startsWith("\"")) {
```
Mattias-Sehlstedt: Defining `.*` in the `"^[0-9].*"` regex seems superfluous/wasteful to me, since that means that the engine will check all characters that are after the initial number before it determines that the string starts with a number. Something akin to `"^[0-9]"` or `"^\d"` should terminate "faster".

daniel-kmiecik:
@zero2top I have moved your changes to our PR  (#5046) due to the lack of response from you.
I will mark you as a co-author of changes.
Thank you for your contribution. We appreciate it!

Issue #4682, opened 2024-06-03
BUG properties of type Object do not have any type information when generating open api 3.1.0


Lorenzo-Bracci:
## Problem
When generating OpenAPI 3.1.0 spec, properties with the java type `Object` do not have any type information in the generated open api but are instead empty. As an example assume the following java model:
```
private static class PojoUsingObjectField {
        private Object myField;

        public Object getMyField() {
            return myField;
        }

        public void setMyField(Object myField) {
            this.myField = myField;
        }
}
```
When generating open api for 3.0.1 the generated model would be:
```
PojoUsingObjectField:
  type: object
  properties:
    myField:
      type: object
```
But when generating open api 3.1.0 the generated model is:
```
PojoUsingObjectField:
  type: object
  properties:
    myField: {}
```

## Expected behaviour 
Same as in open api 3.0.1, properties have a `type` field with value `object`.

## Reproducer
The issues is reproduced in the following test (note that the test is successful if the model resolvers generate OAS3.0.1 instead):

```
    @Test(description = "Shows that type information in properties using a raw Object is lost when using oas 3.1.0")
    public void testModelUsingObjectTypedPropertyLosesTypeInformationForOas31() {

        String expectedYaml = "PojoUsingObjectField:\n" +
                "  type: object\n" +
                "  properties:\n" +
                "    myField:\n" +
                "      type: object";

        Map<String, io.swagger.v3.oas.models.media.Schema> stringSchemaMap = ModelConverters.getInstance(true).readAll(PojoUsingObjectField.class);
        // fails as the type: object will not be added for oas 3.1.0, instead we will have "myField: {}"
        SerializationMatchers.assertEqualsToYaml31(stringSchemaMap, expectedYaml);
    }

    private static class PojoUsingObjectField {
        private Object myField;

        public Object getMyField() {
            return myField;
        }

        public void setMyField(Object myField) {
            this.myField = myField;
        }
    }
```

## Investigation 

[...]

PR #5034, opened 2026-01-10
fix: OAS 3.1 schema generation for raw Object properties

xeulbn:
# Pull Request

---

## Description

Related issue: #4682
This PR fixes an issue in OpenAPI 3.1 schema generation where properties
declared as raw Java Object were rendered as an empty schema ({}).

In the OAS 3.1 path, PrimitiveType.OBJECT#createProperty31() returned an empty
JsonSchema without populating Schema#types. Since OpenAPI 3.1 relies on
JSON Schema semantics and uses types during serialization, the missing type
information caused the schema to be emitted as {}.

To resolve this, the OAS 3.1 schema generation for raw Object types has been
updated to explicitly set types = ["object"].

Additionally, regression tests have been added to cover:
- raw `Object` properties
- `List<Object>` item schemas
- `Map<String, Object>` additionalProperties schemas

Fixes: #4682 

## Type of Change
- [x] 🐛 Bug fix
- [x] 🧪 Tests

## Checklist

The tests were executed locally with:
```bash
mvn -pl modules/swagger-core -Dtest=ObjectFieldOas31ReproTest test
```

### Before (OAS 3.1)

Before

### After (OAS 3.1)

After

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
@@ -679,11 +679,6 @@ public abstract class AnnotationsUtils {
 
         if (schema.types().length > 0) {
             schemaObject.setTypes(new LinkedHashSet<>(Arrays.asList(schema.types())));
-            if (schema.types().length == 1) {
-                schemaObject.setType(schema.types()[0]);
-            } else {
-                schemaObject.setType(null);
-            }
         }
         if (StringUtils.isNotBlank(schema.$id())) {
             schemaObject.set$id(schema.$id());


daniel-kmiecik:
@xeulbn please also check for errors in builds and code scanning jobs.

xeulbn:
Sorry I missed the comment before the build test.
I’ll check the build results and update the commit accordingly.

xeulbn:
The build failure was caused by the updated OAS 3.1 handling for raw `Object`, which changes the expected YAML output in `ModelResolverOAS31Test`:

- `status` type: [string, number] → [object, string, number]
- `extraObject`: {} → { type: object }

I updated the OAS 3.1 expected output accordingly.

Verified locally on JDK 17 with:
mvn -pl modules/swagger-core test

All tests passed.
Thanks for the review and the helpful feedback!

xeulbn:
Another build failure was caused by the updated OAS 3.1 handling for raw Object, which changes the expected YAML output in several OAS 3.1 test resources and reader tests.

Empty object schemas are now consistently rendered as `type: object`, and union types inferred from raw `Object` include `object` in addition to the previously inferred primitive types.

I updated the affected OAS 3.1 YAML expectations and reader tests accordingly.

xeulbn:
In OpenAPI 3.1, `type` may be a union (`Schema#types`).

Previously, inferred types (e.g. `object` from raw Java `Object`) were merged with
explicit `@Schema(types=...)`, which could unintentionally widen the declared union.

While looking into this, I considered an explicit `@Schema(types=...)` declaration
as an intention to fully define the allowed types, rather than to merge them with
inferred defaults. Based on that interpretation, this change ignores inferred types
when `types` is explicitly provided, while preserving inferred `object` when it is not.

Tests were updated accordingly.

daniel-kmiecik (changes requested):

daniel-kmiecik (approved):

daniel-kmiecik:
@xeulbn thank you for your PR and for taking the time to contribute to the community. Your effort are truly appreciated!

xeulbn:
Thank you for the review and guidance throughout the process.

I appreciate the detailed feedback — it helped me better understand the differences between OAS 3.0 and 3.1 schema handling.
Looking forward to contributing more in the future!

Mattias-Sehlstedt:
Hi, the changes introduced in this PR seem to conflict with what has previously controlled with the explicit setting "explicit-object-schema", which was introduced and described in https://github.com/swagger-api/swagger-core/pull/4868#issue-2963063228.

Further, now a Map<String, Object> becomes
```json
"additionalProperties": {
  "type": "object"
 }
```
Rather than the previous
```json
"additionalProperties": {}
```
I can see from the tests that this is intentional. But to my knowledge this new schema expresses a new type of restriction on the request. For example the request
```json
"key": "value"
```
Is now an invalid request, since `"value"` is not a JSON-object. 

So the addition of `"type": "object"` more aligns with a Java representation that is `Map<String, Map<String, Object>>`.

Would it be considered correct if I investigated how to best revert the change for a additionalProperties field?

xeulbn:
Thanks for raising this.

I think that is a fair point. My intention with this change was to improve how raw `Object` is represented in OAS 3.1, especially to avoid cases where it was rendered as an empty schema unintentionally.

That said, I agree `additionalProperties` may be a special case. For `Map<String, Object>`, changing

`"additionalProperties": {}`
to
`"additionalProperties": { "type": "object" }`

can indeed narrow the allowed values and change the schema semantics in a way that is closer to `Map<String, Map<String, Object>>`.

So I would not see this as invalidating the overall change, but I do think it makes sense to investigate reverting or refining the behavior specifically for `additionalProperties`.

Issue #5041, opened 2026-01-21
[Feature]: I want to configure the default value of @Schema RequiredMode to other value, than AUTO


VOJAKIS:
## Feature Description

I want to configure the RequiredMode for annotation @Schema programatically to REQUIRED. Right now, most of the attributes we use are required, like 95%, and other are not required.

## Use Case

When I am making an application with backend and frontend, most of the times the frontend wants to see every attribute as a response from the backend.
Right now we have to configure that (or use another annotations, like @NotNull) to do that.

## Suggested Solution (optional)

There should be a way to overload the default value to a specified value, when not stated otherwise.

If a value is declared in the annotation, e.g.
`@Schema(requiredMode = Schema.RequiredMode.NOT_REQUIRED`
then Swagger should ignore the default value and use the one declared.
Otherwise use the default value.

## Alternatives Considered

I have tried to do this with Spring Boot annotation @NotNull, OpenAPICustomizer and a custom meta-annotation.

## Additional Context

I use Spring Boot v4.

SpringDOC OpenAPI Starter WebMVC UI v2.8.10

## Checklist

- [x] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) to ensure this is not a duplicate.
- [x] This feature would be useful to more than just my use case.
- [x] I have provided enough detail for the maintainers to understand the scope of the request.

Mattias-Sehlstedt:
Hi, I believe that the easiest way of achieving this together with springdoc-openapi is by doing [this](https://github.com/springdoc/springdoc-openapi/issues/2926#issuecomment-2692842442).

VOJAKIS:
> Hi, I believe that the easiest way of achieving this together with springdoc-openapi is by doing [this](https://github.com/springdoc/springdoc-openapi/issues/2926#issuecomment-2692842442).

Thank you for your reply. I used this https://github.com/springdoc/springdoc-openapi/issues/2992#issuecomment-2873939361

VOJAKIS:
This removes the option for an attribute to retain it's "null" type.

This schema:
`@Schema(types = { "string", "null" }, nullable = true) String fileReference`
or this one
`@Schema(types = { "string", "null" }) String fileReference`

both produce this in Swagger:
<img width="306" height="34" alt="Image" src="https://github.com/user-attachments/assets/409135ac-745c-49b5-9429-712d3e1b917a" />

<br>

I want to produce this:
[...]

PR #5042, opened 2026-01-22
refactor: add an overloaded method for resolving required so that the field type can be utilized

Mattias-Sehlstedt:
# Pull Request

---

## Description

TL;DR:
Makes it possible to easily determine the required mode for a property based upon the class type. This is done by introducing an overloaded method `resolveRequiredMode` that also takes a `JavaType`, and then making the default behavior that the application passes the already accessible `propType` to the method.

----

This is a subject I have been asked about before both in [swagger-core](https://github.com/swagger-api/swagger-core/issues/5041) but also in [springdoc-openapi](https://github.com/springdoc/springdoc-openapi/issues/2926#issuecomment-2692842442).

There exists good documentation for how to achieve the change for the default for all (define a new `ModelResolver` and override the  `resolveRequiredMode` metod). But if we wanted to control it based upon the field/property type, we would have to override the entire `ModelResolver` and its 1000+ lines `resolve`-method. 

With this change we can rather go back to only overriding a small, clearly defined, part of the class to achieve the behavior. I have illustrated how this can be achieved in a test.

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist


daniel-kmiecik:
@Mattias-Sehlstedt great work. Thank you for your contribution!

PR #5044, opened 2026-01-26
feat: add support for Positive, PositiveOrZero, Negative, and NegativeOrZero constraints in model validation

NaMinhyeok:
## Description

- Add support for Bean Validation sign constraints: @Positive, @PositiveOrZero, @Negative, @NegativeOrZero.
- Map them to OpenAPI schema bounds (minimum/maximum + exclusive*).

Fixes: #5035 

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] ♻️ Refactor (non-breaking change)
- [ ] 🧪 Tests
- [ ] 📝 Documentation
- [ ] 🧹 Chore (build or tooling)

## Checklist

- [x] I have added/updated tests as needed
- [ ] I have added/updated documentation where applicable
- [x] The PR title is descriptive
- [x] The code builds and passes tests locally
- [x] I have linked related issues (if any)

PR #5047, opened 2026-01-30
refactor: introduce class responsible for how bean validation annotations affect schemas

Mattias-Sehlstedt:
# Pull Request

---

## Description

This is a step towards exposing an easier way of extending how `BeanValidatorAnnotations` can be introspected and added to the generated schema. It is a [sometimes reoccurring request](https://github.com/swagger-api/swagger-core/pull/5044) to add different annotations, and sometimes custom ones too. The current main way of extending it is to override two large methods (see links in third paragraph).

I would find beneficial if there was an isolated method that was responsible for applying bean validations, so that clients could then only overload that. The end result being something similar to how springdoc offers this type of [customization functionality](https://github.com/springdoc/springdoc-openapi/issues/3180#issuecomment-3693299126).

Currently the annotation -> contraint logic is defined in two locations, depending on whether validation groups are [in effect](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L1779) or [not](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L1990). An illustration of this can be seen in the [PR that introduces additional validations](https://github.com/swagger-api/swagger-core/pull/5044/changes).

Fixes: 

Centralizes the definitions of how a bean validation annotation should affect a schema's constraints into a single location/class. Currently it is defined in two locations, where it is also coupled to the logic that determines whether the constrains should be added or not to the schema (depending on validation groups).

This PR decouples them and places the definitions in a separate component (ValidationAnnotationsUtils) and lets the logic for application then invoke that and ask for the definitions if necessary.

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/util/ValidationAnnotationsUtilsTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/util/ValidationAnnotationsUtilsTest.java
@@ -1,0 +1,588 @@
+package io.swagger.v3.core.util;
+
+import io.swagger.v3.oas.models.media.ArraySchema;
+import io.swagger.v3.oas.models.media.IntegerSchema;
+import io.swagger.v3.oas.models.media.NumberSchema;
+import io.swagger.v3.oas.models.media.ObjectSchema;
+import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.oas.models.media.StringSchema;
+import org.testng.annotations.Test;
+
+import javax.validation.constraints.*;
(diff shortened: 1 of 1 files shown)


ewaostrowska:
Hi @Mattias-Sehlstedt thank you for committing your time to make the swagger code cleaner :)

Issue #5051, opened 2026-02-05
[Bug]: Array schema description leaks into items


jochenberger:
I have an annotation like this
```java
 @ApiResponse(
      responseCode = "201",
      content = @Content(
              mediaType = MediaType.APPLICATION_JSON_VALUE,
              array =
                  @ArraySchema(
                          arraySchema = @Schema(description = "Array description"),
                      schema =
                          @Schema(
                              description = "Item description",
                              implementation = Item.class
                              )))))
List<Item> doSomething();
```
The generated schema looks like this:
```
[
  type:array,
  description: Array description,
  items: [
    $ref:#/components/schemas/Item,
    description:Array description
  ]
]
```
The array description overrides the item description.

If I remove the `implementation = Item.class`, I get
```
[
  type:array,
  description: Array description,
  items: [
    type:string,
    description:Item description
  ]
]
```
The item description is there, but the type is wrong. IIUC, It should be deduced from the method signature.

If I remove the whole  `schema = ...` part, I get
```
[
  type:array,
  description: Array description
]
```
The items section is missing completely. At least the type info should be there.

I remember there were similar issues in the *components* schema generation. This concerns the API response schemas. I assumed they were using the same logic.

Mattias-Sehlstedt:
Hi @jochenberger, could we specify whether you are resolving to a 3.1 or a 3.0 OpenAPI specification? I believe there are certain issues that are specifically tied to 3.1, and this sound very similar to that kind of issue.

jochenberger:
Yes, I'm generating 3.1 with the latest version. Sorry, I should have included that in the description right away.

Mattias-Sehlstedt:
It is most likely tied to the issue [discussed here](https://github.com/swagger-api/swagger-core/pull/5048), were the introspection is overeager with trying to handle potential OAS 3.1 siblings, and thus branch into a path that does not produce the same result as 3.0.

I could investigate once my my [current fix](https://github.com/swagger-api/swagger-core/pull/5049) on the area are reviewed and merged.

jochenberger:
I can test once a new version including your fix is released.

Mattias-Sehlstedt:
I have investigated the issue a bit further and to me it seems like the current intended behavior of `content = @Content(...)` is to allow for a complete redefinition of what the endpoint serves. I.e., if annotations are present, those and those alone are used to generate the OpenApi-specification. And if they are omitted it deduces the types as per usual. 

[...]

PR #5055, opened 2026-02-16
fix: calculate sibling properties based upon whether the schema is the array or the item in the array

Mattias-Sehlstedt:
# Pull Request

---

## Description

Fixes: #5051

I have formalized the [extra schema-annotation check that is done for a 3.1. specification](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java#L972). This check is done to properly add sibling properties (e.g., adding a description together with the `#ref`).

We now only expect one of the three different annotation scenarios to have an effect (being `schemaAnnotation`, `schemaAnnotation `within the `arrayAnnotation`, or the `arrayAnnotation` itself). The basis for this being that they are mutually exclusive. We either have a `Schema`-annotation or an `ArraySchema`-annotation. The second scenario was not handled before, but is now considered and supported.

This change prevents the data leakage that occurred before due to the `arraySchema`-annotation being applied to the object within the array.

We can also see that this fixed a data leakage issue that occurred in the test for [ModelWithArrayOfSubclasses](https://github.com/swagger-api/swagger-core/compare/master...Mattias-Sehlstedt:swagger-core:array-list-schema-in-response-content?expand=1#diff-3540483c20908e194940b191071b7eeae6b012f738797ef7186366d0a225ac93), where the list properties ("description", "minItems", "uniqueItems") where also placed on the list object too.

This:
```json
"baseArray" : {
  "type" : "array",
  "description" : "Thingy",
    "items" : {
      "$ref" : "#/components/schemas/Base",
      "description" : "Thingy",
      "minItems" : 0,
      "uniqueItems" : true
    },
  "minItems" : 0,
  "uniqueItems" : true
}
```
Should be this
```json
"baseArray" : {
  "type" : "array",
  "description" : "Thingy",
    "items" : {
      "$ref" : "#/components/schemas/Base"
    },
  "minItems" : 0,
  "uniqueItems" : true
}
```
Since the annotation is:
```java
@ArraySchema(
  schema = @Schema(implementation = Base.class), 
  arraySchema = @Schema(
    type = "array", 
    description = "Thingy"), 
  minItems = 0, 
  uniqueItems = true
)
public Base[] getBaseArray() {
```

I also took the opportunity to adjust the test introduced with https://github.com/Mattias-Sehlstedt/swagger-core/commit/0a5556a4226d9b09742bf7f09909e787df1ed556 so that the expected json-definitions are placed in the resource folder like how it is with most tests.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/Issue5055Test.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/Issue5055Test.java
@@ -1,0 +1,285 @@
+package io.swagger.v3.core.converting;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.swagger.v3.core.converter.ModelConverters;
+import io.swagger.v3.core.converter.ResolvedSchema;
+import io.swagger.v3.core.util.Json31;
+import io.swagger.v3.oas.annotations.media.ArraySchema;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+import static org.testng.Assert.*;
+
+/**
+ * test documenting the behavior of sibling @Schema and @ArraySchema annotations.
+ */
+public class Issue5055Test {
+
+
+    @Test
+    public void testArrayMetadataDoesNotLeakToItemsRef() throws Exception {
+        ResolvedSchema schema = ModelConverters.getInstance(true).readAllAsResolvedSchema(
+            TestModels.ArrayWithFullMetadata.class
+        );
+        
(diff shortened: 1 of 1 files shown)


ewaostrowska:
Hi @Mattias-Sehlstedt! 
Thank you for investing your time in this topic here. We do appreciate it a lot. I have added tests to fully document the behavior and test the edge cases. but otherwise looks good :)

Issue #5013, opened 2025-11-04
[Bug]: @ArraysSchema applied on a stream property results in object type in OAS 3.1


martin-rais-brcm:
@ArraysSchema applied on a stream property results in object type in OAS 3.1

## Description of the problem/issue
We are using the springdoc with the swagger-core dependency. Once we switched from OpenApi 3.0 to 3.1 the `Stream` property of a POJO model is described as object instead of an array even if the @ArraySchema is applied to that property.

## Affected Version

2.2.40

Earliest version the bug appears in (if known): we found it in 2.2.36.

## Steps to Reproduce

Use this POJO model with a `Stream` property, while generating OAS 3.1:

``` java
@Getter
@Setter
public class ModelWithStream {

  @ArraySchema(schema = @Schema(implementation = Greeting.class))
  private Stream<Greeting> greetings;
}
```

## Expected Behavior

The OAS 3.1 should contain this definition in components.schemas. The `greetings` property should be generated as a type array.

```json
"ModelWithStream": {
  "type": "object",
  "properties": {
    "greetings": {
      "type": "array",
      "items": {
        "$ref": "#/components/schemas/Greeting"
      }
    }
  }
}
```

## Actual Behavior

The `greetings` property is generated as a type object with the `Stream.parallel` property exposed in the schema.

```json
"ModelWithStream": {
  "type": "object",
  "properties": {
    "greetings": {
      "type": "object",
      "items": {
        "$ref": "#/components/schemas/Greeting"
      },
      "properties": {
        "parallel": {
          "type": "boolean"
        }
      }
    }
  }
}
```

## Logs / Stack Traces

N/A.

## Additional Context

The expected model is generated when switching springdoc back to generate the OAS 3.0 version.

## Checklist

- [x] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) and this is not a duplicate.
- [x] I have provided sufficient information for maintainers to reproduce the issue.

PR #5064, opened 2026-02-27
fix: treat streams with arraySchema annotation as array type (#5013)

ewaostrowska:
Special handling for Stream types with ArraySchema:
Without this, Stream<T> with @ArraySchema would be incorrectly resolved as "object" type instead of "array" type. Examples where ArraySchema must be preserved:
1. @ArraySchema(schema = @Schema(implementation = Greeting.class)) on Stream<Greeting>   
-> Should produce: type=array, items=$ref to Greeting
2. @ArraySchema(minItems=1, maxItems=10, ...) on Stream<T> 
-> Should preserve array constraints (minItems, maxItems, uniqueItems, etc.)
Without preserving ArraySchema, Stream is treated as generic object instead of iterable collection
 The ArraySchema is needed for the Stream to be recognized as an array, but it doesn't leak to items because the container path filters it out.
 
 Also refactored the filtering of Schema/ArraySchema annotations with siblings to a separate class.

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/override/GenericModelConverter.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/override/GenericModelConverter.java
@@ -11,9 +11,9 @@ import io.swagger.v3.oas.models.media.Schema;
 import java.util.Iterator;
 import java.util.Map.Entry;
 
-public class GericModelConverter extends AbstractModelConverter {
+public class GenericModelConverter extends AbstractModelConverter {
 
-    protected GericModelConverter() {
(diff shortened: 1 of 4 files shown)


modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
Mattias-Sehlstedt: Could we adjust the check in this if to 
`AnnotationsUtils.areSiblingsAllowed(resolvedSchemaResolution, openapi31)`
since that it the centralized definition of this reoccurring check

Issue #5077, opened 2026-03-04
[Bug]: Issues using @Nullable annotations in OpenAPI 3.1 schemata


pmerschmann:
## Description of the problem/issue

The created OpenApiSpec for nullable objects is not correct.

The nullable annotations do not work proper on objects using OpenApi 3.1 schemata, but only on simple attributes (e.g. strings). With the support for nullable and `@Nullable` annotations within #5001 and #4555 this issue should have been solved, but wasn't.

## Affected Version

2.2.44

Earliest version the bug appears in (if known):  
2.2.44

## Steps to Reproduce

Code to reproduce can be found here:
- https://github.com/pmerschmann/swagger-nullability-demo

Simply start the application and check the api under http://localhost:9090/v3/api-docs (for json formatted) or under http://localhost:9090/swagger-ui/index.html for visual appearance via swagger.

## Actual Behavior

For simple objects (like Strings) everything seems to work out just fine:
```
"someString": {
            "type": ["string", "null"]
          },
```

For lists it can be argued that it is correct as well, as the list itself can be null and any of the contained items as well in my example. Was unexpected for me at first honestly, but can be accepted. Yet the question come to my mind: can we somehow manage that the type can be array or null, but the items themselves are non-nullable?
```
"listOfString": {
            "type": ["array", "null"],
            "items": {
              "type": ["string", "null"]
            }
          }
```

For custom classes things break, as the type is not recognized as `object`, but as `null` and only as `null`.
```
"foo": {
  "$ref": "#/components/schemas/Foo"
},

"Foo": {
        "type": "null",
        "properties": {
          "id": {
            "type": "string",
            "format": "uuid"
          },
}
```

## Checklist

- [X] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) and this is not a duplicate.
- [X] I have provided sufficient information for maintainers to reproduce the issue.

Mattias-Sehlstedt:
Hi, I am unable to reproduce the issue with the model getting type `null`. Could you provide a reproducible example?

The issue might be that you are using swagger-core 2.2.44 with sprindoc, while [springdoc only support 2.2.43](https://github.com/springdoc/springdoc-openapi/blob/v3.0.2/pom.xml#L58).

pmerschmann:
[...]

PR #5079, opened 2026-03-05
fix: do not process a nullable annotation for container items

Mattias-Sehlstedt:
# Pull Request

---

## Description

Fixes part of the issue reported in #5077 where the items within a List (container) are affected by a `@Nullable` annotation. I.e.,
```java
@Nullable
List<String> listOfStrings
```
This should only mark the list itself as nullable, while the items would be controller either by an explicit `@ArraySchema` annotation or something similar to 
```java
List<@Nullable String> listOfStrings
```
*Although `@Nullable` does not apply to a `Type` field)

This change is achieved by making sure that `@Nullable` is dropped when annotations are forwarded for a container type field.

The second part of issue #5077 is not handled, where `#ref`:s are not correctly marked as nullable. 

I have intentionally not handled that in this PR since that is a much larger task, since the issue is that there does not seem to exist any framework to support expressing that. I.e., either having 
```yaml
schema:
  allOf:
    - $ref: '#/components/schemas/User'
  nullable: true
```
for OAS 3.0 or 
```yaml
schema:
  oneOf:
    - $ref: '#/components/schemas/User'
    - type: 'null'
```
or
```yaml
schema:
  type: [object, "null"]
  $ref: '#/components/schemas/User'
```
for OAS 3.1.

Fixes: 

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/issues/Issue5077Test.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/issues/Issue5077Test.java
@@ -7,13 +7,22 @@ import io.swagger.v3.core.converter.ResolvedSchema;
 import io.swagger.v3.core.util.Json;
 import io.swagger.v3.core.util.Json31;
 import io.swagger.v3.core.util.ResourceUtils;
+import io.swagger.v3.oas.models.media.ArraySchema;
+import io.swagger.v3.oas.models.media.Schema;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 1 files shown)


ewaostrowska:
@Mattias-Sehlstedt , thanks for fixing this issues. I have added few additional tests, but otherwise this looks good :)

Issue #5111, opened 2026-04-04
[Bug]: A Java raw Object is always resolved to OAS 3.1 object


Mattias-Sehlstedt:
## Description of the problem/issue

With PR https://github.com/swagger-api/swagger-core/pull/5034 the default behavior of how a Java `Object` for OAS 3.1 is resolved was changed from being an empty type definition to always being type `object`. This can cause incorrect schema generation, since `Object` is not necessary a JSON-object (e.g., "string" is a valid Java `Object`, but not a valid JSON-object).

The interpretation of a Java `Object` should instead be tied to `explicit-object-schema`, which was introduced and described in https://github.com/swagger-api/swagger-core/pull/4868. It also correctly states that this behavior should be an explicit opt-in (and also highlights that the current behavior of 3.0 is incorrect, but that it is kept/not corrected due to backwards compatibility concerns).

## Affected Version

2.2.45

Earliest version the bug appears in (if known):  
2.2.44

## Steps to Reproduce

Have a class that have a property with type `Object`. E.g. 
```java
private static class PojoUsingObjectField {
  private Object myField;

  public Object getMyField() {
    return myField;
  }

  public void setMyField(Object myField) {
    this.myField = myField;
  }
}
```

## Expected Behavior

A Java Object should be resolved to an empty type.

## Actual Behavior

A Java Object is resolved to type `object`.

## Checklist

- [x] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) and this is not a duplicate.
- [x] I have provided sufficient information for maintainers to reproduce the issue.

PR #5112, opened 2026-04-04
fix: OAS 3.1 schema generation for raw Object properties is controlled with explicit-object-schema flag

Mattias-Sehlstedt:
## Description

Reverts changes from https://github.com/swagger-api/swagger-core/pull/5034 and restores the default resolving behavior for an OAS 3.1 Java `Object` to an empty type rather than type `object`. 

Adds tests to highlight that such a resolving logic should be achieved by setting the `explicit-object-schema` flag to true.

### OAS 3.1
In recent versions, raw Object values in OAS 3.1 were emitted as:
`type: object`

This behavior has now been reverted.

For OAS 3.1, raw Object is again rendered as an empty schema by default:

`{}`

This applies to: `Object` properties, `List<Object>` item schemas and `Map<String, Object>` value schemas. In **OpenAPI 3.1,** Schema Objects follow **JSON Schema semantics** more closely. An empty schema means “any JSON value is allowed,” which is a better match for Java Object. Type: object means only JSON objects are allowed, which is too restrictive for raw Java Object, since it may also contain: `strings`, `numbers`, `booleans`, `arrays`, `null`

#### explicit-object-schema
To force type: object to be generated: 

`explicit-object-schema=true`

### OAS 3.0
Behavior is unchanged. 

`type: object`

is generated by default. This can be changed by explicitly setting `explicit-object-schema=false`

---
Fixes: #5111

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -3571,7 +3571,8 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
     }
 
     private boolean isExplicitObjectType() {
-        return (openapi31 && Boolean.TRUE.equals(PrimitiveType.explicitObjectType)) ||
-               (!openapi31 && (!Boolean.FALSE.equals(PrimitiveType.explicitObjectType)));
+        return openapi31
+                ? Boolean.TRUE.equals(PrimitiveType.explicitObjectType)
+                : !Boolean.FALSE.equals(PrimitiveType.explicitObjectType);
     }
 }


modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
```diff
@@ -3572,4 +3569,10 @@ private boolean isStreamType(JavaType type) {
                type.getRawClass() != null &&
                java.util.stream.Stream.class.isAssignableFrom(type.getRawClass());
     }
+
+    private boolean isExplicitObjectType() {
+        return openapi31
```
ewaostrowska, line 3574: Just a small logic simplification here :)

ewaostrowska:
Thank you @Mattias-Sehlstedt for taking you time to work on this issue and pushing small readability changes on the occasion as well!

Issue #5091, opened 2026-03-18
[Bug]: Exponential recursive model processing


ldt-fweichert:
At my company we have a model that is a discriminated union of 8 subtypes. I sadly can not share the models or the code they depend on due to an nda. But i will try to explain my best:
Supertype: 8 Subtypes (just an interface)
Subtype1-8: They reference another subtype in at least 2 fields, most of them by collections. Some of the subtypes also reference the supertype by collections.

I have observed exponential growth in processing time, when generating a specification (with springdoc webmvc):
1 Models: 8s
2 Models: 9s
3 Models: 30s
4 Models: 187s
5 Models: 1,973s
6 Models: Didn't have the time to calculate (approx.: 18,000s / 5 hours)
(I actually thought it doesn't work for more than 3 models at first, but it was just taking a lot longer.)

The profiler analysis showed that `ModelResolver.resolve()` was excessively called multiple times for the same type during processing, suggesting that some form of caching could aid the processing time.

The issue happens at version `2.2.41` and `2.2.45`. Interistingly version `2.2.40` directly threw a StackOverflow, this changed likely through #5004, so i guess this is somewhat related.

I am happy to help if you have further questions about the models at my company.

Mattias-Sehlstedt:
In order to be able to investigate I believe it would be necessary for you to provide a minimal working examples, so something that mimics the structure of your model, but both obfuscated and reduced in size. Trying to express anything complex in text will just lead to confusion and guesswork. A concrete reproducible piece of code is needed.

ldt-fweichert:
Thats fair. I was able to reproduce the exponential processing time by [writing a new test](https://github.com/ldt-fweichert/swagger-core/commit/91b2dec22f050c6af9a5a6c7f241640d0daec6eb)

ldt-fweichert:
I've observerd, that `ModelConverterContextImpl` gets instantiated multiple times while resolving a single model, effectively pruning the `ModelConverterContextImpl.processedTypes` which acts as a cache for already resolved types afaict.

The reinstantiation happens from within `AnnotationUtils.resolveSchemaFromType()` through `resolvedSchema = ModelConverters.getInstance(openapi31).readAllAsResolvedSchema(new AnnotatedType().type(schemaImplementation).components(components).jsonViewAnnotation(jsonViewAnnotation));`.
[...]

PR #5114, opened 2026-04-07
fix: AnnotationUtils cache misses due to reinstantiation of ModelConverterContext

ldt-fweichert:
## Description
`ModelConverterContextImpl` stores each already processed type and it's corresponding `Schema` internally. When types are discovered multiple times during resolution, they should be converted only once. Multiple usages of static helper functions on `AnnotationUtils` instantiated a new `ModelConverterContext` instead of reusing the initial one effectivley ignoring the cache of processed types.

This PR adds a `ModelConverterContext context`-Parameter to some `AnnotationUtils` helpers to reuse the instance during resolution. This speeds up resolution, especially with Recursive Models, where resolving Models with 10 subtypes could require at least multiple hours of processing time.

Two questions from my side remain:
- Should `ModelConverterContext` be instantiated inside `AnnotationUtils` at all? If no, the overloads created in this PR should be removed entirely, to ensure correct usage. An additional null-check should be added as well.
- My added test of checking for big differences in timing between models with 2 subtypes and models with 10 subtypes is admittedly not that maintainable. I'd be open to alternative ideas for testing, although it served my use-case pretty well.

Fixes: #5091 

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/ComplexPolymorphicModelTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/converting/ComplexPolymorphicModelTest.java
@@ -6,13 +6,9 @@ import io.swagger.v3.core.converter.ModelConverterContextImpl;
 import io.swagger.v3.core.converter.ModelConverters;
 import io.swagger.v3.core.converter.ResolvedSchema;
 import io.swagger.v3.core.jackson.ModelResolver;
-import io.swagger.v3.core.oas.models.ModelWithArrayOfSubclasses;
 import io.swagger.v3.core.oas.models.ModelWithManySubtypesAndRecursion;
 import io.swagger.v3.oas.models.media.ArraySchema;
 import io.swagger.v3.oas.models.media.Schema;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.testng.AssertJUnit;
 import org.testng.annotations.Test;
 
 import java.lang.reflect.Field;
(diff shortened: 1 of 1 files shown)


ewaostrowska:
Hi @ldt-fweichert, 
Thanks for providing a fix for this issue! I have added a fix for one additional case where the context has also been dropped. Now this PR should fix references by allOf and self-referencing nodes.

ewaostrowska:
About your questions:
- Actually `AnnotationUtils` does import and pass `ModelConverterContext` and accept it as param, but it is not instantiated. So this is fine. About the overloads - there are cases where the context is not passed and caching is not needed. For example in  `getHeader() ` or `getActualSchema()`. 
- I do agree that the test using timing is not a perfect approach here and I have also added tests simply counting how many times was `resolve()` invoked for the same type. This does check if the type has been cached, but not how many times the cache has been hit, but I think this should be good enough. I have also now removed the tests that use timing.

ldt-fweichert:
> I do agree that the test using timing is not a perfect approach here and I have also added tests simply counting how many times was `resolve()` invoked for the same type. This does check if the type has been cached, but not how many times the cache has been hit, but I think this should be good enough. I have also now removed the tests that use timing.

Good approach, this will improve the predictability of the test

Thanks for the quick merge, the feedback and your optimizations, appreciate the effort!

One semi-off-topic question: Under which circumstances do the models need to be referenced explicitly  with an `@ArraySchema` annotation?

PR #5018, opened 2025-11-17
5001: Add support for @Nullable annotations in OpenAPI 3.1 schemas

zdary:
# Pull Request

Thank you for contributing to **swagger-core**!

Please fill out the following information to help us review your PR efficiently.

---

## Description

Enhanced schema generation by recognizing 

- @Nullable annotations as nullable and adding "null" to the type array in OpenAPI 3.1. 
- nullable = true in schema and adding "null" to the type array in OpenAPI 3.1. 

Introduced a test case to validate the behavior and ensure backward compatibility with explicit @Schema annotations.

Fixes:[ #5001](https://github.com/swagger-api/swagger-core/issues/5001) , [#4555](https://github.com/swagger-api/swagger-core/issues/4555)

## Type of Change

- [X] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] ♻️ Refactor (non-breaking change)
- [X] 🧪 Tests
- [ ] 📝 Documentation
- [ ] 🧹 Chore (build or tooling)

## Checklist

- [X] I have added/updated tests as needed
- [ ] I have added/updated documentation where applicable
- [X] The PR title is descriptive
- [X] The code builds and passes tests locally
- [X] I have linked related issues (if any)

PR #5124, opened 2026-04-11
fix: ensure that @Nullable does not incorrectly affect object schemas

### LINKED ISSUE #5115: [Bug]: @Nullable annotation incorrectly affects object and sets the type to null

PR https://github.com/swagger-api/swagger-core/pull/5018 introduced support for resolving a `@Nullable` annotation and having it set the nullability in the respective OAS representation `(nullable: true` or `type: null`).

The implementation currently does not handle `object`-fields correctly, since the nullable is forwarded and retain for schemas that are later used as `ref`. See https://github.com/swagger-api/swagger-core/issues/5077 for additional details. An example is
```java
public static class ModelWithObject {

    @Nullable
    private Model nullableModel;

    private Model model;

    @Nullable
    public Model getNullableModel() {
        return nullableModel;
    }

    public void setNullableModel(@Nullable Model model) {
        this.nullableModel = model;
    }

    public Model getModel() {
        return model;
    }

    public void setModel(Model model) {
        this.model = model;
    }
}

public static class Model {

    private NestedModel nestedModel;

    @Nullable
    private NestedModel nullableNestedModel;

    @Nullable
    private NestedModel2 nullableNestedModel2;

    private NestedModel2 nestedModel2;

    public NestedModel getNestedModel() {
        return nestedModel;
    }

    public void setNestedModel(NestedModel nestedModel) {
        this.nestedModel = nestedModel;
    }

    @Nullable
    public NestedModel getNullableNestedModel() {
        return nullableNestedModel;
    }

    public void setNullableNestedModel(@Nullable NestedModel nullableNestedModel) {
        this.nullableNestedModel = nullableNestedModel;
    }

    @Nullable
    public NestedModel2 getNullableNestedModel2() {
        return nullableNestedModel2;
    }

    public void setNullableNestedModel2(@Nullable NestedModel2 nullableNestedModel2) {
        this.nullableNestedModel2 = nullableNestedModel2;
    }

    public NestedModel2 getNestedModel2() {
        return nestedModel2;
    }

    public void setNestedModel2(NestedModel2 nestedModel2) {
        this.nestedModel2 = nestedModel2;
    }
}

public static class NestedModel {
    private String field;

    public String getField() {
        return field;
    }

    public void setField(String field) {
        this.field = field;
    }
}

public static class NestedModel2 {
    private String field2;

    public String getField2() {
        return field2;
    }

    public void setField2(String field2) {
        this.field2 = field2;
    }
}
```
and it being resolved to:
```json
{
  "schema" : {
    "type" : "object",
    "properties" : {
      "nullableModel" : {
        "$ref" : "#/components/schemas/Model"
      },
      "model" : {
        "$ref" : "#/components/schemas/Model"
      }
    }
  },
  "referencedSchemas" : {
    "Model" : {
      "type" : "object",
      "properties" : {
        "nestedModel" : {
          "$ref" : "#/components/schemas/NestedModel"
        },
        "nullableNestedModel" : {
          "$ref" : "#/components/schemas/NestedModel"
        },
        "nullableNestedModel2" : {
          "$ref" : "#/components/schemas/NestedModel2"
        },
        "nestedModel2" : {
          "$ref" : "#/components/schemas/NestedModel2"
        }
      }
    },
    "ModelWithObject" : {
      "type" : "object",
      "properties" : {
        "nullableModel" : {
          "$ref" : "#/components/schemas/Model"
        },
        "model" : {
          "$ref" : "#/components/schemas/Model"
        }
      }
    },
    "NestedModel" : {
      "type" : "null",      <-- This should not be null
      "properties" : {
        "field" : {
          "type" : "string"
        }
      }
    },
    "NestedModel2" : {
      "type" : "object",
      "properties" : {
        "field2" : {
          "type" : "string"
        }
      }
    }
  }
}
```

## Affected Version

e.g. 2.2.26

e.g. 2.2.24

## Steps to Reproduce

1. ...
2. ...
3. ...

## Expected Behavior

The `null` is not kept for objects and `null` does not override/replace `object`.

## Actual Behavior

The `null` might leak depending on what order that fields (models) are read.

## Checklist

- [ ] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) and this is not a duplicate.
- [ ] I have provided sufficient information for maintainers to reproduce the issue.

Mattias-Sehlstedt:
# Pull Request

---

## Description

Changes so that a `@Nullable` annotation present on a model does not incorrectly set `type: null` for OAS 3.1 or `nullable: true` for OAS 3.0. The nullable is instead dropped entirely. This since the resolving currently does not support an easy way of expressing nullable with `ref`s correctly as per:
OAS3.0
```json
{ "nullable": true, "allOf": [{ "$ref": "#/components/schemas/NestedObject" }] }
```
or
OAS3.1
```json
{ "oneOf": [{ "$ref": "#/components/schemas/NestedObject" }, { "type": "null" }] }
```

Fixes: #5115

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -183,19 +183,9 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
         }
 
         final Annotation resolvedSchemaOrArrayAnnotation = AnnotationsUtils.mergeSchemaAnnotations(annotatedType.getCtxAnnotations(), type);
-        final io.swagger.v3.oas.annotations.media.Schema resolvedSchemaAnnotation =
-                resolvedSchemaOrArrayAnnotation == null ?
-                        null :
-                        resolvedSchemaOrArrayAnnotation instanceof io.swagger.v3.oas.annotations.media.ArraySchema ?
-                                ((io.swagger.v3.oas.annotations.media.ArraySchema) resolvedSchemaOrArrayAnnotation).schema() :
-                                (io.swagger.v3.oas.annotations.media.Schema) resolvedSchemaOrArrayAnnotation;
+        final io.swagger.v3.oas.annotations.media.Schema resolvedSchemaAnnotation = getSchemaAnnotation(resolvedSchemaOrArrayAnnotation);
 
-        final io.swagger.v3.oas.annotations.media.ArraySchema resolvedArrayAnnotation =
-                resolvedSchemaOrArrayAnnotation == null ?
-                        null :
-                        resolvedSchemaOrArrayAnnotation instanceof io.swagger.v3.oas.annotations.media.ArraySchema ?
-                                (io.swagger.v3.oas.annotations.media.ArraySchema) resolvedSchemaOrArrayAnnotation :
-                                null;
+        final io.swagger.v3.oas.annotations.media.ArraySchema resolvedArrayAnnotation = getArraySchemaAnnotation(resolvedSchemaOrArrayAnnotation);
 
         final BeanDescription beanDesc;
         {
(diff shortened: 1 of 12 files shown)


ewaostrowska:
@Mattias-Sehlstedt thank you for taking your time to work on this issue! 

I do agree that fixing https://github.com/swagger-api/swagger-core/issues/5077 under no circumstances should introduce https://github.com/swagger-api/swagger-core/issues/5115. I have extended your PR to try to cover both of these issues. The commit preserves the nullable by wrapping the property's $ref at resolution time: oneOf for OAS 3.1 and nullable:true + allOf for OAS 3.0. The component schema should remain type:object and not be affected.

Mattias-Sehlstedt:
Due to my experience with tooling tied to OAS I would suggest that the implementation of `wrapNullableRef` be split into a separate issue and then be implemented as a `feat`-commit (same as what I suggested [here](https://github.com/swagger-api/swagger-core/issues/5077#issuecomment-4198211731)). This since many tools might not handle nullable ref:s at all, and that might lead to people raising even more issues).

There is also the issue that for OAS 3.1., there are many "correct" ways of expressing nullable ref:s. So to better capture this, one could preferably have a more semi-formal feat that explained what has been introduced and why (and the basis for the choice of how nullable ref:s are expressed in OAS 3.1).

So the new issue would capture why and how, **but** it could also ensure that `wrapNullableRef` is made as a protected method, so that if anyone believes "it should be a type array and not oneOf with a null child", then we could instantly direct them to the option "Then you can override `wrapNullableRef` with your preferred behavior" (or disable it altogether easily as well). Similar to what was done [here](https://github.com/swagger-api/swagger-core/pull/5042) to easily give the user the possibility to inject small behavioral changes if they want to.

So we would keep this PR as a simple "prevent accidental leakage of `@Nullable` for objects introduce in https://github.com/swagger-api/swagger-core/pull/5018", which would just restore expected and previous behavior. And the real behavior-change is then tracked and explained in isolation and clarity.

daniel-kmiecik (approved):

ewaostrowska:
@Mattias-Sehlstedt I think your concern is valid and this is something we should consider. I have reverted the changes and we will think what is the best way to approach the nullability issues to resolve them completely. But you're right that a dedicated PR will be the best approach here.
Thanks for sharing your thoughts on this one!

PR #5140, opened 2026-04-15
fix: Avoid locale-dependent NumberFormat in IntegerSchema

### LINKED ISSUE #4223: bug: IntegerSchema returns null for negative values for some locales
NumberFormat.getInstance().parse(Integer.toString(-1)) always throws exception in the following locales: ar· ·ckb· ·he· ·ks· ·pa_Arab· ·ur· ·fa· ·ps· ·uz_Arab· ·eo· ·et· ·fi· ·fo· ·gsw· ·ksh· ·lt· ·nb· ·nn· ·rm· ·se· ·sv· (see https://bugs.openjdk.java.net/browse/JDK-8189097)

This means that schemas with negative enum values will replace these with `null` if swagger parser is used on a computer with one of the special locales, making development hard for e.g. Norwegian or Finnish developers.

I've created a small pull request which fixes the problem at #4178

Mattias-Sehlstedt:
# Pull Request

---

## Description

This is a revival of https://github.com/swagger-api/swagger-core/pull/4178.

NumberFormat.getInstance().parse(Integer.toString(-1)) always throws exception in the following locales: ar· ·ckb· ·he· ·ks· ·pa_Arab· ·ur· ·fa· ·ps· ·uz_Arab· ·eo· ·et· ·fi· ·fo· ·gsw· ·ksh· ·lt· ·nb· ·nn· ·rm· ·se· ·sv· (see https://bugs.openjdk.java.net/browse/JDK-8189097)

To see this in action, try running the following code:

```java
import java.text.NumberFormat;
import java.text.ParseException;
import java.util.Locale;

public class NumberFormatIsBroken {
    public static void main(String[] args) throws ParseException {
        Locale.setDefault(new Locale("no"));
        NumberFormat.getInstance().parse(Long.toString(-1));
    }
}
```
It should be valid to ignore locale for `IntegerSchema`, since we are not conducting any parsing based upon the locale (since we are doing `Integer.parseInt()` and `Long.parseLong()`, without considering the `Number casted` that we instantiate).

The locale is mainly used for different decimal representations and thousand separators, but these already fail silently in the current implementation. This since the `Number casted` is properly read and populated. E.g., `2 000` to `2000`, but we then do Integer.parseInt() with `2 000`, which will fail.

Fixes: #4223

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/IntegerSchema.java
+++ b/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/IntegerSchema.java
@@ -1,6 +1,5 @@
 package io.swagger.v3.oas.models.media;
 
-import java.text.NumberFormat;
 import java.util.Objects;
 
 /**
@@ -34,13 +33,14 @@ public class IntegerSchema extends Schema<Number> {
     protected Number cast(Object value) {
         if (value != null) {
             try {
-                Number casted = NumberFormat.getInstance().parse(value.toString());
-                if (Integer.MIN_VALUE <= casted.longValue() && casted.longValue() <= Integer.MAX_VALUE) {
-                    return Integer.parseInt(value.toString());
+                String stringValue = value.toString();
+                long casted = Long.parseLong(stringValue);
+                if (withinIntegerBounds(casted)) {
+                    return Integer.parseInt(stringValue);
                 } else {
-                    return Long.parseLong(value.toString());
+                    return casted;
                 }
-            } catch (Exception e) {
+            } catch (Exception ignored) {
             }
         }
         return null;
@@ -74,5 +74,9 @@ public class IntegerSchema extends Schema<Number> {
         sb.append("    ").append(toIndentedString(super.toString())).append("\n");
         sb.append("}");
         return sb.toString();
+    }
+
+    private boolean withinIntegerBounds(long value) {
+        return Integer.MIN_VALUE <= value && value <= Integer.MAX_VALUE;
     }
 }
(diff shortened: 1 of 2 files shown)


modules/swagger-models/src/test/java/io/swagger/v3/oas/models/media/IntegerSchemaTest.java
```diff
@@ -0,0 +1,86 @@
+package io.swagger.v3.oas.models.media;
+
+import org.testng.annotations.Test;
+
+import java.util.Locale;
+
+import static org.testng.Assert.*;
+
+public class IntegerSchemaTest {
```
Mattias-Sehlstedt, line 9: If it is of interest I could add the tests first in a PR to showcase the current behavior, and then do a PR afterwards that changes the implementation (but also showcases that the behavior is the same).

ewaostrowska: I think the current test coverage is sufficient

ewaostrowska:
Thank you for reviewing this comment @Mattias-Sehlstedt :)

ewaostrowska (approved):

modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/IntegerSchema.java
```diff
@@ -34,13 +33,14 @@ public IntegerSchema _default(Number _default) {
     protected Number cast(Object value) {
         if (value != null) {
             try {
-                Number casted = NumberFormat.getInstance().parse(value.toString());
-                if (Integer.MIN_VALUE <= casted.longValue() && casted.longValue() <= Integer.MAX_VALUE) {
-                    return Integer.parseInt(value.toString());
+                String stringValue = value.toString();
+                long casted = Long.parseLong(stringValue);
+                if (withinIntegerBounds(casted)) {
+                    return Integer.parseInt(stringValue);
```
andrei-ivanov, line 39: why parse the `stringValue` again? wouldn't a cast to `int` of `casted` have worked?

PR #5161, opened 2026-04-23
feat: add Schema.NullableMode for explicit per-property nullable override (#5160)

### LINKED ISSUE #5160: Proposal: introduce Schema.NullableMode enum to mirror RequiredMode / AccessMode pattern
## Background

swagger-core has twice introduced enum-based "Mode" replacements for boolean fields on `@Schema` whose `default false` made it impossible to distinguish "user explicitly set false" from "not specified":

- **2018** — `AccessMode { AUTO, READ_ONLY, WRITE_ONLY, READ_WRITE }` introduced (#2675, #2677), deprecating `readOnly` and `writeOnly`.
- **2022** — `RequiredMode { AUTO, REQUIRED, NOT_REQUIRED }` introduced (#4221, #4286, commit b1729fcf1), deprecating `required`. Commit message rationale: *"so we can have a property annotated @NotNull but still have required = false in the openapi spec."*

The `nullable` field has the same shape — `boolean nullable() default false` — and is now hitting the same problem class as `required` did pre-2022.

## Why now

Auto-detection of nullable has expanded substantially in the last year:

- #5018 (Nov 2025) added native auto-detection of `@Nullable` annotations (Jakarta / Spring / JetBrains / JSR-305) with OAS 3.1 type-array translation.
- Downstream, springdoc-openapi added auto-detection of Kotlin `T?` types in springdoc/springdoc-openapi#3256 (Apr 2026), which has now been reverted in springdoc/springdoc-openapi#3276 due to the same missing-override problem this proposal addresses.
- Recent bug iterations on the auto-detection edges: #5077, #5158, upcoming #5124.

Both auto-detection paths feed the same `Schema.nullable` boolean and have no way to express "user explicitly opts this property out of being marked nullable, regardless of heuristics." Users who try `@Schema(nullable = false)` find it has no effect (it's the default value and reflection can't distinguish it from absence). See springdoc/springdoc-openapi#3138 and the discussion under springdoc/springdoc-openapi#3256 for concrete user reports.

## Proposal

Add to `Schema`:

```java
/**
 * Allows specifying the nullable mode (NullableMode.AUTO, NULLABLE, NOT_NULLABLE).
 *
 * NullableMode.AUTO: lets the library decide based on heuristics (e.g. @Nullable
 *   annotations, downstream Kotlin T? detection in springdoc).
 * NullableMode.NULLABLE: forces the item to be considered nullable regardless
 *   of heuristics.
 * NullableMode.NOT_NULLABLE: forces the item to be considered not nullable
 *   regardless of heuristics.
 */
NullableMode nullableMode() default NullableMode.AUTO;

enum NullableMode {
    AUTO,
    NULLABLE,
    NOT_NULLABLE;
}
```

Deprecate `boolean nullable()` in favor of `nullableMode()`, mirroring the deprecation pattern used for `required` and `readOnly`/`writeOnly`.

## Precedence

Following the precedent set in #4533 ("give precedence to requiredMode annotation"), `nullableMode` should take precedence over the legacy `nullable` boolean and over auto-detection heuristics. Suggested precedence:

1. `nullableMode = NULLABLE` or `NOT_NULLABLE` → use that
2. Legacy `nullable = true` (deprecated path) → treat as `NULLABLE`
3. `nullableMode = AUTO` (default) → run heuristics (`@Nullable` annotations, downstream Kotlin reflection, etc.)

## OAS 3.0 vs 3.1 mapping

Same as the existing handling for `nullable: true`:
- **OAS 3.0:** sets `nullable: true` on simple types; for `$ref`, wraps in `allOf` with `nullable: true` sibling.
- **OAS 3.1:** adds `"null"` to the `type` array; for `$ref`, wraps in `oneOf` with `{ "type": "null" }`.

`NOT_NULLABLE` simply suppresses any of the above, regardless of what heuristics would otherwise have applied.

## Impact on downstream auto-detection

The `@Nullable` auto-detection added in #5018 should defer to `nullableMode != AUTO`. Likewise, downstream libraries (springdoc's planned re-introduction of `KotlinNullablePropertyCustomizer`) can defer to the same field. This gives users a single, uniform escape hatch across all auto-detection sources.

## Happy to follow up with a PR

If this direction is acceptable, I'm willing to put up a PR following the pattern of #4221 (RequiredMode introduction) and #4533 (precedence handling). Confirming intent here first since this is an annotation-surface change.

thejeff77:
## Summary

Introduces `Schema.NullableMode { AUTO, NULLABLE, NOT_NULLABLE }` mirroring the established `RequiredMode` (#4221) and `AccessMode` (#2675) pattern, providing the missing explicit per-property override for nullable.

Closes #5160.

## Motivation

The existing `nullable()` boolean defaults to `false`, which is indistinguishable from "not specified" via reflection. Users cannot opt a property out of nullable auto-detection (added in #5018 for `@Nullable` annotations, and used downstream by springdoc-openapi for Kotlin `T?` reflection in springdoc/springdoc-openapi#3256, since reverted in springdoc/springdoc-openapi#3276 pending this proposal).

This is the same problem class that prompted `RequiredMode` in 2022. From commit b1729fcf1: *"so we can have a property annotated @NotNull but still have required = false in the openapi spec"* — identical situation, just `nullable` instead of `required`.

## Precedence

Mirrors PR #4533 ("give precedence to requiredMode annotation"):

1. `nullableMode = NULLABLE` or `NOT_NULLABLE` → use that
2. Legacy `nullable = true` (deprecated path) → treat as `NULLABLE`
3. `nullableMode = AUTO` (default) → run heuristics (`@Nullable` annotations from #5018, downstream Kotlin reflection, etc.)

## Changes

**`modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/media/Schema.java`**
- Add `NullableMode` enum (AUTO, NULLABLE, NOT_NULLABLE).
- Add `nullableMode()` annotation field defaulting to `AUTO`.
- Deprecate `nullable()` boolean pointing to `nullableMode()`.

**`modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java`**
- `resolveNullable` honors `NullableMode` precedence. When the @Schema annotation is not passed directly (typical for property-level resolution), it is extracted from the `annotations` array.
- Adds `isSchemaAnnotationNullable` helper for downstream "is this nullable" checks (used at the `defaultValue="null"` and `example="null"` literal handling sites, where the existing `schema.nullable()` checks now route through the helper to also honor `NullableMode`).

**`modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java`**
- Parity with `RequiredMode` handling in `hasSchemaAnnotation` (NullableMode != AUTO counts as "has annotation"), `equals` (compares `nullableMode`), and `mergeSchemaAnnotations` (master-wins precedence on `nullableMode`).
- Property-level `@Schema` processing (line ~842) honors `NullableMode`. When `NOT_NULLABLE` is set, explicitly clears any prior nullable indication on the schema object (e.g. `null` already added to the types array by upstream `@Nullable` auto-detection from #5018).

**`modules/swagger-core/src/test/java/io/swagger/v3/core/util/AnnotationsUtilsTest.java`**
- One existing anonymous Schema implementation needed the new `nullableMode()` method override (delegating to the wrapped schema, mirroring the existing `requiredMode()` delegation).

## Tests

`modules/swagger-core/src/test/java/io/swagger/v3/core/issues/Issue5160Test.java` — 9 new tests covering all combinations across OAS 3.0 and OAS 3.1:

- `testModeNullableForcesNullable{OAS30,OAS31}` — `nullableMode=NULLABLE` produces nullable output.
- `testModeNotNullableOverridesNullableAnnotation{OAS30,OAS31}` — `nullableMode=NOT_NULLABLE` overrides upstream `@Nullable` auto-detection.
- `testModeAutoWithAnnotationStillDetectsNullable{OAS30,OAS31}` — `AUTO` (default) lets `@Nullable` detection apply unchanged.
- `testModeAutoWithoutAnyNullableSignalIsNotNullableOAS30` — `AUTO` with no signals is not nullable (no regression).
- `testLegacyNullableTrueStillWorks{OAS30,OAS31}` — backward compatibility for `nullable = true`.

Full existing suite passes (660 tests, 0 failures, 0 errors).

## OAS 3.0 vs 3.1 mapping

Same as the existing handling for `nullable: true`:

- **OAS 3.0:** sets `nullable: true` on the schema object.
- **OAS 3.1:** adds `"null"` to the `type` array.
- `NOT_NULLABLE` suppresses any of the above and clears any prior nullable indication.

## Impact on downstream

- `@Nullable` auto-detection added in #5018 continues to work unchanged when `nullableMode = AUTO`.
- Provides the override mechanism springdoc-openapi needs to safely re-introduce its Kotlin `T?` auto-detection (springdoc/springdoc-openapi#3276 reverted that work pending this proposal).
- `nullable()` boolean is now `@Deprecated` but still honored, mirroring the deprecation pattern used for `required`/`readOnly`/`writeOnly`.

## Test plan

- [ ] CI passes
- [x] Local: `./mvnw -pl modules/swagger-core test -Dmaven.javadoc.skip=true` → 660/660 tests pass including 9 new
- [x] Manual verification of OAS 3.0 / 3.1 output for all five test cases via the snapshot dump in `Issue5160Test`

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -2287,12 +2287,12 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
             try {
                 ObjectMapper mapper = ObjectMapperFactory.buildStrictGenericObjectMapper();
                 JsonNode node = mapper.readTree(schema.defaultValue());
-                // Only return null for "null" string when nullable=true
+                // Only return null for "null" string when the schema is effectively nullable
                 if (node.isNull()) {
-                    if (schema.nullable()) {
+                    if (isSchemaAnnotationNullable(schema)) {
                         return null;
                     } else {
-                        // When nullable=false, treat "null" as literal string
+                        // When not nullable, treat "null" as literal string
                         return schema.defaultValue();
                     }
                 }
@@ -2331,7 +2331,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
                     ObjectMapper mapper = ObjectMapperFactory.buildStrictGenericObjectMapper();
                     JsonNode node = mapper.readTree(schema.example());
                     if (node.isNull()) {
-                        if (schema.nullable()) {
+                        if (isSchemaAnnotationNullable(schema)) {
                             return null;
                         } else {
                             return schema.example();
@@ -2450,8 +2450,30 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
     }
 
     protected Boolean resolveNullable(Annotated a, Annotation[] annotations, io.swagger.v3.oas.annotations.media.Schema schema) {
-        if (schema != null && schema.nullable()) {
-            return true;
+        // Resolve the effective @Schema annotation - prefer the one passed directly, otherwise scan the annotations array.
+        // Some callers (notably property-level resolution) pass schema=null and put @Schema in the annotations array instead.
+        io.swagger.v3.oas.annotations.media.Schema effectiveSchema = schema;
+        if (effectiveSchema == null && annotations != null) {
+            for (Annotation annotation : annotations) {
+                if (annotation instanceof io.swagger.v3.oas.annotations.media.Schema) {
+                    effectiveSchema = (io.swagger.v3.oas.annotations.media.Schema) annotation;
+                    break;
+                }
+            }
+        }
+        // NullableMode takes precedence over both the legacy nullable boolean and auto-detection.
+        if (effectiveSchema != null) {
+            io.swagger.v3.oas.annotations.media.Schema.NullableMode mode = effectiveSchema.nullableMode();
+            if (mode == io.swagger.v3.oas.annotations.media.Schema.NullableMode.NULLABLE) {
+                return true;
+            }
+            if (mode == io.swagger.v3.oas.annotations.media.Schema.NullableMode.NOT_NULLABLE) {
+                return false;
+            }
+            // mode == AUTO: honor legacy nullable boolean if set, otherwise fall through to heuristics.
+            if (effectiveSchema.nullable()) {
+                return true;
+            }
         }
         if (annotations != null) {
             for (Annotation annotation : annotations) {
@@ -2461,6 +2483,30 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
             }
         }
         return null;
+    }
+
+    /**
+     * Returns true if the @Schema annotation effectively declares nullable, honoring
+     * the {@code nullableMode} field with precedence over the legacy {@code nullable}
+     * boolean. Returns false if the annotation explicitly declares not-nullable, or
+     * if no nullability intent is expressed.
+     *
+     * <p>Used for derivative decisions (e.g. whether a "null" defaultValue/example
+     * literal should be treated as actual null). Does not consult auto-detected
+     * @Nullable annotations - for full resolution use {@link #resolveNullable}.
+     */
+    private static boolean isSchemaAnnotationNullable(io.swagger.v3.oas.annotations.media.Schema schema) {
+        if (schema == null) {
+            return false;
+        }
+        io.swagger.v3.oas.annotations.media.Schema.NullableMode mode = schema.nullableMode();
+        if (mode == io.swagger.v3.oas.annotations.media.Schema.NullableMode.NULLABLE) {
+            return true;
+        }
+        if (mode == io.swagger.v3.oas.annotations.media.Schema.NullableMode.NOT_NULLABLE) {
+            return false;
+        }
+        return schema.nullable();
     }
 
     protected BigDecimal resolveMultipleOf(Annotated a, Annotation[] annotations, io.swagger.v3.oas.annotations.media.Schema schema) {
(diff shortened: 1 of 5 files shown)


modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
```diff
@@ -2450,8 +2450,34 @@ protected Boolean resolveReadOnly(Annotated a, Annotation[] annotations, io.swag
     }
 
     protected Boolean resolveNullable(Annotated a, Annotation[] annotations, io.swagger.v3.oas.annotations.media.Schema schema) {
-        if (schema != null && schema.nullable()) {
-            return true;
+        // Resolve the effective @Schema annotation - prefer the one passed directly, otherwise scan the annotations array.
```
Mattias-Sehlstedt: As far as I am aware the general idea is that an `Annotation[]` should never have any `@Schema` annotation considered, that is explicitly what the separate `schema` field is for. 

This is especially of importance since sibling handling is managed with the annotation passed in `schema`.

thejeff77: Good catch — agreed. Pushed a28ebc4 removing that scan. The extraction was redundant defensive code: `NullableMode` handling for property-level `@Schema` is already centralized in `AnnotationsUtils`, where the `@Schema` annotation is the canonical source. Trusting the explicit `schema` parameter here preserves the project's separation (and keeps sibling handling correct). All 10 `NullableMode` tests still pass, full suite remains 661/661 green.

thejeff77:
FYI @ewaostrowska — this builds on the `@Nullable` auto-detection you landed in #5018 by adding the per-property override mechanism (mirroring `RequiredMode`) so users can opt out when needed. Would appreciate your eyes when you have a moment.

thejeff77:
Quick status: rebased on latest master (commit 5603efc5e), 686/686 tests passing, CI green. cc @djankows — pinging in case @ewaostrowska is busy. This adds a per-property \`NullableMode\` override mirroring the established \`RequiredMode\` / \`AccessMode\` pattern, complementing the \`@Nullable\` auto-detection that landed in #5018. Ready for review whenever.

PR #5179, opened 2026-05-12
fix: exclude overridable annotation values when parsing composed annotations

### LINKED ISSUE #5178: [Bug]: OverridesAttribute meta annotation takes the default value

Tied to the behavior introduced in https://github.com/swagger-api/swagger-core/pull/4986 and discussed in https://github.com/swagger-api/swagger-core/pull/4986#issuecomment-4409119218. The behavior is also documented in a [test introduced with the PR](https://github.com/swagger-api/swagger-core/pull/4986/changes#diff-ae61006a93cae59199db3c06af6a10124e5040979a89f229b618c62893669cc4R202).

I am currently using my own injected behavior for Hibernates `@Range`, and this new behavior will affect that since it will parse the default values present on the `@Range` annotation.

I would suggest that the application's behavior would be to omit any overridable value, to ensure that no incorrect default value gets forwarded to the specification.

Both swagger-core and springdoc-openapi have great extension possibilities to inject processing of these types of composed/meta annotations. E.g., as per [this example](https://github.com/springdoc/springdoc-openapi/issues/3180#issuecomment-3693299126) in springdoc.

## Affected Version

2.2.50

2.2.50

## Steps to Reproduce

See [test](https://github.com/swagger-api/swagger-core/pull/4986/changes#diff-ae61006a93cae59199db3c06af6a10124e5040979a89f229b618c62893669cc4R202).

## Actual Behavior

The `@Min` and `@Max` are excluded since we do not read the configured values.

The `@Min` and `@Max` are parsed, as per the shared test.

## Checklist

Mattias-Sehlstedt:
# Pull Request

---

## Description

Change the meta-annotation parsing so that annotations that utilize `@OverridesAttribute` do not take the overridden annotation into consideration. This since the library currently does not support parsing the `@OverridesAttribute` value definition.

I know that the Hibernate validator most likely has code to manage this, but I do not believe that is a dependency that this project needs or wants as of now. Both swagger-core and springdoc-openapi both have extension possibilities for injecting interpretation logic of annotations, so that is how these annotations can be managed currently.

Fixes: #5178

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/ValidationAnnotationsUtils.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/ValidationAnnotationsUtils.java
@@ -3,8 +3,10 @@ package io.swagger.v3.core.util;
 import io.swagger.v3.oas.models.SpecVersion;
 import io.swagger.v3.oas.models.media.Schema;
 
+import javax.validation.OverridesAttribute;
 import javax.validation.constraints.*;
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
 import java.math.BigDecimal;
 import java.util.ArrayDeque;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


ewaostrowska:
Hi @Mattias-Sehlstedt!
Thank you for working on this issue :) 
I agree that clearing the generated constraint data here is better than keeping values that may be wrong.

I also added implementation to cover repeatable` @OverridesAttribute` annotations. Adding full `@OverridesAttribute ` should stay out of scope for now.

ewaostrowska (approved):

PR #5183, opened 2026-05-21
feat: allow @ExampleObject values to be loaded from a file (#5180)

### LINKED ISSUE #5180: [Feature]: Allow values in @ExampleObject to be read from a file
## Feature Description
I would like to see a way to define example payloads in the @ExampleObject annotation from a file, rather than declaring constants for it.
- What problem does it solve?
 For APIs with larger payloads, defining payloads with @ExampleObjects becomes unmanageable because the examples must be defined as constants, resulting in boilerplate code in Java classes.
- How would it improve swagger-core?
It will make it easier to maintain large example payloads without creating constants in Java.
- Is it related to OpenAPI spec support, annotation improvements, integration, etc.?
It would be categorized as annotation improvement.

## Use Case
@ExampleObject ( description = "test description", file = "classpath:/main/resources/openapi/test-example.json")

## Suggested Solution (optional)
Add a feature in the @ExampleObject annotation to read a value from a file.

## Checklist

- [x] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) to ensure this is not a duplicate.
- [x] This feature would be useful to more than just my use case.
- [x] I have provided enough detail for the maintainers to understand the scope of the request.

kuntal1461:
## Summary

- Add `externalFile()` field to `@ExampleObject` annotation to load example payloads from classpath or filesystem files
- Modify `AnnotationsUtils.resolveExample()` to read the file content and process it identically to the inline `value` field (JSON-parsed, falling back to raw string)
- Supports `classpath:` and `file:` prefixes; bare paths default to classpath
- When both `externalFile` and `value` are set, `externalFile` takes precedence; gracefully falls back to `value` if the file is not found

## Motivation

Closes #5180. For APIs with large payloads, defining examples inline in `@ExampleObject(value = "...")` is unmanageable — it requires defining enormous string constants in Java classes. This feature lets users store example payloads in separate files:

```java
@ExampleObject(
    name = "Create User",
    externalFile = "classpath:openapi/examples/create-user-request.json",
    summary = "Full user creation example"
)
```

## Test plan

- [ ] Unit tests added to `AnnotationsUtilsTest`: file-found (JSON parsed as `JsonNode`), `externalFile` precedence over `value`, missing-file fallback to `value`, `classpath:` prefix handling
- [ ] Integration test added to `ExamplesTest`: `ExternalFileRequestBodyExample` loads JSON from `examples/external/user-request.json` and verifies OpenAPI YAML output
- [ ] All 917 existing tests pass (`swagger-core`: 691, `swagger-jaxrs2`: 226)
- [ ] Java 8 compatible (uses `ByteArrayOutputStream` + byte-loop instead of `InputStream.readAllBytes()`)


modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
```diff
@@ -432,6 +437,42 @@ public static Optional<Example> getExample(ExampleObject example, boolean ignore
         return Optional.empty();
     }
 
+    private static String loadExternalFile(String path) {
```
Mattias-Sehlstedt, line 440: Note: I am not affiliated with this repo. These are my own opinions.

----

I believe file handling is outside of what `AnnotationUtils` responsibilities, and that it would make sense to extract this to some separate file-handling util class. This so that the `AnnotationUtils` only know "I should fetch this from a file", but it does not have to consider itself in any way of how files are actually fetched and parsed.

Issue #4404, opened 2023-05-14
Add support for more java.time objects


rupert-madden-abbott:
In [this issue](https://github.com/swagger-api/swagger-core/issues/1445), support for java.time.Instant was added.

Please can support for more of java.time be added, specifically java.time.Duration would be helpful.

aronstrandberg:
This is really frustrating for us as well.

MoellJ:
Just opened a PR for java.time.Duration. If this is accepted I'll also work on other java.time* 'primitives'

daniel-kmiecik:
Closing this because #5184 addresses the issue.

PR #5184, opened 2026-05-21
feat: support Java 8 date/time formats per OpenAPI Formats Registry (#5172)

### LINKED ISSUE #5172: [Feature]: Support for Java 8 date formats
## Feature Description

Java 8 introduced the Date/Time API that has largely replaced Joda-Time.
Relatively recently, OpenAPI has created a [Formats Registry](https://spec.openapis.org/registry/format/index.html) that contains format that map nicely to the Java 8 objects.

## Use Case

Date/Time is a [really complex](https://gist.github.com/timvisee/fcda9bbdff88d45cc9061606b4b923ca) system

The current version (2.2.49) only support 2 types:

```java
    static class TestObjectDates {
        public ZonedDateTime zonedDateTime = ZonedDateTime.parse("2007-12-03T10:15:30+01:00[Europe/Paris]"); // ❌ 'date-time', no format exists in the registry for this type yet
        public OffsetDateTime offsetDateTime = OffsetDateTime.parse("2007-12-03T10:15:30+01:00"); // 🐛 complex object, should be 'date-time'
        public Instant instant = Instant.parse("2007-12-03T10:15:30+01:00"); // ✅
        public LocalDateTime localDateTime = LocalDateTime.parse("2007-12-03T10:15:30"); // 🐛 'date-time', should be 'date-time-local'
        public LocalDate localDate = LocalDate.parse("2007-12-03"); // ✅
        public OffsetTime offsetTime = OffsetTime.parse("10:15:30+01:00"); // 🐛 complex object, should be 'time'
        public LocalTime localTime = LocalTime.parse("10:15:30"); // 🐛 complex object, should be 'time-local'
        public Duration duration = Duration.parse("PT10H15M30S"); // 🐛 complex object, should be 'duration'
    }
```

## Suggested Solution 

Built-in support in swagger-core, I opened https://github.com/swagger-api/swagger-core/pull/5113 to support 'date-time-local'. I don't mind fixing every other types if this issue is accepted

## Checklist

- [X] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) to ensure this is not a duplicate.
- [X] This feature would be useful to more than just my use case.
- [X] I have provided enough detail for the maintainers to understand the scope of the request.

kuntal1461:
## Description

Fixes #5172

Java 8 introduced `java.time.*` types that have become the standard for date/time handling, largely replacing `java.util.Date` and Joda-Time. The [OpenAPI Formats Registry](https://spec.openapis.org/registry/format/) now defines standard format strings that map to these types. This PR brings `swagger-core` into alignment with that registry.

### What was wrong

| Java Type | Before this PR | Should be |
|---|---|---|
| `java.time.OffsetTime` | Expanded as complex object (broken) | `"time"` |
| `java.time.Duration` | Expanded as complex object (broken) | `"duration"` |
| `java.time.LocalDateTime` | `"date-time"` (wrong format name) | `"date-time-local"` |
| `java.time.LocalTime` | Complex object, or `"partial-time"` via opt-in | `"time-local"` |

`OffsetTime` and `Duration` had no mapping at all — they fell through to Jackson's object expansion, producing unusable schemas with internal JDK fields. `LocalDateTime` was mapped to `"date-time"` which is the RFC 3339 *offset* date-time format — incorrect for a type that carries no timezone.

### Why not just fix everything as the default?

`LocalDateTime → "date-time"` is the current default and has been since the early versions of this library. Many existing users depend on this mapping in generated specs, client SDKs, and validation logic. Silently changing it would break them on upgrade with no warning.

`OffsetTime` and `Duration` **are** fixed as the new default because their previous output (a complex object schema) was never usable — no user could have a working integration built on that output.

### What this PR does

**Fixed by default (no user action required):**
- `java.time.OffsetTime`  → `string / "time"`
- `java.time.Duration`    → `string / "duration"`

**Opt-in via `PrimitiveType.enableJava8Formats()`:**
- `java.time.LocalDateTime` → `string / "date-time-local"`
- `java.time.LocalTime`     → `string / "time-local"`

```java
// Add once at application startup to get fully correct OpenAPI Formats Registry mappings
PrimitiveType.enableJava8Formats();
```

This follows the same opt-in pattern already established by `enablePartialTime()` in this codebase.

**`enablePartialTime()` is deprecated** in favour of `enableJava8Formats()`, which supersedes it with the correct registry format name (`"time-local"` instead of `"partial-time"`). The old method is kept as-is so existing callers are not broken.

### Migration path

```
v2.2.51 (this PR) → call enableJava8Formats() to opt in
v3.x (next major) → enableJava8Formats() behaviour becomes the default
```

### New schema model classes (swagger-models)

| Class | Format |
|---|---|
| `DateTimeLocalSchema` | `"date-time-local"` |
| `TimeSchema` | `"time"` |
| `TimeLocalSchema` | `"time-local"` |
| `DurationSchema` | `"duration"` |

Each follows the same structure as the existing `DateTimeSchema` and `DateSchema`.

## Type of Change

- [x] ✨ New feature
- [x] 🐛 Bug fix (`OffsetTime` and `Duration` were broken by default)
- [x] 📝 Documentation (`enablePartialTime()` deprecation Javadoc)

## Checklist

- [x] I have added/updated tests as needed (`Java8DateFormatsTest` covers both default and opt-in behaviour)
- [x] The code builds and passes tests locally (689 tests, 0 failures)
- [x] I have linked related issues (#5172)

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
@@ -5,13 +5,17 @@ import io.swagger.v3.oas.models.media.BinarySchema;
 import io.swagger.v3.oas.models.media.BooleanSchema;
 import io.swagger.v3.oas.models.media.ByteArraySchema;
 import io.swagger.v3.oas.models.media.DateSchema;
+import io.swagger.v3.oas.models.media.DateTimeLocalSchema;
 import io.swagger.v3.oas.models.media.DateTimeSchema;
+import io.swagger.v3.oas.models.media.DurationSchema;
 import io.swagger.v3.oas.models.media.FileSchema;
 import io.swagger.v3.oas.models.media.IntegerSchema;
 import io.swagger.v3.oas.models.media.JsonSchema;
 import io.swagger.v3.oas.models.media.NumberSchema;
 import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.oas.models.media.StringSchema;
+import io.swagger.v3.oas.models.media.TimeLocalSchema;
+import io.swagger.v3.oas.models.media.TimeSchema;
 import io.swagger.v3.oas.models.media.UUIDSchema;
 import org.apache.commons.lang3.StringUtils;
 
@@ -221,6 +225,46 @@ public enum PrimitiveType {
             return new JsonSchema().typesItem("string").format("partial-time");
         }
     },
+    DATE_TIME_LOCAL(java.time.LocalDateTime.class, "date-time-local") {
+        @Override
+        public Schema createProperty() {
+            return new DateTimeLocalSchema();
+        }
+        @Override
+        public Schema createProperty31() {
+            return new JsonSchema().typesItem("string").format("date-time-local");
+        }
+    },
+    TIME(java.time.OffsetTime.class, "time") {
+        @Override
+        public Schema createProperty() {
+            return new TimeSchema();
+        }
+        @Override
+        public Schema createProperty31() {
+            return new JsonSchema().typesItem("string").format("time");
+        }
+    },
+    TIME_LOCAL(java.time.LocalTime.class, "time-local") {
+        @Override
+        public Schema createProperty() {
+            return new TimeLocalSchema();
+        }
+        @Override
+        public Schema createProperty31() {
+            return new JsonSchema().typesItem("string").format("time-local");
+        }
+    },
+    DURATION(java.time.Duration.class, "duration") {
+        @Override
+        public Schema createProperty() {
+            return new DurationSchema();
+        }
+        @Override
+        public Schema createProperty31() {
+            return new JsonSchema().typesItem("string").format("duration");
+        }
+    },
     FILE(java.io.File.class, "file") {
         @Override
         public FileSchema createProperty() {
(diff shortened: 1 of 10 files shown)


daniel-kmiecik:
@kuntal1461 `ModelDeserializer.java` wasn't updated for the 4 new formats. I fixed deserialization for these formats and tests have been added.

daniel-kmiecik (approved):

modules/swagger-core/src/main/java/io/swagger/v3/core/util/PrimitiveType.java
```diff
@@ -581,10 +632,28 @@ private DateStub() {
...
     public static void enablePartialTime() {
         customClasses().put("org.joda.time.LocalTime", PrimitiveType.PARTIAL_TIME);
         customClasses().put("java.time.LocalTime", PrimitiveType.PARTIAL_TIME);
     }
+
+    /**
+     * Opts in to the OpenAPI Formats Registry mapping for {@code java.time.LocalDateTime}:
+     * maps it to format {@code "date-time-local"} instead of the default {@code "date-time"}.
+     *
+     * @since 2.2.51
```
ewaostrowska: Please change to the next release version

ewaostrowska (approved):

daniel-kmiecik:
This one also fixes: #4404

daniel-kmiecik:
@kuntal1461 thanks for the PR!  We really appreciate you taking the time to contribute to the project. Great work!

PR #5186, opened 2026-06-01
fix: let explicit @Schema(format) override type-derived format (#5185)

### LINKED ISSUE #5185: [Bug]: @Schema with format = "uri-reference" on java.net.URI field resolves to format: uri when using ModelConverters

When annotating a field of the class java.net.URI with `@Schema(format="uri-reference")` and then resolving the class to which this field belongs by using ModelConverters.getInstance().read() or any of the read methods (Note: not using the openApi31 option) then the output ResolvedSchema or similar will map it to format: uri.

This can cause Open Api linters to throw warnings or errors when combined with an example field of a non-absolute URI (which is allowed by uri-reference, but not uri)

## Affected Version

2.2.47

2.0.3

## Steps to Reproduce

1. Create any dummy class for testing.
2. Create a field of java.net.URI and annotate it with `@Schema(format="uri-reference")`
3. Resolve it to a Schema or similar
4. Read the format of your field property

Example:
```java
//Step 1
private class UriExample{
     //Step 2
     @io.swagger.v3.oas.annotations.media.Schema(format = "uri-reference")
     public URI uriReferenceValue;
}
//Step 3
final ResolvedSchema model = ModelConverters.getInstance().readAllAsResolvedSchema(UriExample.class);
//Step 4
System.out.println(((Schema)model.schema.getProperties().get("uriReferenceValue")).getFormat());
```
The printed output will be "uri", not "uri-reference"

## Expected Behavior

uri-reference is a valid format specified by OpenApi (see [OpenApi Format Registry](https://spec.openapis.org/registry/format/uri-reference.html)) 
The functionality specified by uri-reference and [RFC 3986 Section 4](https://www.rfc-editor.org/info/rfc3986/#section-4) is also supported by [java.net.URI](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#:~:text=URI%20syntax%20and%20components). Therefore, as it is valid to set a URI field to a non-absolute URI, we expect it to map correctly in the OpenApiSpec when declaring the format to explicitly use uri-reference.

Additionally, the documentation for `@Schema` claims that format `Provides an optional override for the format.`.
Therefore it is expected that
 `@Schema(format="uri-reference") URI myUri ` 
maps to
```yaml
myUri:
  type: string
  format: uri-reference
```
when resolving with ModelConverters.

## Actual Behavior

When using the aforementioned setup and reading such a field with ModelConverters.getInstance(), the Resolved Schema gets assigned the format uri instead of the custom override

## Cause

The bug is caused in `io.swagger.v3.core.util.PrimitiveType.` at the method `fromTypeAndFormat(Type type, String format)` (line 446)

```java
public static PrimitiveType fromTypeAndFormat(Type type, String format) {
        final Class<?> raw = TypeFactory.defaultInstance().constructType(type).getRawClass();
        final Collection<PrimitiveType> keys = MULTI_KEY_CLASSES.get(raw);
        if (keys == null || keys.isEmpty() || StringUtils.isBlank(format)) {
            return fromType(type);
        } else {
            return keys
                .stream()
                .filter(t -> t.getCommonName().equalsIgnoreCase(format))
                .findAny()
                .orElse(null);
        }
    }
```
The method gets a raw Class from the type, which happens to be java.net.URI. It then searches for this raw Type in MULTI_KEY_CLASSES, a map containing enum-entries of the same class but different formats. As there is no URI in this Map, keys becomes null and `fromType(type)` is invoked. This in turn ignores the format and resolves the OpenApi Format from java.net.URI directly, which happens to be 
```yaml
myUri:
  type: string
  format: uri
```

## Suggested Fix
Introduce a new PrimitiveType entry `URI_REFERENCE` that handles "uri-reference" explcitly 

```java
URI_REFERENCE(java.net.URI.class, "uri-reference") {
        @Override
        public Schema createProperty() {
            return new StringSchema().format("uri-reference");
        }
        @Override
        public Schema createProperty31() {
            return new JsonSchema().typesItem("string").format("uri-reference");
        }
    },

```
and add both URI and URI_REFERENCE to MULTI_KEY_CLASSES.
```java
addMultiKeys(multiKeyClasses, URI, java.net.URI.class);
addMultiKeys(multiKeyClasses, URI_REFERENCE, java.net.URI.class);
```

**I can provide a PR with the fix with tests** (I have already created a local branch and can push it)

## Checklist

seonwooj0810:
Fixes #5185

## Problem

When a field carries an explicit `@Schema(format = "...")` whose owning type already resolves to a built-in format, the explicit format is silently dropped. The reported case is `java.net.URI` annotated with `@Schema(format = "uri-reference")`: the type resolves to `format: uri` first, and because the schema already has a (type-derived) format, the explicit override is ignored — producing `format: uri` instead of the requested `uri-reference`.

`@Schema#format` is documented as *"Provides an optional override for the format"*, so an explicitly supplied format should win over the one inferred from the property type.

## Change

In `ModelResolver.resolveSchemaMembers(...)` the format was only applied when the schema did not already have one:

```java
if (StringUtils.isNotBlank(format) && StringUtils.isBlank(schema.getFormat())) {
    schema.format(format);
}
```

This is relaxed so that a non-blank explicit format always takes precedence over a type-derived format:

```java
if (StringUtils.isNotBlank(format)) {
    schema.format(format);
}
```

This fixes the `uri-reference` case in #5185 and any other type whose explicit `@Schema(format)` differs from its inferred format, without needing a per-type enum entry. A field with no explicit `@Schema(format)` still falls back to the type-derived format (e.g. plain `URI` → `uri`).

## Tests

Added `Ticket5185Test` covering both OpenAPI 3.0 and 3.1 resolution:
- `@Schema(format = "uri-reference") URI` resolves to `format: uri-reference`
- a plain `URI` (no annotation) still resolves to `format: uri`

## Test evidence

```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- in io.swagger.v3.core.resolving.Ticket5185Test
```

Full `swagger-core` module suite after the change:

```
Tests run: 689, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Verification done: (1) No in-flight PR — `gh pr list --search "5185" / "uri-reference"` returned none; no open PR touches this path. (2) No active claim — issue is unassigned and has no claim comments. (3) Code-focused — change is in `ModelResolver.java` plus a Java test. (4) Bug still present on `master` — the guarded condition is unchanged at `ModelResolver` line 3184. (5) No parent epic closing this. Branch is based directly on current `master` (no rebase needed).

Note: the issue suggested adding a dedicated `URI_REFERENCE` `PrimitiveType` enum entry. This PR instead fixes the root cause — explicit `@Schema(format)` should always override the inferred format — which covers `uri-reference` and any future explicit-override case with a smaller, more general change. Happy to switch to the enum-based approach if maintainers prefer it.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -3176,7 +3176,9 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
             schema.title(title);
         }
         String format = resolveFormat(a, annotations, schemaAnnotation);
-        if (StringUtils.isNotBlank(format) && StringUtils.isBlank(schema.getFormat())) {
+        if (StringUtils.isNotBlank(format)) {
+            // An explicit format on @Schema is the user's intent and must take precedence
+            // over a format derived from the property type (e.g. java.net.URI -> "uri").
             schema.format(format);
         }
         Object defaultValue = resolveDefaultValue(a, annotations, schemaAnnotation);
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/Ticket5185Test.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/Ticket5185Test.java
@@ -1,0 +1,136 @@
+package io.swagger.v3.core.resolving;
+
+import io.swagger.v3.core.converter.AnnotatedType;
+import io.swagger.v3.core.converter.ModelConverterContextImpl;
+import io.swagger.v3.core.jackson.ModelResolver;
+import io.swagger.v3.oas.annotations.media.Schema;
+import org.testng.annotations.Test;
+
+import javax.validation.constraints.Email;
+import java.net.URI;
+
+import static io.swagger.v3.core.resolving.SwaggerTestBase.mapper;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+
+public class Ticket5185Test {
+
+    @Test
+    public void testExplicitFormatOverridesTypeDerivedFormat() {
+        final ModelResolver modelResolver = new ModelResolver(mapper());
+        final ModelConverterContextImpl context = new ModelConverterContextImpl(modelResolver);
+
+        io.swagger.v3.oas.models.media.Schema schema = context.resolve(new AnnotatedType(UriExample.class));
+
+        assertNotNull(schema);
+        io.swagger.v3.oas.models.media.Schema uriReference =
+                (io.swagger.v3.oas.models.media.Schema) schema.getProperties().get("uriReferenceValue");
+        assertNotNull(uriReference);
+        assertEquals(uriReference.getFormat(), "uri-reference");
+
+        // A URI field without an explicit @Schema(format=...) still resolves to the default "uri" format
+        io.swagger.v3.oas.models.media.Schema plainUri =
+                (io.swagger.v3.oas.models.media.Schema) schema.getProperties().get("uriValue");
+        assertNotNull(plainUri);
+        assertEquals(plainUri.getFormat(), "uri");
+    }
+
+    @Test
+    public void testExplicitFormatPrecedenceForStringAndNumericTypes() {
+        final ModelResolver modelResolver = new ModelResolver(mapper());
+        final ModelConverterContextImpl context = new ModelConverterContextImpl(modelResolver);
+
+        io.swagger.v3.oas.models.media.Schema schema = context.resolve(new AnnotatedType(ExplicitFormatExample.class));
+
+        assertNotNull(schema);
+        assertProperty(schema, "uuidValue", "string", "uuid");
+        assertProperty(schema, "emailValue", "string", "email");
+        assertProperty(schema, "emailValidatedValue", "string", "email");
+        assertProperty(schema, "beanValidationEmailValue", "string", "email");
+        assertProperty(schema, "longValue", "integer", "int64");
+        assertProperty(schema, "integerAsInt64Value", "integer", "int64");
+    }
+
+    @Test
+    public void testExplicitFormatOverridesTypeDerivedFormatOpenApi31() {
+        final ModelResolver modelResolver = new ModelResolver(mapper());
+        modelResolver.setOpenapi31(true);
+        final ModelConverterContextImpl context = new ModelConverterContextImpl(modelResolver);
+
+        io.swagger.v3.oas.models.media.Schema schema = context.resolve(new AnnotatedType(UriExample.class));
+
+        assertNotNull(schema);
+        io.swagger.v3.oas.models.media.Schema uriReference =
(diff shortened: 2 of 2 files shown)


VadimKobyakov:
I am not a maintainer but I feel a little unwell with this "global" fix. It does fix my issue, but it might allow for unintended behavior of the resolving. For example, I might annotate a int field with @Schema(format = "email") and the resolver would allow it. On the other hand, the 3.1 version allows that out of the box. A maintainer should look into that I just wanted to voice my concerns

seonwooj0810:
Thanks for raising this, @VadimKobyakov — it's a fair point to think through.

I'd argue the broader behavior is actually correct here. In both OpenAPI and JSON Schema, `format` is an open-ended **annotation** keyword: it's informational and explicitly *not* coupled to type-level validation. The spec allows custom/arbitrary format values, and validators are free to ignore ones they don't recognize. So `@Schema(format = "email")` on an `int` isn't something the resolver is expected to reject — that's a user-authoring choice, the same way it already is in 3.1 (as you noted).

The `@Schema#format` javadoc also describes it as *"an optional override for the format"*, which reads as explicit user intent that should win over the type-derived default.

swagger-core doesn't validate format/type compatibility anywhere else either, so adding a type-vs-format gate just for this path would be inconsistent and arguably out of scope for the resolver. This change only makes 3.0 behave like 3.1 already does.

That said, I fully agree a maintainer should make the final call. If they'd prefer a narrower fix (e.g. a dedicated `URI_REFERENCE` `PrimitiveType` entry as the original issue suggested), I'm happy to switch to that approach.

Mattias-Sehlstedt:
Everything in the annotation is generally meant to offer a user full control over what is shown in the specification. You can for example make an `int` field a `boolean` if you want. It is not up to the library to decide how the actual object serialization is defined, and thus it is best to offer the possibility to decorate the specification as they please.

So the project's _introspection_ tries to draw a general logical relation between object definition and expected serialization, while it also offer the user the possibility for full customization as well if it is needed through the annotations.

seonwooj0810:
Thanks @Mattias-Sehlstedt — that matches my understanding as well: introspection draws a sensible default from the object definition, while the annotation exists precisely to let users override what ends up in the specification. That's exactly the behavior this PR restores for `format` in 3.0 (and how 3.1 already behaves).

ewaostrowska:
Hi! Thank you, @seonwooj0810, for the PR, and thank you @Mattias-Sehlstedt and @VadimKobyakov for participating in the discussion :)

My understanding of the OAS documentation is that an explicitly supplied `format` should not be restricted to the formats predefined for a particular type.

OAS 3.0 defines `format` as an open string-valued property that can contain any value. OAS 3.1 follows JSON Schema, where `format` is treated as a non-validating annotation by default. So, I do not think swagger-core should reject an explicit value such as `@Schema(format = "email")` on an `Integer`. 

One related case:

```java
@Email
@Schema(format = "custom-email")
public String customEmailValidatedValue;
```

Bean Validation is applied after the property schema has been resolved. Because the `@Email` handler currently sets the format to `email` unconditionally, it appears that `custom-email` may still be overwritten. I would leave that out of scope of this PR.

ewaostrowska (approved):

Issue #2635, opened 2018-01-24
@XmlElement(name="myName") not honored for fields starting with "is"


Treppenhouse:
| Q                                 | A
| --------------------------------- | -------
| Bug or feature request?           |  Bug
| Which Swagger-Core version?       |  1.5.13, 2.0.0-rc2
| Which Java version?               | 1.8
| Which JAX-RS framework & version? | Jersey 2.25.1
 

I have a model object that looks like this (unrelated parts are left out):

```
@XmlAccessorType(XmlAccessType.FIELD)
public class MySuperClass {

}

class MyClass extends MySuperClass {

  @XmlElement(name="beerDrink")
  private String isotonicDrink;

  @XmlElement(name="sugarDrink")
  private String softDrink;

}

```
I believe the class hierarchy is not important but I'll include it here anyway. 
My variable is named ```isotonicDrink``` but when e.g. serializing it to JSON or when looking at my model via swagger UI I would like it to be displayed as ```beerDrink```. 

Now, serializing to JSON works perfectly (I am using Jackson).
However, when looking at this model in my swagger.json or in the "Model" tab in the Swagger UI, the variable is still named ```isotonicDrink```. Renaming ```softDrink``` to ```sugarDrink``` is not a problem though! None of my other fields made any problems, only isotonicDrink. And I didn't even use a fancy format (like underscore). That is when I got really confused.

Now, after some debugging I found the issue #415 where the following hack (d183d8cd0f735fc18f5a50dcd52e2076da69b5df) was introduced in the class ```ModelResolver#resolve(JavaType, ModelConverterContext, Iterator)```:
```
            // hack to avoid clobbering properties with get/is names
            // it's ugly but gets around https://github.com/swagger-api/swagger-core/issues/415
            if (propDef.getPrimaryMember() != null) {
                java.lang.reflect.Member member = propDef.getPrimaryMember().getMember();
                if (member != null) {
                    String altName = member.getName();
                    if (altName != null) {
                        final int length = altName.length();
                        for (String prefix : Arrays.asList("get", "is")) {
                            final int offset = prefix.length();
                            if (altName.startsWith(prefix) && length > offset
                                    && !Character.isUpperCase(altName.charAt(offset))) {
                                propName = altName;
                                break;
                            }
                        }
                    }
                }
            }
```

[...]

Issue #415, opened 2014-01-08
Play2/scala: A property named "is_persistent" gets shortened in the json to "_persistent"


kieranbenton:
As the subject line says - other properties that contain _'s work correctly?

fehguy:
Looks like this is an issue with both the internal (1.3.x) introspection and jackson as well (1.5.x).  Since the focus is on 1.5.x right now, I'm looking into a solution.

The Jackson `BeanDescription` method `findProperties()` is taking public property names like such:

`is_persistent`
`gettersAndHaters`

and mangling them to `_persistent` and `tersAndHaters`

I could use some help from @cowtowncoder on this one:

https://github.com/swagger-api/swagger-core/blob/develop_2.0/modules/swagger-core/src/main/java/com/wordnik/swagger/jackson/ModelResolver.java#L172

```java
    for (BeanPropertyDefinition propDef : beanDesc.findProperties()) {
      Property property = null;
      String propName = propDef.getName();
```

Where the properties in this test class:

https://github.com/swagger-api/swagger-core/blob/develop_2.0/modules/swagger-core/src/test/scala/models/ModelPropertyName.java

```java
public class ModelPropertyName {
  public boolean is_persistent() { return true; }
  public String isometric() { return null; }
  public String gettersAndHaters() { return null; }
}
```

emerge mangled

cowtowncoder:
Right: `isXxx` methods are auto-detected if enabled by visibility (separately controlled as "is-getters", distinct from "regular" getters) AND return type is `boolean` or `Boolean`.
If detected, name mangling is identical to `get` methods.

Jackson 2.5 has a new `MapperFeature`, `USE_STD_BEAN_NAMING`: http://fasterxml.github.io/jackson-databind/javadoc/2.5/com/fasterxml/jackson/databind/MapperFeature.html#USE_STD_BEAN_NAMING
which can be enabled to change it slightly (earlier `getURL` results in `Url`; with STD_BEAN_NAMING_ENABLED, remains `URL`), but I don't think it'd have effect here.

If the goal is NOT to detect these methods, it is easy enough to change by changing visibility levels.
One way is with `@JsonAutoDetect`; another with `ObjectMapper.setVisibility()`, which can either:

* Change level for specific kind of accessor (setter, getter, field, is-getter, constructor), or
* Use a fully custom `VisibilityChecker`

[...]

PR #5192, opened 2026-06-06
Honor PropertyNamingStrategy for get/is-prefixed property names

seonwooj0810:
## Description

When a custom `PropertyNamingStrategy` (e.g. `SNAKE_CASE`) is configured on the mapper, `ModelResolver` produced **inconsistent** schema property names: most fields were translated, but any property whose member name starts with a `get`/`is` prefix followed by a lower-case letter kept its original camelCase name.

Example (`SNAKE_CASE`, Java record):

```java
public record PidLookupResponse(
    String familyName, String expiryDate,
    String issuanceDate, String issuingCountry, String issuingAuthority) {}
```

Produced:

```
family_name, expiry_date, issuanceDate, issuingCountry, issuingAuthority
```

instead of the expected `issuance_date`, `issuing_country`, `issuing_authority`.

### Root cause

The "avoid clobbering properties with get/is names" hack (added for #415) overwrites the Jackson-resolved property name with the **raw member name** whenever the member name starts with `get`/`is` + lower-case. For record components the accessor's physical name is the bare component name (`issuanceDate`), which starts with `is`, so the hack replaced the strategy-translated `issuance_date` with `issuanceDate`.

### Fix

Only apply the hack when **no** `PropertyNamingStrategy` is configured. When a strategy is active the resolved name is authoritative and must be honored uniformly. The hack was only ever needed to restore raw names in the absence of a strategy:
- **#415** (Scala accessor `is_persistent` wrongly shortened to `_persistent`) — default mapper, no strategy → hack still runs.
- **#2635** (JAXB-renamed `is`-prefixed fields keeping their original name) — JAXB mapper, no strategy → hack still runs.

Both existing test cases continue to pass.

This bug surfaces through springdoc and was traced to swagger-core: springdoc/springdoc-openapi#3293.

## Type of Change

- [x] 🐛 Bug fix
- [x] 🧪 Tests

## Checklist

- [x] I have linked related issues

## Test evidence

Added `RecordPropertyNamingStrategyTest` (in `swagger-java17-support`, which compiles at Java 17 for records). It fails on `master` (`issuanceDate` not translated) and passes with this change.

Verification done: reproduced on `master`; ran the full test suites for `swagger-core`, `swagger-jaxrs2` (incl. `ReaderTest#testModelResolverXMLPropertiesName` for #2635) and `swagger-java17-support` — all green (`BUILD SUCCESS`).

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -669,7 +669,15 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
 
             // hack to avoid clobbering properties with get/is names
             // it's ugly but gets around https://github.com/swagger-api/swagger-core/issues/415
-            if (propDef.getPrimaryMember() != null) {
+            //
+            // This restores the raw member name for members that merely start with a get/is prefix
+            // (e.g. a Scala accessor "is_persistent", or a JAXB-renamed field, see #415 and #2635).
+            // It must NOT run when a custom PropertyNamingStrategy (e.g. SNAKE_CASE) is configured:
+            // in that case the strategy legitimately translates names such as "issuanceDate" ->
+            // "issuance_date" and the raw member name must not clobber the translated one
+            // (see springdoc/springdoc-openapi#3293).
+            if (propDef.getPrimaryMember() != null
+                    && _mapper.getSerializationConfig().getPropertyNamingStrategy() == null) {
                 final JsonProperty jsonPropertyAnn = propDef.getPrimaryMember().getAnnotation(JsonProperty.class);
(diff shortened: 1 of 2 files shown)


ewaostrowska:
Thank you @seonwooj0810 for providing this fix!

I added one extra guard. It now also skips the if/ get workaround when Jackson exposes class-level `@JsonNaming` on the resolved bean.

Example:

```java
@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
record Example(String issuanceDate, String getawayDate) {}
```

Here the mapper itself can still have no global naming strategy:

```java
mapper.getSerializationConfig().getPropertyNamingStrategy() == null
```

but Jackson has already resolved the schema names as `issuance_date` and `getaway_date`. Without the extra `@JsonNaming` guard, the old workaround can still see raw member names like `issuanceDate` / `getawayDate` and overwrite the resolved names back to camelCase.

I have also expanded the tests.

daniel-kmiecik (approved):

Issue #5126, opened 2026-04-13
[Bug]: Null key for a Map not allowed in JSON


XhstormR:
## Description of the problem/issue

When swagger-core resolves a bean property annotated with `@JsonUnwrapped`, the generated OpenAPI schema's `properties` map may contain a `null` key entry, which causes Jackson serialization to fail with `JsonMappingException: Null key for a Map not allowed in JSON`.

This is triggered by Spring HATEOAS `EntityModel<T>`, where `getContent()` is annotated with `@JsonUnwrapped`. When swagger-core resolves `EntityModel<SomeDto>`, it unwraps the DTO's properties into the parent schema. However, the unwrapped property `Schema` objects may have `name == null` due to `Schema.getName()` being `@JsonIgnore` and being lost during JSON-based clone operations within `ModelResolver`. These null-named schemas are then inserted into the properties map via `modelProps.put(prop.getName(), prop)`, producing a `null` key.

### Root Cause Analysis

The bug is in `ModelResolver.java` and involves three interacting mechanisms:

**1. `handleUnwrapped()` (line ~1461) extracts Schema values from the inner model's properties map:**

```java
private void handleUnwrapped(List<Schema> props, Schema innerModel,
                              String prefix, String suffix, List<String> requiredProps) {
    if (StringUtils.isBlank(suffix) && StringUtils.isBlank(prefix)) {
        if (innerModel.getProperties() != null) {
            props.addAll(innerModel.getProperties().values());  // Schema values added directly
            // ...
        }
    }
    // ...
}
```

**2. `Schema.getName()` is `@JsonIgnore`, lost during any JSON round-trip clone:**

```java
// io.swagger.v3.oas.models.media.Schema
@JsonIgnore
public String getName() {
    return this.name;
}
```

The `AnnotationsUtils.clone()` method only restores the name for the **top-level** schema, not for nested property schemas:

```java
public static Schema clone(Schema schema, boolean openapi31) {
    String cloneName = schema.getName();
    schema = Json.mapper().readValue(Json.pretty(schema), Schema.class);  // name lost for nested schemas
    schema.setName(cloneName);  // only top-level name is restored
    return schema;
}
```

**3. Null key insertion (line ~940):**

```java
for (Schema prop : props) {
    modelProps.put(prop.getName(), prop);   // null key if prop.getName() is null
}
```

[...]

PR #5193, opened 2026-06-07
Restore inner property name from map key in handleUnwrapped

seonwooj0810:
## What

Makes `ModelResolver.handleUnwrapped` defensive against inner property schemas whose `name` has been lost (e.g. via a JSON-based clone like `AnnotationsUtils.clone`, since `Schema.getName()` is `@JsonIgnore`). It restores each property name from the authoritative inner-properties **map key** before forwarding it to the outer model, so the later `modelProps.put(prop.getName(), prop)` can never insert a `null` key — which otherwise breaks serialization with `JsonMappingException: Null key for a Map not allowed in JSON`. The prefix/suffix branch falls back to the map key the same way.

## Scope / honesty note

Relates to #5126 (reported with Spring HATEOAS `EntityModel<T>`). As @Mattias-Sehlstedt rightly questioned, I could **not** reproduce the reported failure end-to-end from a plain annotated model — I tried an `EntityModel<T>`-like generic wrapper with an `@JsonUnwrapped` content getter and it resolves cleanly on `master` (no null key). So this PR is intentionally **narrowed** to a defensive invariant on `handleUnwrapped`'s output rather than a claim to have reproduced and fixed one specific resolver path.

## Test

`Ticket5126Test` drives `handleUnwrapped` directly with inner property schemas in the post-clone (null-name) state and asserts the resulting property names come from the map key (with and without prefix/suffix). It **fails on `master`** (null name / `p_null_s`) and passes with the fix.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -95,7 +95,6 @@ import java.lang.reflect.Type;
 import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -1461,7 +1460,18 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
     private void handleUnwrapped(List<Schema> props, Schema innerModel, String prefix, String suffix, List<String> requiredProps) {
         if (StringUtils.isBlank(suffix) && StringUtils.isBlank(prefix)) {
             if (innerModel.getProperties() != null) {
-                props.addAll(innerModel.getProperties().values());
+                // Schema.getName() is @JsonIgnore, so any prior JSON-based clone of innerModel
+                // (e.g. AnnotationsUtils.clone) leaves nested property schemas with a null name
+                // while the properties-map key still carries the correct name. Restore from the
+                // map key so the eventual `modelProps.put(prop.getName(), prop)` does not insert
+                // a null key (see swagger-api/swagger-core#5126).
+                for (Map.Entry<String, Schema> entry : ((Map<String, Schema>) innerModel.getProperties()).entrySet()) {
+                    Schema prop = entry.getValue();
+                    if (prop.getName() == null) {
+                        prop.setName(entry.getKey());
+                    }
+                    props.add(prop);
+                }
                 if (innerModel.getRequired() != null) {
                     requiredProps.addAll(innerModel.getRequired());
                 }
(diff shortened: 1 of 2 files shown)


Mattias-Sehlstedt:
Are we sure that the root cause analysis of the reported issue is correct? I am [unable to properly reproduce it myself](https://github.com/Mattias-Sehlstedt/springdoc-openapi/commit/24aecc33cd8f9cf646da157eef06ce596f2be0dd). I do not find that the unit tests in the fix really show anything concrete that can be immediately tied to the issue and it being resolved, since the test relies so heavily on creating schema-instances in weird states.

[I have tried to get feedback from the initial reporters](https://github.com/springdoc/springdoc-openapi/issues/3263#issuecomment-4315498801) but no one has gotten back with a more concrete showcase.

seonwooj0810:
Thanks, that's a fair concern. The current tests prove the resolver branch that loses the map-key-backed property name, but I agree they do not make the link to the reported springdoc scenario obvious enough.

I'll try to strengthen this with a more representative reproduction, ideally one that starts from a small annotated model / unwrapped-map shape instead of constructing schema instances in an unusual intermediate state. If I cannot make that path reproduce the reported behavior, I'll narrow the PR or close it rather than keep a fix whose root cause is not demonstrated clearly.

seonwooj0810:
Following up, @Mattias-Sehlstedt — you were right to push on this.

I tried to build the realistic reproduction I promised: an `EntityModel<T>`-style generic wrapper (`class W<T> extends Base { @JsonUnwrapped @JsonProperty(access = READ_ONLY) T getContent() }`) resolved as `W<SomeDto>`. It resolves cleanly on `master` — no null key — so I could **not** reproduce the reported end-to-end failure from a plain annotated model either. That matches your experience and the lack of a concrete showcase from the reporters.

Rather than keep a fix that asserts an unproven root cause, I've **narrowed** the PR:

- Reframed it as a defensive invariant: `handleUnwrapped` must never forward null-named (→ null-keyed) properties, restoring the name from the authoritative inner-properties map key. For normally-resolved inner models the names are already present, so it's a no-op there — it can't make output worse.
- The unit test now stands explicitly as a guard on that method contract: it fails on `master` (null name / `p_null_s`) and passes with the fix — rather than posing as a springdoc reproduction.
- Updated the PR description and commit message to drop the "Fixes #5126 / surfaces with HATEOAS" claim in favor of "Relates to #5126", with the trigger noted as unconfirmed.

If you'd prefer to only merge fixes with a confirmed end-to-end repro, I'm equally happy to close this. But since the guard is cheap and strictly safe, I'd lean toward keeping it as defense-in-depth. Your call — thanks for the careful review.

Mattias-Sehlstedt:
I am only a contributor to this repo, so I leave it to the maintainers to reason about the last question (but do note that they seem to not be able to be too active as of current, so it might be a while if and when you get feedback from them).

I believe your investigations are now a much better showcase of the lack of reproducibility than what I have done in springdoc, since yours are closer to the source. So if anything that is a very good reason to keep this.

We'll just have to see if the people reporting the issue will ever get back to us with more clear details. My current assumption is that they have injected some Mixin/ModelResolver logic that is affecting something they do not expect it to affect, and thus that the issue is really on their end.

seonwooj0810:
Thanks @Mattias-Sehlstedt, much appreciated. Agreed on all points — I'll leave the final call to the maintainers and keep the PR as-is. Your Mixin/ModelResolver hypothesis matches what I saw; if the reporters come back with a concrete config we can revisit, but the fix here stands on its own as a correctness improvement for the map-key-backed property name path.

seonwooj0810:
Gentle ping in case this slipped off the radar — it's been open about a month. The change is small and fully scoped (restoring the inner property name from the map key in `handleUnwrapped`). Would a maintainer be able to take a look when time allows? Happy to adjust if anything's needed. Thanks!

ewaostrowska:
Thank you @seonwooj0810 for creating this PR and then actively taking care of it. @Mattias-Sehlstedt thank you for your input here! 

I added a few more tests to make sure that the clone case is covered and that existing non-null schema names are not overwritten.
Since the original Spring HATEOAS path still has not been reproduced end to end I have also changed the commit description and comment in the test.

ewaostrowska (approved):

PR #5205, opened 2026-06-16
fix: emit $ref for array items when cycle guard suppresses implementation processing

### LINKED ISSUE #5187: [Bug]:  `ModelResolver` cycle guard drops `items` from recursive `List<T>` schemas (regression since 2.2.41)
### Description

  PR #5004 added a cycle guard (`typesBeingResolved`, `resolveArraySchemaWithCycleGuard`) to prevent `StackOverflowError` on recursive models. When the guard triggers, `AnnotationsUtils.getArraySchema(...)` is invoked with `processSchemaImplementation = false` and the `setItems(...)`
   branch is skipped:

  ```java
  if (arraySchema.schema() != null) {
      if (arraySchema.schema().implementation().equals(Void.class)) {
          // setItems called
      } else if (processSchemaImplementation) {
          // setItems called
      }
      // else: items is never set
  }
  ```

  The resulting schema is missing its `items` (and degrades to `type: object`); in deeper cycles the property is dropped from the schema entirely. The referenced type is already registered in the `ModelConverterContext` at that point, so a `$ref` could safely be emitted.

  ### Swagger-core version

  - Broken: `2.2.41` – `2.2.49` (verified against `master`)
  - Last working: `2.2.40`

  ### How to reproduce

  ```java
  class A { B b; }
  class B { List<A> children; }

  var schemas = ModelConverters.getInstance().readAll(B.class);
  System.out.println(Json.pretty(schemas.get("B")));
  ```

  **Expected (≤ 2.2.40):**

  ```json
  "children": {
    "type": "array",
    "items": { "$ref": "#/components/schemas/A" }
  }
  ```

  **Actual (≥ 2.2.41):**

  ```json
  "children": { "type": "object" }
  ```

  ### Suggested fix

  When the guard prevents `setItems`, fall back to a `$ref` to the item type already registered in the `ModelConverterContext`, instead of leaving `items` unset. This preserves the StackOverflow fix from #5004 while emitting a valid schema.

  ### Related

  - #5004 — introduced the regression

seonwooj0810:
## Description

Fixes #5187.

`AnnotationsUtils.getArraySchema(...)` is invoked with `processSchemaImplementation = false` whenever the cycle guard added in #5004 detects that the array's annotated type is already being resolved further up the stack. In that branch both `setItems(...)` paths are skipped:

```java
if (arraySchema.schema() != null) {
    if (arraySchema.schema().implementation().equals(Void.class)) {
        // setItems called
    } else if (processSchemaImplementation) {
        // setItems called
    }
    // else: items is never set — array degrades to {type: array, no items}
}
```

So when a recursive model uses `@ArraySchema(schema = @Schema(implementation = X.class))` and the guard fires, the resulting array schema loses its `items` field. In OAS 3.1 the existing schema passed in usually already had `items` from `context.resolve(...)`, but the cycle-guard branch itself contributes nothing — and when downstream code starts from an empty array schema (no pre-existing items), the property degrades to bare `{ type: "array" }` or gets pruned entirely. This is the regression #5187 reports against 2.2.41 — 2.2.49.

## Root cause

The cycle guard was introduced in #5004 to prevent `StackOverflowError` on recursive models. To avoid recursing it forces `processSchemaImplementation = false`, but the recovery path then drops the `items` field rather than reaching for the schema that the converter context is already (or about to be) holding for that implementation type.

## Fix

When the cycle guard fires AND `arraySchemaObject.getItems() == null`, look up the implementation class in `ModelConverterContext.getDefinedModels()` and emit `items: { $ref: "#/components/schemas/<Name>" }`. The lookup honours an explicit `@Schema(name = "...")` declaration on the implementation type and falls back to the simple class name (matching the default Jackson convention used elsewhere). If neither is registered we leave `items` unset — so the fallback is purely additive: schemas that populated `items` through other paths are untouched.

The implementation type is exactly the one the user named in `@Schema(implementation = X.class)`, so emitting a `$ref` to its already-registered component preserves the StackOverflow fix while keeping the array schema valid.

## Tests

Added `Issue5187Test` with five focused regressions, exercising:

- OAS 3.1 fallback emits `$ref` to the registered implementation.
- OAS 3.0 fallback emits `$ref` to the registered implementation.
- Explicit `@Schema(name = "Renamed")` is honoured over the simple class name.
- Pre-existing `items` on the array schema are preserved (the fallback only runs when `items` is null).
- If the implementation type isn't registered in the context, `items` is left unset (no invented refs).

Without the fix three of these tests fail (confirmed by reverting `AnnotationsUtils.java` locally); with it the full `swagger-core` suite (692 tests) passes.

```
[INFO] Tests run: 692, Failures: 0, Errors: 0, Skipped: 0
```

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

- [x] I have linked the related issue (#5187)

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
@@ -559,6 +559,17 @@ public abstract class AnnotationsUtils {
             } else if (processSchemaImplementation) {
                 getSchema(arraySchema.schema(), arraySchema, false, arraySchema.schema().implementation(), components, jsonViewAnnotation, openapi31, context)
                         .ifPresent(arraySchemaObject::setItems);
+            } else if (arraySchemaObject.getItems() == null && context != null) {
+                // Cycle guard active: the implementation type is currently being
+                // resolved up the call stack, so we cannot recurse into it.
+                // The type has already been registered as a component, so emit a
+                // $ref instead of leaving the array's items unset (gh-5187).
+                String implName = findRegisteredSchemaName(arraySchema.schema().implementation(), context);
+                if (implName != null) {
+                    Schema itemsRef = openapi31 ? new JsonSchema() : new Schema();
+                    itemsRef.set$ref(COMPONENTS_REF + implName);
+                    arraySchemaObject.setItems(itemsRef);
+                }
             }
         }
 
@@ -2100,6 +2111,37 @@ public abstract class AnnotationsUtils {
             if (annotation instanceof io.swagger.v3.oas.annotations.media.Schema) {
                 return (io.swagger.v3.oas.annotations.media.Schema) annotation;
             }
+        }
+        return null;
+    }
+
+    /**
+     * Looks up the registered schema name for an implementation class in the
+     * given {@link ModelConverterContext}.
+     * <p>
+     * Prefers an explicit {@code @Schema(name = "...")} declaration on the
+     * class, then the type-name resolved by the underlying {@code ObjectMapper}
+     * (when available), and finally the simple class name. Returns {@code null}
+     * if no matching schema is registered.
+     * <p>
+     * Used to recover a usable {@code $ref} when array-schema processing is
+     * suppressed by a cycle guard (gh-5187).
+     */
+    static String findRegisteredSchemaName(Class<?> impl, ModelConverterContext context) {
+        if (impl == null || impl.equals(Void.class) || context == null) {
+            return null;
+        }
+        Map<String, Schema> defined = context.getDefinedModels();
+        if (defined == null || defined.isEmpty()) {
+            return null;
+        }
+        io.swagger.v3.oas.annotations.media.Schema annotated = impl.getAnnotation(io.swagger.v3.oas.annotations.media.Schema.class);
+        if (annotated != null && StringUtils.isNotBlank(annotated.name()) && defined.containsKey(annotated.name())) {
+            return annotated.name();
+        }
+        String simpleName = impl.getSimpleName();
+        if (StringUtils.isNotBlank(simpleName) && defined.containsKey(simpleName)) {
+            return simpleName;
         }
         return null;
     }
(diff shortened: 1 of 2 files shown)


caoimhebyrne:
@ewaostrowska I don't think this should have the 'backlog' label... any array type within a schema that references a non-primitive (e.g. string etc) type is generated incorrectly at the moment.

seonwooj0810:
Thanks for the bump, @caoimhebyrne — agreed this hurts real payloads and shouldn't sit in the backlog.

One scoping note so maintainers can size it accurately: the regression this PR fixes is specifically the **cycle-guard branch** added in #5004. When `ModelResolver` detects that an array's annotated implementation type is already being resolved higher up the stack, it calls `getArraySchema(..., processSchemaImplementation = false)`, and that recovery path never sets `items` — so a `List<T>` on a recursive/self-referential model degrades to `{ type: "array" }` with no `items` (exactly #5187, regressed 2.2.41–2.2.49).

The fix is purely additive: it only fills `items` with a `$ref` to the already-registered component when `items` is otherwise `null`, so arrays that already resolve correctly today are left untouched.

If you're seeing arrays of non-primitive types dropped **outside** the recursive/cycle-guard case, that would be a different code path from what this PR touches — a minimal repro would be really helpful, and I'm happy to extend this fix or open a follow-up to cover it.

For reference, the PR adds five focused regression tests and the full 692-test suite passes locally, so it's ready for review whenever a maintainer has a moment.

ewaostrowska:
Thank you @seonwooj0810 and @caoimhebyrne  for your engagement and for helping clarify the scope and impact of this issue. 

One limitation remains here. The guarded fallback does not resolve component names produced by FQN mode or a custom `TypeNameResolver`. Supporting those configurations reliably would require a broader change, such as moving the recovery logic into `ModelResolver`, where the naming strategy is available.

That would involve additional implementation and testing work, so it can be considered out of scope for this issue and addressed separately as a follow-up.

Thanks again for the contribution :)

ewaostrowska (approved):

PR #5224, opened 2026-07-02
Test operation security serialization

kriptoburak:
## Summary

- Add a Swagger Core model serialization regression for operation security alternatives.
- Assert the serialized OpenAPI keeps the anonymous requirement before the API key requirement.

## Validation

- Static whitespace check passed before commit.
- Not run: this environment does not have a Java runtime.


kriptoburak:
I re-audited this against current `master` and the repository history. I cannot provide the requested issue, user report, or divergent observed output: current `master` already serializes the operation security list as `[{}, {"ApiKeyAuth": []}]`, which is also the expected output, and repository searches found no related regression report. The patch therefore only pins existing behavior without a demonstrated coverage need or production fix. I am withdrawing it rather than inventing a regression rationale. Thank you for the clear review criteria.

PR #5230, opened 2026-07-13
fix: keep polymorphic subtype composition when re-resolved standalone (#5028)

### LINKED ISSUE #5028: [Bug]: Polymorphic types don't include composite schema reference

This [project](https://github.com/user-attachments/files/24174758/PolyTypes.zip) illustrates the problem. When having polymorphic types and using the `swagger-gradle-plugin` to generate the OpenAPI descriptor file, subtypes are missing the combined schema reference (`allOf`) to the parent type.

This is the current output in 2.2.41:
```json
"DateAttributeTypeImpl" : {
  "type" : "object",
  "properties" : { }
}
```

And this was the expected output in 2.2.40:
```json
"DateAttributeTypeImpl" : {
  "type" : "object",
  "allOf" : [ {
    "$ref" : "#/components/schemas/AttributeType"
  } ]
}
```

I debugged and the first pass of the model resolver for the parent type populates the reference correctly, however the model is redefined [here](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java#L1133) and the reference is lost. I don't know enough of the model resolution to isolate a fix.

## Affected Version

`2.2.41`

`2.2.41`

## Steps to Reproduce

Expand the project attached and execute:

```shell
./gradlew clean resolve
```

## Expected Behavior

Subtypes should include a composite schema to the parent type like so

```json
"DateAttributeTypeImpl" : {
  "type" : "object",
  "allOf" : [ {
    "$ref" : "#/components/schemas/AttributeType"
  } ]
}
```

## Actual Behavior

Subtypes are created as independent models with no reference to the parent:

```json
"DateAttributeTypeImpl" : {
  "type" : "object",
  "properties" : { }
}
```

## Logs / Stack Traces

none
## Checklist

seonwooj0810:
Fixes #5028

## Root cause

Polymorphic subtypes lose their `allOf` reference to the parent (regression in 2.2.41): a subtype that resolved to
```json
"DateAttributeTypeImpl": { "type": "object", "allOf": [ { "$ref": "#/components/schemas/AttributeType" } ] }
```
degrades to `{ "type": "object", "properties": {} }`.

The parent's `resolveSubtypes` builds each subtype's composed schema (`allOf` → parent) and registers it via `defineModel`. However, when the same subtype is resolved **standalone** afterwards — here while resolving the parent's `@DiscriminatorMapping`, whose `resolveDiscriminator` calls `context.resolve(new AnnotatedType().type(mapping.schema()))` **without** the `subtype(true)` flag — the resolution-cache lookup misses (`AnnotatedType` equality was tightened in #4975 / #5005), so the subtype is re-resolved as a plain object and `modelByName` is overwritten, dropping the composition.

## Fix

Guard model registration in `ModelConverterContextImpl` so a schema **without** composition keywords (`allOf`/`anyOf`/`oneOf`) never overwrites an already-registered **composed** schema of the same name. The forward direction (plain → composed) is unaffected, so the composed subtype definition survives the later plain re-resolution. Applied at both storage sites (`defineModel` and the `resolve()` name mapping).

## Tests

Added `Ticket5028Test` reproducing the reporter's hierarchy (interface parent with `subTypes` + `discriminatorMapping`, subtype via `@Schema(implementation=...)`). Both assertions fail on master and pass with the fix, for OpenAPI 3.0 and 3.1.

## Verification done
- (1/2) No in-flight PR: searched open/all PRs for the symbol and my own open PRs — none cover this root cause; only cross-ref is a Copilot WIP bot issue.
- (4) Confirmed the overwrite path on master (`resolveDiscriminator` L2746 → `ModelConverterContextImpl.resolve`/`defineName`).
- (3) Regression source #5003 is closed; this is the residual case.
- Full `swagger-core` module suite (722 tests) + focused composition/subtype/discriminator suite (38 tests) pass with the change. `git commit -s` (DCO).

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/ModelConverterContextImpl.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/ModelConverterContextImpl.java
@@ -57,7 +57,7 @@ public class ModelConverterContextImpl implements ModelConverterContext {
         if (LOGGER.isTraceEnabled()) {
             LOGGER.trace(String.format("defineModel %s %s", name, model));
         }
-        modelByName.put(name, model);
+        putModelByName(name, model);
 
         if (StringUtils.isNotBlank(prevName) && !prevName.equals(name)) {
             modelByName.remove(prevName);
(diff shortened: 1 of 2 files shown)


seonwooj0810:
Bumping this fix whenever there's time — happy to add more regression tests around the polymorphic re-resolution case if helpful.

PR #5232, opened 2026-07-14
fix(model-resolver): retain identical recursive properties

### LINKED ISSUE #5226: [Bug]: Swagger does not create all the elements in a component
## Swagger is not creating one of the elements in a data structure

In this component, Swagger does not generate the "slots" element. I've tried numerous @Schema annotations to try and force it to create 'slots', but none have been successful. I presume it has something to do with the recursive nature of the element?

```
package org.openhab.core.ui.components;

import java.util.ArrayList;
import java.util.HashMap;
import java.util.List;
import java.util.Map;

import org.eclipse.jdt.annotation.NonNullByDefault;
import org.eclipse.jdt.annotation.Nullable;

import io.swagger.v3.oas.annotations.media.ArraySchema;
import io.swagger.v3.oas.annotations.media.Schema.AdditionalPropertiesValue;

/**
 * A UIComponent represents a piece of UI element for a client frontend to render; it is kept very simple and delegates
 * the actual rendering and behavior to the frontend.
 *
 * It has"s reference to a component's name as defined by the frontend, a map of configuration parameters, and several
 * named "slots", or placeholders, which may contain other sub-components, thus defining a tree.
 *
 * No checks are performed on the actual validity of configuration parameters and their values, the validity of a
 * particular slot for a certain component or the validity of certain types of sub-components within a particular slot:
 * that is the frontend's responsibility.
 *
 * @author Yannick Schaus - Initial contribution
 */
@NonNullByDefault
public class UIComponent {
    String component;

    @Schema(additionalProperties = AdditionalPropertiesValue.TRUE)
    @Nullable
    Map<String, Object> config;

    @ArraySchema(schema = @Schema(implementation = UIComponent.class, description = "The nested sub-components inside this slot"))
    @Nullable
    Map<String, List<UIComponent>> slots;

    /**
     * Empty constructor for deserialization.
     */
    public UIComponent() {
        this.component = "";
        this.config = new HashMap<>();
    }

    /**
     * Constructs a component by its type name - component names are not arbitrary, they are defined by the target
     * frontend.
     *
     * @param componentType type of the component as known to the frontend
     */
    public UIComponent(String componentType) {
        this.component = componentType;
        this.config = new HashMap<>();
    }

    /**
     * Retrieves the type of the component.
     *
     * @return the component type
     */
    @Schema(hidden = true)
    public String getType() {
        return component;
    }

    /**
     * Retrieves the type of the component.
     *
     * @return the component type
     */
    public String getComponent() {
        return component;
    }

    /**
     * Sets the type of the component.
     *
     * @param component the component type
     */
    public void setComponent(String component) {
        this.component = component;
    }

    /**
     * Gets all the configuration parameters of the component
     *
     * @return the map of configuration parameters
     */
    public Map<String, Object> getConfig() {
        return config;
    }

    /**
     * Sets all the configuration parameters of the component
     *
     * @param config the map of configuration parameters
     */
    public void setConfig(Map<String, Object> config) {
        this.config = config;
    }

    /**
     * Adds a new configuration parameter to the component
     *
     * @param key the parameter key
     * @param value the parameter value
     */
    public void addConfig(String key, Object value) {
        this.config.put(key, value);
    }

    /**
     * Returns all the slots of the component including their sub-components
     *
     * @return the slots and their sub-components
     *         Sets all the slots of the component
     *
     * @param slots the slots and their sub-components
     */
    public void setSlots(Map<String, List<UIComponent>> slots) {
        this.slots = slots;
    }

    /**
     * Adds a new empty slot to the component
     *
     * @param slotName the name of the slot
     * @return the empty list of components in the newly created slot
     */
    @Nullable
    public List<UIComponent> addSlot(String slotName) {
        if (slots == null) {
            slots = new HashMap<>();
        }
        List<UIComponent> newSlot = new ArrayList<>();

        if (slots == null) {
            return null;
        }

        this.slots.put(slotName, newSlot);

        return newSlot;
    }

    /**
     * Gets the list of sub-components in a slot
     *
     * @param slotName the name of the slot
     * @return the list of sub-components in the slot
     */
    @Nullable
    public List<UIComponent> getSlot(String slotName) {
        if (slots == null) {
            return null;
        }
        return this.slots.get(slotName);
    }

    /**
     * Add a new sub-component to the specified slot. Creates the slot if necessary.
     *
     * @param slotName the slot to add the component to
     * @param subComponent the sub-component to add
     */
    public void addComponent(String slotName, UIComponent subComponent) {
        List<UIComponent> slot;
        if (slots == null || !slots.containsKey(slotName)) {
            slot = addSlot(slotName);
        } else {
            slot = getSlot(slotName);
        }

        if (slot == null) {
            return;
        }

        slot.add(subComponent);
    }
}
```

The element generated in the openapi spec is the following:

```
            "UIComponent": {
                "type": "object",
                "properties": {
                    "component": {
                        "type": "string"
                    },
                    "config": {
                        "type": "object",
                        "additionalProperties": true
                    }
                },
                "description": "The nested sub-components inside this slot"
            },
```

## Affected Version

2.2.49

<!-- Can you identify when the issue was introduced?
We just noticed this, so I presume this has always been the case.

## Expected Behavior
I expect that this code would generate the following:

```
"UIComponent": {
  "required": ["component", "config"],
  "type": "object",
  "properties": {
    "component": {
      "type": "string"
    },
    "config": {
      "type": "object",
      "additionalProperties": true
    },
    "slots": {
      "type": "object",
      "additionalProperties": {
        "type": "array",
        "items": {
          "$ref": "#/components/schemas/UIComponent"
        }
      }
    }
  }
}
```

## Actual Behavior

The "slots" element is not being created for this component.

## Additional Context
This is part of the openhab opensource project.

## Checklist

### LINKED ISSUE #5231: [Bug]: Recursive properties disappear when sharing identical annotations with wrapper class

A recursive property silently disappears from the generated OpenAPI schema if it happens to have the exact same name and annotations as a property in its wrapper class.

If I have a `nodes` property with `@JsonProperty` and `@JsonPropertyDescription` inside a wrapper class, and the exact same property signature inside the child class, the child's property gets completely dropped from the generated schema.

It looks like the cycle detection cache (`processedTypes`) gets confused because `AnnotatedType.equals()` doesn't check the parent class. It sees the same property name and annotations across the two classes and incorrectly thinks it hit an infinite recursion loop.

## Affected Version

Tested on `master` (2.2.53-SNAPSHOT), but likely present across 2.2.x.

## Steps to Reproduce

1. Define a `TestNodeDTO` with a recursive `List<TestNodeDTO> nodes` property.
2. Define a `WrapperDTO` that also has a `List<TestNodeDTO> nodes` property.
3. Add the exact same annotations to both `nodes` properties (e.g., `@JsonProperty` and `@JsonPropertyDescription`).
4. Generate the schema for `WrapperDTO`.  

```java
class WrapperDTO {
    @JsonProperty("nodes")
    @JsonPropertyDescription("Child nodes")
    public List<TestNodeDTO> nodes;
}

class TestNodeDTO {
    @JsonProperty("name")
    public String name;

    @JsonProperty("nodes")
    @JsonPropertyDescription("Child nodes")
    public List<TestNodeDTO> nodes;
}
```
## Expected Behavior

The  TestNodeDTO  schema should include the  nodes  property as an array ref. The cycle guard shouldn't confuse properties from two different classes just because they share the same signature.

## Actual Behavior

The  nodes  property is completely missing from the generated  TestNodeDTO  schema.

## Logs / Stack Traces

No errors or logs, the property is just dropped silently.

## Additional Context

I already prepared a fix for this by updating  AnnotatedType.equals()  to include the parent class name for schema properties. I'll link the PR shortly!

## Checklist

[✓] I have searched the existing issues and this is not a duplicate.
[✓] I have provided sufficient information for maintainers to reproduce the issue.

bender316:
# Pull Request

## Description

This PR fixes a bug where recursive properties are silently dropped from the generated OpenAPI schema if they happen to share the exact same name and annotations as a property in their wrapper class.

**Root cause:**
The cycle detection cache (`processedTypes`) in `ModelConverterContextImpl` uses `AnnotatedType` to prevent infinite recursion. Previously, `AnnotatedType.equals()` did not check the `parent` class for schema properties. As a result, when it encountered a child property with the exact same signature as a previously processed wrapper property, it mistakenly treated it as a self-referential cycle and dropped it.

**Fix:**
Updated `AnnotatedType.equals()` and `hashCode()` to conditionally include the `parent` class name when `schemaProperty == true`. This correctly differentiates identically-annotated properties across different classes while preserving the cycle guard for truly self-referential loops.

Fixes: #5231
Fixes: #5226

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

- Includes `RecursivePropertyMissingTest.java` to prevent future regressions.
- Fixed a legacy test expectation (`ArrayOfSubclassTest_expected30.json`) that was previously masking this bug by falsely asserting that the property should be omitted.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/AnnotatedType.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/converter/AnnotatedType.java
@@ -277,9 +277,13 @@ public class AnnotatedType {
         AnnotatedType that = (AnnotatedType) o;
         List<Annotation> thisAnnotations = getProcessedAnnotations(this.ctxAnnotations);
         List<Annotation> thatAnnotations = getProcessedAnnotations(that.ctxAnnotations);
+        String thisParentName = this.parent != null ? this.parent.getName() : null;
+        String thatParentName = that.parent != null ? that.parent.getName() : null;
+
         return  includePropertiesWithoutJSONView == that.includePropertiesWithoutJSONView &&
                 schemaProperty == that.schemaProperty &&
                 isSubtype == that.isSubtype &&
+                (!schemaProperty || Objects.equals(thisParentName, thatParentName)) &&
                 Objects.equals(type, that.type) &&
                 Objects.equals(thisAnnotations, thatAnnotations) &&
                 Objects.equals(jsonViewAnnotation, that.jsonViewAnnotation) &&
@@ -289,7 +293,8 @@ public class AnnotatedType {
     @Override
     public int hashCode() {
         List<Annotation> processedAnnotations = getProcessedAnnotations(this.ctxAnnotations);
-        return Objects.hash(type, jsonViewAnnotation, includePropertiesWithoutJSONView, processedAnnotations, schemaProperty, isSubtype, schemaProperty ? propertyName : null);
+        String parentName = (schemaProperty && this.parent != null) ? this.parent.getName() : null;
+        return Objects.hash(type, jsonViewAnnotation, includePropertiesWithoutJSONView, processedAnnotations, schemaProperty, isSubtype, schemaProperty ? propertyName : null, parentName);
     }
 
     private boolean processableAnnotationPackage(Package pkg) {
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/RecursivePropertyMissingTest.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/RecursivePropertyMissingTest.java
@@ -1,0 +1,43 @@
+package io.swagger.v3.core.resolving;
+
+import io.swagger.v3.core.converter.ModelConverters;
+import io.swagger.v3.oas.models.media.Schema;
+import org.testng.annotations.Test;
+
+import java.util.List;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertEquals;
+
+public class RecursivePropertyMissingTest {
+
+    static class WrapperDTO {
+        @JsonProperty("nodes")
+        @JsonPropertyDescription("Child nodes")
+        public List<TestNodeDTO> nodes;
+    }
+
+    static class TestNodeDTO {
+        @JsonProperty("name")
+        public String name;
+
+        @JsonProperty("nodes")
+        @JsonPropertyDescription("Child nodes")
+        public List<TestNodeDTO> nodes;
+    }
+
+    @Test
+    public void testRecursivePropertyNotMissing() {
+        Map<String, Schema> schemas = ModelConverters.getInstance().readAll(WrapperDTO.class);
+        Schema testNodeDTOSchema = schemas.get("TestNodeDTO");
+        assertNotNull(testNodeDTOSchema);
+        
+        Map<String, Schema> properties = testNodeDTOSchema.getProperties();
+        assertNotNull(properties, "Properties should not be null");
+        assertNotNull(properties.get("nodes"), "The 'nodes' property is missing from TestNodeDTO schema");
+    }
+}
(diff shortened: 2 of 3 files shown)


jsjames:
Any update on this PR?   Thanks.

daniel-kmiecik (approved):

daniel-kmiecik:
@bender316 thanks for the contribution and for taking the time to work on this! Really appreciate it.

Issue #5061, opened 2026-02-23
[Bug]: Number example treated as a string when generating swagger file.


SandorArpa:
## Description of the problem/issue

The following mark up 
`@SchemaProperty(
               name = "price",
               schema = @Schema(example = "9090", maximum = "10000000", type = "number", minimum = "1",
               description = "Price override for the payment amount. If not present this will defaults to price of product.")) (https://github.com/alphagov/pay-products/blob/529c1873e6ff2e7842a63654aeb50d7d19cb1fda/src/main/java/uk/gov/pay/products/resources/PaymentResource.java#L107)

interpreted as a string in version 2.2.43:

`type: number
  description: Price override for the payment amount. If not present this will defaults to price of product.
  example: "9090"
  maximum: 10000000
  minimum: 1`.

In version 2.2.42 it was interpreted as a number:
`type: number
  description: Price override for the payment amount. If not present this will defaults to price of product.
  example: 9090
  maximum: 10000000
  minimum: 1`.

## Affected Version

2.2.42 to 2.2.43

## Steps to Reproduce
Setup a SchemaProperty in a RequestBody where type = "number' and example = "1000"

## Expected Behavior

Example is a number, not a string

## Actual Behavior

Example is a string

## Logs / Stack Traces

N/A

## Additional Context
Change did not affect Java classes, eg:

`@Schema(example = "1000.00") 
 private final BigDecimal averageAmount;`

is still interpreted as a number.

## Checklist

- [x] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) and this is not a duplicate.
- [ x] I have provided sufficient information for maintainers to reproduce the issue.

Darthpi:
I believe the additional context is not correct, given this example.
(This may be a different cause/area but I would title it exactly the same as this bug so I'll include it here)

comparison between 2.2.43 and 2.2.44 (tested with BigDecimal and Double)
Given this example endpoint and dto:

```
@PostMapping(path = "/example", consumes = MediaType.APPLICATION_JSON_VALUE)
public ResponseEntity<ExampleDto> crud() {
  return null;
}
```
  
```
private record ExampleDto(
  @Schema(example = "10.00", description = "Bug test decimal value") BigDecimal exampleDec) {
})
```

api-docs

```
2.2.42
"ExampleDto": {
	"type": "object",
	"properties": {
		"exampleDec": {
			"type": "number",
			"description": "Bug test decimal value",
			"example": 10.0   <------
[...]

PR #5234, opened 2026-07-15
fix: apply explicit @Schema(type) to the types set under OpenAPI 3.1

### LINKED ISSUE #5233: [Bug]: Explicit @Schema(type = "number"|"integer"|"boolean") is emitted as "string" under OpenAPI 3.1

When generating an **OpenAPI 3.1** schema, an explicit `@Schema(type = "...")` on a property is resolved incorrectly. The scalar `type` is set correctly on the `Schema` object, but the `types` set is populated with `["string"]`. Since the 3.1 serializer reads the `types` set, the property is emitted as `type: "string"` regardless of the declared type.

- **Goal:** produce a correct OpenAPI 3.1 document for DTOs that use explicit `@Schema(type = ...)`.
- **Annotation misbehaving:** `@Schema` — an explicit `type = "number" | "integer" | "boolean"` is emitted as `"string"`.
- **Correct case:** properties relying on type *inference* (no explicit `type` in `@Schema`) serialize correctly. Only an explicitly declared scalar `type` is affected.
- **Framework:** reproduces on plain swagger-core (`ModelConverters` + `Json31`); also observed via springdoc-openapi with `springdoc.api-docs.version=openapi_3_1`.
- **Not examples-related:** it reproduces on fields with no `example`, so it is distinct from #5061 / #5062 (numeric example serialization).
- **Downstream:** the resulting spec misreports field types, which breaks OpenAPI-based type/client generation (numbers/booleans generated as `string`).

Under OpenAPI 3.0 (`openapi31 = false`) the same DTO serializes correctly, so this is specific to 3.1 generation.

## Affected Version

2.2.47

Not fully bisected. Confirmed present in **2.2.47** and still present in **2.2.52** (latest release at time of filing).

## Steps to Reproduce

1. Add a DTO with explicit `@Schema(type = ...)` properties and resolve it in OpenAPI 3.1 mode:

   ```java
   import java.math.BigDecimal;

   public class Repro {

     static class Dto {
     }

       ResolvedSchema resolved =

     }
   }
   ```

2. Resolve with `new ModelConverters(true)` (OpenAPI 3.1 mode) and serialize with `Json31`.
3. Observe the serialized output and inspect the resolved `Schema` objects.

## Expected Behavior

The declared type is preserved:

```json
{
  "properties": {
  }
}
```

## Actual Behavior

Every property with an explicit scalar type collapses to `"string"`:

```jsonc
{
  "properties": {
  }
}
```

`inferred` (inferred type) and `unit` (enum) serialize correctly — the defect is limited to an explicitly declared scalar type.

Inspecting the resolved (pre-serialization) `Schema` objects shows the root cause — the scalar type is right, the `types` set is wrong:

```text
amount -> getType()="number"  getTypes()=["string"]
count  -> getType()="integer" getTypes()=["string"]
flag   -> getType()="boolean" getTypes()=["string"]
```

## Logs / Stack Traces

No exception or log output — the schema is generated silently with the wrong type.

## Additional Context

- The same DTO serialized under OpenAPI 3.0 (`new ModelConverters(false)`) produces correct types, confirming this is a 3.1-generation regression.
- Likely related to #4679 (`@Schema` type/implementation ignored under OpenAPI 3.1).
- Distinct from #5061 / #5062, which concern numeric `example` serialization; this issue is about the property `type` and reproduces with no `example` present.
- Current workaround: a post-processing pass that re-syncs each schema's `types` set to its scalar `type` produces a correct 3.1 document.

## Checklist

[Repro.java](https://github.com/user-attachments/files/30044260/Repro.java)

forhadmethun:
## Description

Under OpenAPI 3.1, an explicit `@Schema(type = ...)` was applied via the legacy scalar `setType()`, while the 3.1 serializer reads the `types` set. The set kept its default (`"string"`), so `number` / `integer` / `boolean` fields were rendered as `string`. This change populates the `types` set when `openapi31` is enabled, mirroring the handling of the plural `types()` attribute directly below.

`AnnotationsUtils.getSchemaFromAnnotation`:

```java
if (StringUtils.isNotBlank(schema.type())) {
    if (openapi31) {
        schemaObject.setTypes(new LinkedHashSet<>(Arrays.asList(schema.type())));
    } else {
        schemaObject.setType(schema.type());
    }
}
```

**Before / After (OpenAPI 3.1)** — given `@Schema(type = "number") BigDecimal amount;`:

- **Before:** `amount: { type: "string" }`
- **After:** `amount: { type: "number" }`

This is a bug fix. It is distinct from #5061 / #5062 (numeric **example** serialization); it concerns the property **type** and reproduces with no `example` present. Fields relying on type inference (no explicit `type`) were already correct.

Fixes: #5233

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

Added `io.swagger.v3.core.resolving.v31.Ticket5233Test`, asserting that explicit `number` / `integer` / `boolean` types land in the `types` set under OpenAPI 3.1, with inferred-type and enum properties as controls. The full `swagger-core` module test suite passes locally (721 tests, 0 failures).

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
@@ -685,7 +685,11 @@ public abstract class AnnotationsUtils {
             schemaObject.set$ref(schema.ref());
         }
         if (StringUtils.isNotBlank(schema.type())) {
-            schemaObject.setType(schema.type());
+            if (openapi31) {
+                schemaObject.setTypes(new LinkedHashSet<>(Arrays.asList(schema.type())));
+            } else {
+                schemaObject.setType(schema.type());
+            }
         }
 
         if (schema.types().length > 0) {
--- a/modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/v31/Ticket5233Test.java
+++ b/modules/swagger-core/src/test/java/io/swagger/v3/core/resolving/v31/Ticket5233Test.java
@@ -1,0 +1,60 @@
+package io.swagger.v3.core.resolving.v31;
+
+import io.swagger.v3.core.converter.AnnotatedType;
+import io.swagger.v3.core.converter.ModelConverterContextImpl;
+import io.swagger.v3.core.jackson.ModelResolver;
+import io.swagger.v3.core.resolving.SwaggerTestBase;
+import io.swagger.v3.oas.models.media.Schema;
+import org.testng.annotations.Test;
+
+import java.math.BigDecimal;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Map;
+
+import static org.testng.Assert.assertEquals;
+
+/**
+ * An explicit {@code @Schema(type = "number" | "integer" | "boolean")} must be reflected in the
+ * OpenAPI 3.1 "types" set, not only in the legacy scalar "type" field. Previously the explicit type
+ * was applied via {@code setType} (scalar) while the "types" set kept the default {@code ["string"]},
+ * so the serialized 3.1 schema rendered every explicitly-typed field as {@code string}.
+ */
+public class Ticket5233Test extends SwaggerTestBase {
+
+    @Test
+    public void testExplicitSchemaTypeIsAppliedToTypesSetUnderOpenApi31() {
+        final ModelResolver modelResolver = new ModelResolver(mapper()).openapi31(true);
+        final ModelConverterContextImpl context = new ModelConverterContextImpl(modelResolver);
+        final Schema model = context.resolve(new AnnotatedType(Model.class));
+
+        final Map<String, Schema> properties = model.getProperties();
+        assertEquals(properties.get("amount").getTypes(), Collections.singleton("number"));
+        assertEquals(properties.get("count").getTypes(), Collections.singleton("integer"));
+        assertEquals(properties.get("flag").getTypes(), Collections.singleton("boolean"));
+
+        // controls: inferred type and enum are unaffected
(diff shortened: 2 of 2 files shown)


ewaostrowska:
Hi @forhadmethun, thank you for putting your time into this topic. As described in the https://github.com/swagger-api/swagger-core/issues/5233 we do not want to implement the bridging in between `type` and `types` for now. The behavior where all the scalar values for type default to string shall be fixed in the #5235.

ewaostrowska:
After revisiting the topic a small update here:
The changes from this PR were fetched to the https://github.com/swagger-api/swagger-core/pull/5304 with slight modification of the logic to prioritize the implementation when it is set. 

The bridging seems the simplest solution considering the `addTypeWhenSiblingsAllowed` needing to be fixed anyway since it is not only the OAS3.1 change - it also fires for the OAS 3.0 cases where siblings are allowed. 

Thank you for the PR @forhadmethun and your investigation of the topic :)
Further explanation [here](https://github.com/swagger-api/swagger-core/issues/5233#issuecomment-5451713571).

forhadmethun:
> After revisiting the topic a small update here: The changes from this PR were fetched to the #5304 with slight modification of the logic to prioritize the implementation when it is set.
> 
> The bridging seems the simplest solution considering the `addTypeWhenSiblingsAllowed` needing to be fixed anyway since it is not only the OAS3.1 change - it also fires for the OAS 3.0 cases where siblings are allowed.
> 
> Thank you for the PR @forhadmethun and your investigation of the topic :) Further explanation [here](https://github.com/swagger-api/swagger-core/issues/5233#issuecomment-5451713571).

Thanks for pulling my changes into #5304 and for the extra work on the implementation logic! Glad I could help out. 😊

PR #5237, opened 2026-07-15
fix: correct sibling logic when reading type from schema annotation

### LINKED ISSUE #5235: [Bug]: incorrect sibling logic when reading type from schema annotation

When defining @Schema(type = ...) on a model that is resolved to OAS 3.1, then the property is always marked as `type: string`. This is incorrect and it stems from a bug introduced in https://github.com/swagger-api/swagger-core/pull/4970. The issue is [this change](https://github.com/swagger-api/swagger-core/pull/4970/changes#diff-97fef6917ecf4e462a5f28a77ec994af6ccd7cce9fdde5a00dab818dbb994bf1R2929) and more precisely that `ctxSchema.type().getClass()` will always produce the `String` class.

## Affected Version

2.2.52

2.2.37

This was discovered in https://github.com/swagger-api/swagger-core/issues/5233 and thus reuses their steps to reproduce.

## Steps to Reproduce

```java
import java.math.BigDecimal;

public class Repro {

  static class Dto {
  }

    ResolvedSchema resolved =

  }
}
```

## Expected Behavior

```json
{
  "properties": {
  }
}
```

## Actual Behavior

```jsonc
{
  "properties": {
  }
}
```

## Checklist

Mattias-Sehlstedt:
# Pull Request

---

## Description

Introduces an intermediate fix for the issue reported in https://github.com/swagger-api/swagger-core/issues/5235. It does not solve the sibling handling in its entirety, since it only adjusts the type if `types` contains a single value (this is aimed to give a fallback for the logic implemented for OAS 3.0. This since `type` and `types` are now better isolated from each other, and the resolver expect the user to use the right field for the right OAS version). 

For OAS 3.1 it is currently not possible to entirely handle `types` since it is converted into a single type that is fed into `AnnotatedType`. For a real scenario the entire type array should be utilized and passed along.

This also comes with the behavior of coercing the string value for the `type` into a `PrimitiveType`. This might in turn lead to `format` or similar being calculated for `integer`, which might not always be what one expects when using `@Schema(types = {"integer"})`.

Fixes: https://github.com/swagger-api/swagger-core/issues/5235

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -798,7 +798,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
                 });
 
                 boolean areSiblingsAllowed = AnnotationsUtils.areSiblingsAllowed(resolvedSchemaResolution, openapi31);
-                aType = AnnotationsUtils.addTypeWhenSiblingsAllowed(aType, ctxSchema, areSiblingsAllowed);
+                aType = AnnotationsUtils.addTypeWhenSiblingsAllowed(aType, ctxSchema, openapi31, areSiblingsAllowed);
                 property = context.resolve(aType);
                 property = clone(property);
                 Schema ctxProperty = null;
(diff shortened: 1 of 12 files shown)


Mattias-Sehlstedt:
Closed since this goes further than solving the initial issue and instead also starts to try to implement proper type overriding. That is better left as separate issue since it should be a `feat` if it is even of interest to implement.

PR #5240, opened 2026-07-17
fix: correct sibling logic when reading type from schema annotation

### LINKED ISSUE #5235: [Bug]: incorrect sibling logic when reading type from schema annotation

Mattias-Sehlstedt:
# Pull Request

---

## Description

Introduces an intermediate fix for the issue reported in https://github.com/swagger-api/swagger-core/issues/5235. It does not implement fully fleshed type handling (overriding), but rather implements the smallest change possible while still keeping the previous behavior of being able to set `AnnotationType`'s type to `String`.

Since this changes how `@Schema(type = )` is read (the `type` is now not taken into consideration when resolving to OAS 3.1 in certain scenarios) it could be considered a breaking change. The fallback is to use `@Schema(types = )` when resolving to OAS 3.1. See later parts of [comment](https://github.com/swagger-api/swagger-core/issues/5233#issuecomment-4981108504) for a reasoning regarding why the application should not try to coerce the types into each other between OAS 3.0 and OAS 3.1, and that they should instead be unaware about each other.

A more extensive fix would be to clarify that this branching is not tied to sibling logic, but rather type coercion. But in order to implement that one would need to reason alot of what is the sought after behavior, especially for OAS 3.1 where there is several types with `@Schema(types = {"boolean", "string"})`. But all of this is considered out of scope for an immediate fix.

A known "bug" (that exists since before too) is that the underlying type might still leak data. For example with 
```java
@Schema(type = "boolean")
public Integer count;
```
becoming
```json
{
  "count": {
    "type": "boolean",
    "format": "int32"
  }
}
```
. This occurs since the `AnnotatedType` does not release the field `Integer` type when the user specifies `boolean`. And thus the `format` is still derived and present from resolving the schema-primitive `Integer`. But this is as previously stated more so tied to introducing full type overriding functionality.

Fixes: https://github.com/swagger-api/swagger-core/issues/5235

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -808,7 +808,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
                 });
 
                 boolean areSiblingsAllowed = AnnotationsUtils.areSiblingsAllowed(resolvedSchemaResolution, openapi31);
-                aType = AnnotationsUtils.addTypeWhenSiblingsAllowed(aType, ctxSchema, areSiblingsAllowed);
+                aType = AnnotationsUtils.addTypeWhenSiblingsAllowed(aType, ctxSchema, openapi31, areSiblingsAllowed);
                 property = context.resolve(aType);
                 property = clone(property);
                 Schema ctxProperty = null;
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/util/AnnotationsUtils.java
@@ -3125,14 +3125,32 @@ public abstract class AnnotationsUtils {
         return Schema.SchemaResolution.ALL_OF.equals(resolvedSchemaResolution) || Schema.SchemaResolution.ALL_OF_REF.equals(resolvedSchemaResolution) || openapi31;
     }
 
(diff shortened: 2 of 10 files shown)


ewaostrowska:
Hi @Mattias-Sehlstedt!
Thank you for your investigation and PR once again :) As I have explained in the [comment](https://github.com/swagger-api/swagger-core/issues/5233#issuecomment-5451713571) it seems that intentionally ignoring the @Schema(type) on the OAS 3.1 specification might not be the best approach. 

Part of this commit handling leaking enums on types={"string"} for OAS 3.1 will still be valid after the  https://github.com/swagger-api/swagger-core/pull/5304. For this reason I am not closing this PR for now.

PR #5250, opened 2026-07-20
refactor: add DiscriminatorResolver for handling discriminator properties in schemas

Mattias-Sehlstedt:
# Pull Request

---

## Description

Moves Discriminator resolving logic into a separate `Utils` class.

A commonly requested feature is to [extend the discriminator introspection](https://github.com/swagger-api/swagger-core/issues/3411). There [exists a PR](https://github.com/swagger-api/swagger-core/issues/3411) for this, but I would find it easier to reason about any change if the logic was isolated from the larger monster that is `ModelResolver`. This so that `ModelResolver` is not cluttered with more subdomain (managing discriminators) logic, and that the `ModelResolver` is instead only responsible for exposing an area of these subdomains if there is a reason to provide a helpful extension point.

I.e., `Utils` provide stateless logic for subdomains of handling specifications, and the `ModelResolver` mainly combine these smaller parts into something larger.

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/DiscriminatorResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/DiscriminatorResolver.java
@@ -1,0 +1,79 @@
+package io.swagger.v3.core.jackson;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.JavaType;
+import io.swagger.v3.core.converter.AnnotatedType;
+import io.swagger.v3.core.converter.ModelConverterContext;
+import io.swagger.v3.core.util.AnnotationsUtils;
+import io.swagger.v3.oas.annotations.media.DiscriminatorMapping;
+import io.swagger.v3.oas.models.media.Discriminator;
+import io.swagger.v3.oas.models.media.JsonSchema;
+import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.oas.models.media.StringSchema;
+import org.apache.commons.lang3.StringUtils;
+
+import static io.swagger.v3.core.util.RefUtils.constructRef;
+
+class DiscriminatorResolver {
+
+    private DiscriminatorResolver() {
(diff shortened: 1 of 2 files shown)


ewaostrowska:
Hi @Mattias-Sehlstedt! Thank you for working on improving the code in Swagger Core once again. 
We appreciate a lot any intent to make the Model Resolver any smaller :) That's definitely something that needs to be done.

One issue that I see with this refactor is that it is creating a new public API and it seems unintentional. `DiscriminatorUtils` is public, and both methods are public static methods. Once released, consumers may treat these methods as supported API, creating future compatibility obligations.

The `DiscriminatorUtils` in this form does not establish a strong public use case. I would keep it internal. 

Please consider moving it in the `io.swagger.v3.core.jackson` package and making it package-private.

Mattias-Sehlstedt:
I have moved it to the jackson package and made it package private. I have also added a private constructor to align it with the already existing "utils" class there (`JAXBAnnotationsHelper`). Should it be renamed to `Helper` too? I personally find `Utils` to be the more common and descriptive naming pattern.

The other common pattern is to utilize JavaDoc to explicitly state that the signature is volatile/subject to change/internal. But I could not find that the project has done that before so a bit odd to start with it now.

ewaostrowska:
@Mattias-Sehlstedt thank you for carrying out these changes. Since the project does not use JavaDoc annotations I do not see any reason to introduce it here. I would argue that even `Resolver` instead of `Util` would make sense here, but I am fine with merging it as it is.

ewaostrowska (approved):

Mattias-Sehlstedt:
I agree with the suggestion of changing it to `Resolver` and have adjusted to that. It makes it clearer that the `ModelResolver` is itself a resolver of resolvers that handle sub-subjects. 

It also aligns well with the fact that there already is one such complex area that has been isolated to its own `Resolver` (`TypeNameResolver`).

PR #5265, opened 2026-08-12
fix: _isSetType now detects indirect Set implementations

vpelikh:
### Problem

`_isSetType` in `AbstractModelConverter` failed to recognize indirect `Set` implementations such as `LinkedHashSet`, `HashSet`, and `TreeSet`.

The old implementation only matched an exact `Set.class` equality check, plus a direct interface scan via `Class.getInterfaces()`. Because `getInterfaces()` only returns directly implemented interfaces, subclasses that inherit `Set` behavior through a parent class (for example a custom class extending `AbstractSet` or a collection extending a concrete Set) were not detected, so `uniqueItems` was not set on the generated schema.

### Change

- Replaced `Set.class.equals(cls)` with `Set.class.isAssignableFrom(cls)` in `protected boolean _isSetType(Class<?> cls)`.
- Preserved the existing scala Set detection (checked via direct interfaces) to avoid pulling the scala runtime dependency.
- Made the null case explicit (returns `false`).

### Tests

Added `IsSetTypeTest` with 7 test cases:

- `null` -> not a Set
- `Set` interface -> Set
- `HashSet` -> Set
- `LinkedHashSet` -> Set
- `TreeSet` -> Set
- non-Set collections (`List`, `ArrayList`) -> not a Set
- non-collection types (`String`, `Integer`, `Map`) -> not a Set

All 7 tests pass.

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/AbstractModelConverter.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/AbstractModelConverter.java
@@ -114,17 +114,16 @@ public abstract class AbstractModelConverter implements ModelConverter {
     }
 
     protected boolean _isSetType(Class<?> cls) {
-        if (cls != null) {
-
-            if (java.util.Set.class.equals(cls)) {
+        if (cls == null) {
+            return false;
+        }
+        if (java.util.Set.class.isAssignableFrom(cls)) {
+            return true;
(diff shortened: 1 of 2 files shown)


ewaostrowska:
Thank you @vpelikh for providing this PR, looks like it should help with the Set classification

ewaostrowska (approved):

PR #5266, opened 2026-08-12
fix: deserialize OAS 3.1 scalar schema type into Set

### LINKED ISSUE #5264: [Bug]: JsonSchema fails with JSON round-trip

`JsonSchema` `type` field is asymmetric since there is a [custom serializer](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/mixin/Schema31Mixin.java#L65) that does not have a matching deserializer. This leads to a JSON round-trip failing when a single `type` is set.

## Affected Version

2.2.53

## Steps to Reproduce

1. Create an instance of `JsonSchema` where there is a single `type` defined
2. Serialize it
3. Deserialize it, and see that it fails with reading `type: X` into `types`

## Expected Behavior

Deserialization should work.

## Actual Behavior

Deserialization does not work.

## Checklist

kalayciburak:
# Pull Request

## Description

OpenAPI 3.1 serializes a single schema `type` as a JSON string (`"type":"integer"`) via `Schema31Mixin.TypeSerializer`, while the model field is `Set<String> types`. Deserializing that scalar back into `JsonSchema` failed with:

```
Cannot construct instance of java.util.HashSet ... from String value ('integer')
```

This adds a matching `TypeDeserializer` that accepts both the scalar form and the multi-type array form (`"type":["string","null"]`), so JSON round-trips and `convertValue` keep the type set.

Fixes #5264

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor (non-breaking change)
- [x] Tests
- [ ] Documentation
- [ ] Chore (build or tooling)

## Checklist

## Test plan

Executed:

```
./mvnw -pl modules/swagger-core -am test -Dtest=JsonSchemaTypeRoundTripTest,OpenAPI3_1DeserializationTest,SchemaDeserializationTest,JsonDeserializationTest -Dsurefire.failIfNoSpecifiedTests=false
# 41 tests, 0 failures

./mvnw -pl modules/swagger-core -am test -Dsurefire.failIfNoSpecifiedTests=false
# swagger-core: 752 tests, 0 failures
```

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/mixin/Schema31Mixin.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/mixin/Schema31Mixin.java
@@ -7,12 +7,18 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.util.LinkedHashSet;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/mixin/Schema31Mixin.java
```diff
@@ -42,6 +48,7 @@ public abstract class Schema31Mixin {
 
     @JsonProperty("type")
     @JsonSerialize(using = TypeSerializer.class)
```
Mattias-Sehlstedt, line 50: Note that the serializer is used in an additional location and that it might be of relevance to attach the deserializer there as well.

kalayciburak: yep, added TypeDeserializer on SortedSchemaMixin31 too

PR #5312, opened 2026-09-03
fix: ignore a nullable annotation for enums that use enumAsRef

### LINKED ISSUE #5310: [Bug]: @Nullable property mutates shared enumAsRef component schema in OpenAPI 3.1
# [Bug]: @Nullable property mutates shared enumAsRef component schema in OpenAPI 3.1

## Description

When an enum annotated with `@Schema(enumAsRef = true)` is referenced from both a nullable and a non-nullable property, processing the `@Nullable` property makes the shared enum component schema nullable globally.

This means that other references to the same enum are also effectively nullable, even if the corresponding property is explicitly annotated with `@NotNull`.

The issue is reproducible with OpenAPI 3.1 and appears to be a regression introduced in swagger-core 2.2.44.

## Minimal example

Shared enum:

```java
@Schema(enumAsRef = true)
public enum StatusType {
    NOT_REQUESTED,
    REQUESTED,
    APPROVED,
    REJECTED
}
```

Nullable request property:

```java
public class VehicleUpdateRequestDto {

    @Nullable
    @Schema(
        description = "New status.",
        requiredMode = Schema.RequiredMode.NOT_REQUIRED
    )
    private StatusType status;
}
```

Non-null response property:

```java
public class VehicleResponseDto {

    @NotNull
    @Schema(
        description = "Current status.",
        defaultValue = "NOT_REQUESTED"
    )
    private StatusType status;
}
```

Both properties reference the same component schema because of `enumAsRef = true`.

## Actual behavior

With swagger-core 2.2.44:

```json
"StatusType": {
  "type": [
    "string",
    "null"
  ],
  "enum": [
    "NOT_REQUESTED",
    "REQUESTED",
    "APPROVED",
    "REJECTED"
  ]
}
```

The response property is nevertheless correctly listed as required:

```json
"VehicleResponseDto": {
  "type": "object",
  "properties": {
    "status": {
      "$ref": "#/components/schemas/StatusType",
      "default": "NOT_REQUESTED",
      "description": "Current status."
    }
  },
  "required": [
    "status"
  ]
}
```

The problem is that the referenced shared component itself now allows `null`.

## Expected behavior

Property-level `@Nullable` should not mutate the shared enum component schema.

The shared component should remain:

```json
"StatusType": {
  "type": "string",
  "enum": [
    "NOT_REQUESTED",
    "REQUESTED",
    "APPROVED",
    "REJECTED"
  ]
}
```

Nullability of the request property should be represented without changing the shared component schema used by other non-null references.

## Regression

The behavior was reproduced with the same minimal application by changing only the swagger-core version:

| swagger-core | `components.schemas.StatusType` | Result |
| --- | --- | --- |
| 2.2.41 | `"type": "string"` | OK |
| 2.2.43 | `"type": "string"` | OK |
| 2.2.44 | `"type": ["string", "null"]` | Affected |
| 2.2.55 | `"type": ["string", "null"]` | Still affected |

- **Last known good version:** 2.2.43
- **First known affected version:** 2.2.44
- **Still reproducible with:** 2.2.55

## Possible relation to existing issues

This regression coincides with the introduction of native `@Nullable` support for OpenAPI 3.1 in:

- #5001
- #5018 — Add support for `@Nullable` annotations in OpenAPI 3.1 schemas

In #5018, nullable properties started adding `"null"` to the schema type set for OpenAPI 3.1.

A similar shared-schema mutation was subsequently reported for object schemas and addressed by:

- #5077
- #5115
- #5124 — fix: ensure that `@Nullable` does not incorrectly affect object schemas

This issue appears to be the equivalent case for shared **enum component schemas** created with `@Schema(enumAsRef = true)`.

## Reproducer

A minimal Spring Boot reproducer is attached.

Environment:

- Java 17
- Spring Boot 3.5.16
- SpringDoc 2.9.0
- OpenAPI 3.1
- swagger-core-jakarta 2.2.44

The attached OpenAPI outputs demonstrate the regression directly:

- `swagger-core-2.2.43-schema-reproducer.json` — last known good
- `swagger-core-2.2.44-schema-reproducer.json` — first affected

The relevant difference is:

```diff
"StatusType": {
-  "type": "string",
+  "type": [
+    "string",
+    "null"
+  ],
   "enum": [...]
}
```

## Steps to reproduce

1. Download and extract the attached reproducer.

2. Run:

```bash
mvn clean spring-boot:run
```

3. Open:

```text
http://localhost:8080/v3/api-docs
```

4. Inspect:

```text
components.schemas.StatusType
```

5. With `swagger-core-jakarta` 2.2.44, the generated component contains:

```json
"type": ["string", "null"]
```

6. Change only the Maven property:

```xml
<swagger-core.version>2.2.43</swagger-core.version>
```

7. Restart the application and inspect the same component.

It now contains:

```json
"type": "string"
```

The regression can therefore be reproduced by changing only swagger-core from 2.2.43 to 2.2.44.

## Dependency verification

The resolved Swagger dependencies can be verified with:

```bash
mvn dependency:tree -Dincludes=io.swagger.core.v3
```

`swagger-core-jakarta`, `swagger-annotations-jakarta`, and `swagger-models-jakarta` should all resolve to the version being tested.

## Attachments

Image
[swagger-nullable-enum-shared-schema-reproducer.zip](https://github.com/user-attachments/files/31800759/swagger-nullable-enum-shared-schema-reproducer.zip)
[swagger-core-2.2.43-schema-reproducer.json](https://github.com/user-attachments/files/31800757/swagger-core-2.2.43-schema-reproducer.json)
[swagger-core-2.2.44-schema-reproducer.json](https://github.com/user-attachments/files/31800758/swagger-core-2.2.44-schema-reproducer.json)

Mattias-Sehlstedt:
# Pull Request

---

## Description

Fixes so that an enum that uses `enumAsRef = true` ignores any `@Nullable` annotation when processing for `nullable`. This since the project does not yet support nullable `refs`.

I elected to place it within the `resolveSchemaMembers` next to `resolveNullable` rather than analyzing an enum when it is created as a ref in to make it easier to analyze and adjust the nullable resolving when nullable `refs` become supported.

Fixes: #5310

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

### DIFF
--- a/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
+++ b/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/ModelResolver.java
@@ -129,6 +129,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
 
     private static final int SCHEMA_COMPONENT_PREFIX = "#/components/schemas/".length();
     private static final String OBJECT_TYPE = "object";
+    private static final String NULL_TYPE = "null";
 
     private static final Predicate<Annotation> ANNOTATIONS_THAT_SHOULD_BE_STRIPPED_FOR_CONTAINER_ITEMS = annotation ->
             annotation.annotationType().getName().startsWith("io.swagger") ||
@@ -235,7 +236,7 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
                     schema.specVersion(SpecVersion.V31);
                 }
                 resolveArraySchema(annotatedType, schema, resolvedArrayAnnotation);
-                Schema itemsSchema = openapi31 ? new JsonSchema() : new Schema();
+                Schema itemsSchema = getNewSchema();
                 return schema.items(itemsSchema.$ref(resolvedSchemaAnnotation.ref()).name(name));
             }
         }
@@ -286,12 +287,12 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
                         // create a reference for the items
                         if (context.getDefinedModels().containsKey(innerSchema.getName())) {
                             String ref = constructRef(innerSchema.getName());
-                            innerSchema = openapi31 ? new JsonSchema() : new Schema();
+                            innerSchema = getNewSchema();
                             innerSchema.$ref(ref);
                         }
                     } else if (innerSchema != null && innerSchema.get$ref() != null) {
                         String ref = StringUtils.isNotEmpty(innerSchema.get$ref()) ? innerSchema.get$ref() : innerSchema.getName();
-                        innerSchema = openapi31 ? new JsonSchema() : new Schema();
+                        innerSchema = getNewSchema();
                         innerSchema.$ref(ref);
                     }
                 }
@@ -303,12 +304,12 @@ public class ModelResolver extends AbstractModelConverter implements ModelConver
                     // create a reference for the items
                     if (context.getDefinedModels().containsKey(implSchema.getName())) {
                         String ref = constructRef(implSchema.getName());
-                        implSchema = openapi31 ? new JsonSchema() : new Schema();
+                        implSchema = getNewSchema();
                         implSchema.$ref(ref);
                     }
                 } else if (implSchema != null && implSchema.get$ref() != null) {
                     String ref = StringUtils.isNotEmpty(implSchema.get$ref()) ? implSchema.get$ref() : implSchema.getName();
-                    implSchema = openapi31 ? new JsonSchema() : new Schema();
+                    implSchema = getNewSchema();
                     implSchema.$ref(ref);
                 }
                 return implSchema;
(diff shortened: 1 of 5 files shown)


dpkass:
Thanks for addressing the shared-component mutation. I do not think completely ignoring `@Nullable` for `enumAsRef` is semantically correct for OpenAPI 3.1, though: it fixes the leakage but documents a genuinely nullable property as non-nullable.

The property can preserve its nullability without changing the shared enum component:

```yaml
status:
  anyOf:
    - $ref: "#/components/schemas/StatusType"
    - type: "null"
```

`StatusType` itself should remain `type: string`. `oneOf` would also be valid here because the enum and `null` branches are disjoint.

Could the resolver wrap the reference at the property site instead of discarding nullability, potentially through the overridable/isolated nullable-ref mechanism discussed in #5124? Ignoring the annotation is safer than mutating the component, but it still leaves the generated contract inaccurate.

Mattias-Sehlstedt:
It is intentional to leave out the nullable-ref logic from this PR, for the same reason as in the linked PR.

This is a directed fix towards leaking data due to a `@Nullable` implementation that did not consider the full scope. and the full scope is an implementation that should be formalized and implemented in its separate issue.

Issue #4991, opened 2025-10-10
[Feature] Support Jackson 3.x release line


reta:
## Feature Description

Jackson has released major new 3.x version [1], this is major breaking change [2] comparing to 2.x release line, including both package and artifact changes.  

[1] https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.0
[2] https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-1

## Use Case

Swagger tightly integrates with Jackson in core (not to mention other components) and it may not be possible to support both versions (2.x and 3.x) at the same time. 

## Suggested Solution (optional)

Thinking about possible options, having new Swagger 3.x release line would be probably the best way to move forward.

## Alternatives Considered

Support both versions (2.x and 3.x) at the same time. It may not be possible, or otherwise maintainable.

## Checklist

- [ ] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) to ensure this is not a duplicate.
- [ ] This feature would be useful to more than just my use case.
- [ ] I have provided enough detail for the maintainers to understand the scope of the request.

nnam-droid12:
Thanks for opening this, @reta  this seems like a significant upcoming change since Jackson 3.x introduces breaking API differences.  
I’m following this discussion with interest; if the maintainers decide to move forward, I’d be happy to help with testing or migration work.

reta:
Thanks a lot @nnam-droid12 , I would love to get guidance from maintainers before doing any work there

nnam-droid12:
Sounds good @reta . I’ll keep an eye on this as well. Once the maintainers share their guidance on the preferred approach for Jackson 3.x support, I’d be happy to work along side you with implementation or review if needed.

bnasslahsen:
+1, this is also needed for [springdoc-openapi project ](https://springdoc.org/) since Spring-Boot-4 has fully transitioned to [Jackson 3 as its primary implementation](https://spring.io/blog/2025/10/07/introducing-jackson-3-support-in-spring). 

It would be great to see Swagger Core adopt the modern Jackson version as well!

vpelikh:
Is anyone already working on this issue?

reta:
> Is anyone already working on this issue?

Not yet, it would be great to get +1 from maintainers before investing time

yuanjinyong:
Is there any progress?

vpelikh:
> Is there any progress?

[...]

PR #5314, opened 2026-09-09
Update to jackson 3

daniel-kmiecik:
# Pull Request

## Description

This pull request upgrades the project to use Java 17 as the minimum required version and migrates the codebase from the legacy `com.fasterxml.jackson` Jackson 2 library to the new `tools.jackson` library (Jackson 3.x). It also updates the GitHub Actions workflows to use newer Java versions, updates dependencies, and refactors serializers to match the new Jackson API. These changes modernize the codebase, improve compatibility, and ensure ongoing support.

**Java Version and Build System Updates**
- Changed the minimum required Java version from 11 to 17 in documentation and all GitHub Actions workflows; workflows now test/build against Java 17 and 25 instead of 11 and 17. (`README.md`, `.github/workflows/codeql-analysis.yml`, `.github/workflows/maven.yml`, `.github/workflows/maven-pulls.yml`, `.github/workflows/prepare-release.yml`, `.github/workflows/release.yml`) 

**Jackson 3.x Migration**
- Replaced all imports and usages of `com.fasterxml.jackson` with `tools.jackson` throughout the codebase to support Jackson 3.x, and updated dependency group IDs and versions in `pom.xml`. (`modules/swagger-core/pom.xml`, multiple Java files)
- Updated Jackson-related code to use new APIs, such as `TypeFactory.createDefaultInstance()`, `serializationConfig()`, and updated serializer and introspection methods to match Jackson 3.x conventions. (`ModelConverters.java`, `AbstractModelConverter.java`) 

**Serializer Refactoring**
- Refactored custom serializers to extend `tools.jackson.databind.ValueSerializer` instead of the old `JsonSerializer`, and updated method signatures and logic to use the new Jackson 3.x serialization API. (`ApiResponsesSerializer.java`, `CallbackSerializer.java`, `ExampleSerializer.java`, `MediaTypeSerializer.java`)

**Dependency and Compatibility Updates**
- Updated Jackson dependencies to version 3.2.2 and changed group IDs to `tools.jackson.*` in `pom.xml`. Also added `jakarta.annotation-api` as a provided dependency. (`modules/swagger-core/pom.xml`)
- Removed the explicit version for `jakarta.xml.bind-api`, likely to use the version managed by a parent or BOM. (`modules/swagger-core/pom.xml`)

**Minor Code Adjustments**
- Updated exception handling to use `tools.jackson.core.JacksonException` instead of `IOException` for Jackson serialization errors. (`SpecFilter.java`, serializer classes)

Fixes: #4991 

[...]

PR #5320, opened 2026-09-18
Add @JsonSubTypes polymorphic schema generation with oneOf, dedup and WRAPPER_OBJECT support

vpelikh:
### Why this PR

The parent model of a polymorphic type was not composed: types annotated with `@JsonSubTypes` or `@Schema(subTypes)` were resolved as independent models but the parent schema got no reference to its subtypes at all. The parent needs a `oneOf` (with `$ref`s) to express "this is an abstract/polymorphic type" so tools generate correct client/codegen hierarchies and discriminators.

### Problem (root causes)

- `resolveSubtypes()` resolved subtypes independently but never composed the parent into a `oneOf` schema, and it bypassed `@JsonView`-suffixed model names.
- Jackson 3's `AnnotationIntrospectorPair` can return duplicate subtype entries, so the `oneOf` would contain repeated (invalid) entries.
- `@JsonView`-suffixed model names were ignored when building `oneOf` `$ref` entries (`getSimpleName()` was used instead).
- `WRAPPER_OBJECT` wrapping was skipped entirely when `oneOf` was present, hiding the nested wrapper property.
- `@Schema(oneOf)` annotation processing ran even when `@JsonSubTypes` already produced a `oneOf`, causing a double/duplicated `oneOf`.

### Change

- `resolveSubtypes()` detects `@JsonSubTypes` / `@Schema(subTypes)`, deduplicates subtypes via `LinkedHashSet`, resolves each subtype with `allOf` composition referencing the parent, and replaces the parent model with a `ComposedSchema` containing `oneOf`.
- `@JsonView`-aware refs: subtype schemas are resolved with the `jsonViewAnnotation` and their resolved names stored in a map; `oneOf` `$ref` entries use those names.
- `resolveWrapping()` fix: when the model has `oneOf`, the wrapper property (`@JsonTypeInfo(include=WRAPPER_OBJECT)`) is added directly alongside `oneOf` instead of nesting the whole model inside a wrapper schema.
- Early skip: `@Schema(oneOf)` annotation processing is skipped when `@JsonSubTypes` is present, preventing a duplicate `oneOf`.

**Result:** parent schemas expose `oneOf` referencing their resolved subtypes; subtypes use `allOf` composition; `@JsonView`-suffixed names resolve correctly in `oneOf` refs; `WRAPPER_OBJECT` properties coexist with `oneOf`.

**Verification:** new `JsonSubTypesAndSchemaOneOfTest` covers `@JsonSubTypes`+`@Schema(oneOf)`, interface implementation, and `@JsonSubTypes`-only interface; full `swagger-core` module suite passes.

Related: #4991 

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 🧪 Tests
- [ ] 📝 Documentation

## Checklist


vpelikh:
CC @Mattias-Sehlstedt

Mattias-Sehlstedt:
Could we share what tools that prefer a cyclic polymorphic structure? As shared in https://github.com/swagger-api/swagger-core/pull/5314#issuecomment-5623981034 my experience is that most actually prefer it to not exists, and that even OAS documentation has previously documented it as "the two different polymorphic structures (e.g., the `oneOf` and the `allOf` approach).

To me
> Early skip: @Schema(oneOf) annotation processing is skipped when @JsonSubTypes is present, preventing a duplicate oneOf

feels like the opposite of how the annotations function in general? Isn't the general idea that first heuristics are used to derive the OAS structure, but that we then provide these annotations so that a user can force the structure to follow something that the user explicitly would like to correct? Now we instead force the generation to follow the internal interpretation, and ignore any explicit structure that the user has asked for? Basically copying what happened in https://github.com/springdoc/springdoc-openapi/issues/2915?

If someone would like to restore the previous behavior, have we thought about making that easily-achievable? I would recommend ensuring that there are good extensions points for it, and that it is illustrated with a test similar to https://github.com/swagger-api/swagger-core/pull/5030.

vpelikh:
Thanks for the detailed review, @Mattias-Sehlstedt. I've addressed both of your points with two follow-up commits on this branch.

## 1. Extension point to restore the previous behavior (easy opt-out)

`resolveSubtypes()` was `private`; it's now `protected`, so a downstream `ModelResolver` subclass can override it to return `false` and opt out of the automatic `@JsonSubTypes` → `oneOf` composition — restoring the previous behavior. This is illustrated with a test:

- `JsonSubTypesAndSchemaOneOfTest.extensionPointAllowsSubclassToRestorePreviousBehavior`

which shows a subclass override preventing the `oneOf` conversion (same pattern suggested in PR #5030).

## 2. Explicit `@Schema(oneOf)` now wins over `@JsonSubTypes`

Agreed on the philosophy point. The previous "early skip" inverted user intent: it skipped the explicit `@Schema(oneOf)` whenever `@JsonSubTypes` was present, forcing the internal composition over the user's explicit structure.

I changed this so:

- The explicit `@Schema(oneOf)` is **always** processed, even when `@JsonSubTypes` is present (the early-skip was removed).
- `resolveSubtypes()` now returns early when a class declares its own `@Schema(oneOf)`, so the explicit composition takes precedence over the `@JsonSubTypes`-derived one.

This also removes the cyclic `allOf <-> oneOf` structure: when a class declares `@Schema(oneOf)`, the parent keeps only its shared discriminator property, and subtypes become plain object schemas (no recursive `allOf` → parent ref).

## On "which tools prefer a cyclic polymorphic structure"

The original motivation for the `@JsonSubTypes` → `oneOf` combination was to move away from the `allOf`-heavy form (including requests for `@JsonSubTypes`-driven generation). I agree the cyclic `allOf` + `oneOf` hybrid is often undesirable. With `@Schema(oneOf)` now winning, users who want the explicit non-cyclic form get it; the auto-composition remains only for `@JsonSubTypes`-only types, and users can always opt out via the `resolveSubtypes` extension point.

Happy to take further direction if you'd prefer a different default.

Mattias-Sehlstedt:
What should be default I will leave to swagger-core to decide. I am very happy as long as there is an easy way to control the behavior with configurations/extensions. 

As a human I prefer the more logical `allOf` <-> `oneOf`, since it makes sense to the parts to be aware of each other. But I have gotten used to it always being exactly one of them, since swagger-ui and openapi-generator both dislike cyclic structures (I have not tried scalar to see if it handles it properly).

I can see if I have time to investigate further the history behind it and what might actually be considered best-practice. Do note that any findings should not impact this PR in any way, since it gives full control to the user, and thus findings will not impact code but at most recommendations for users.

Mattias-Sehlstedt:
I've searched around a bit and all I can find point to the best approach being to **not** have them together. If one wants to create a "link" between the parent and the children, the best way to express that is with the `discriminatorMapping` (i.e., basically what is done in https://github.com/swagger-api/swagger-core/pull/4983).

One can for example query de-detoxified Stack overflow (read: AI) with a biased prompt "I want to strengthen the relation between parent and child with `oneOf` + `allOf` because..." and it will generally even advice you against doing it even if you are biased and insistent on it being correct. This issue with it being that the `oneOf` becomes entirely a validation constraint, and that the validation cycle introduced cannot terminate (A `Dog` is a `Pet` and a `Pet` needs to be a `Dog`...). The recommendation instead is a self-contained `oneOf` union, e.g.,
```yaml
Pet:
  oneOf:
    - $ref: '#/components/schemas/Dog'
    - $ref: '#/components/schemas/Cat'
  discriminator:
    propertyName: petType
    mapping:
      dog: '#/components/schemas/Dog'
      cat: '#/components/schemas/Cat'
```

So my personal suggestion would be to not make `oneOf` + `allOf` as the default, but instead rather make it an easy opt-in for those that find it beneficial.

vpelikh:
@Mattias-Sehlstedt, thanks for your input! Now let's wait for a reply from the team regarding the @Schema/@JsonSubTypes allOf/oneOf relationship.

PR #5325, opened 2026-09-21
feat: Set Swagger mapper defaults after the Jackson 3 migration

ewaostrowska:
# Pull Request

Follow-up: #4991 

This change sets the mapper behavior that Swagger Core needs to keep stable after the Jackson 3 migration. It restores selected Jackson 2 behavior and states Swagger Core's existing choices explicitly.

## Defaults set by this PR
This change sets the mapper behavior that Swagger Core needs to keep stable after the Jackson 3 migration. It restores selected Jackson 2 behavior and states Swagger Core's existing choices explicitly.

## Defaults set by this PR

| Behavior | Jackson enum or mechanism | Value on this branch | Reason |
|---|---|---|---|
| Unknown input properties | `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` | Ignore them. | Swagger Core already accepted input with extra properties. |
| Empty beans | `SerializationFeature.FAIL_ON_EMPTY_BEANS` | Write `{}`. | Swagger Core already allowed empty objects. |
| Dates | `DateTimeFeature.WRITE_DATES_AS_TIMESTAMPS` | Write text instead of timestamps. | This keeps Swagger Core's date output policy. |
| Null values | `JsonInclude.Include.NON_NULL` for values and content | Exclude null bean properties and null map entries. | This keeps the existing output policy. Null list elements keep their positions. |
| Bean property order | `MapperFeature.SORT_PROPERTIES_ALPHABETICALLY` | Keep the established order instead of sorting names. | This limits unrelated changes in generated documents. |
| `BigDecimal` output | `StreamWriteFeature.WRITE_BIGDECIMAL_AS_PLAIN` | Write plain numbers without exponent notation. | This keeps the existing output policy. |
| Decimal tree scale | `JsonNodeFeature.STRIP_TRAILING_BIGDECIMAL_ZEROES` | Strip insignificant trailing zeroes from decimal tree nodes. For example, `1.2300` becomes `1.23` with scale `2`. | This restores the Jackson 2 default for tree values created from `BigDecimal`. Direct `BigDecimal` binding keeps its input scale. |
| `Duration` output | `DateTimeFeature.WRITE_DURATIONS_AS_TIMESTAMPS` | Write numeric timestamps. | This keeps Jackson 2's numeric output, such as `2.234` seconds. |
| UTC offset text | `DateTimeFeature.WRITE_UTC_AS_OFFSET` | Write `+00:00` for classic UTC `Date` and `Calendar` values. | This keeps the Jackson 2 text form for a zero offset. |
| `null` for primitives | `DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES` | Accept `null` and convert to the primitive default (`0`, `false`, etc.). | Swagger Core previously accepted null input for primitive fields. Rejecting it would break existing documents. |
| Ordered maps with incomparable keys | `SerializationFeature.FAIL_ON_ORDER_MAP_BY_INCOMPARABLE_KEY` | Fail when key ordering is requested. | The mapper must not silently return an unsorted map. |

The strict generic mapper keeps the shared Swagger rules for unknown properties, empty beans, dates, nulls, and decimal tree scale. It keeps declaration property order and rejects trailing root content. The normal output rules for plain `BigDecimal`, `Duration`, UTC offset text, and ordered maps do not apply to it. Its zero UTC offset remains `Z`.

## Real-world impact

Without these settings, Jackson 3's changed defaults would silently alter serialized OAS output. Each setting guards a concrete failure mode:

| Setting | Failure without it |
|---|---|
| `STRIP_TRAILING_BIGDECIMAL_ZEROES` | `minimum: 1.50` instead of `minimum: 1.5` — breaks spec consumers that compare numeric values |
| `FAIL_ON_EMPTY_BEANS=false` | Serializing extension objects or custom model classes with no properties throws at runtime |
| `WRITE_DATES_AS_TIMESTAMPS=false` | Date fields in `Example` values emit `1234567890` instead of `"2024-01-01T00:00:00Z"` |
| `FAIL_ON_UNKNOWN_PROPERTIES=false` | Deserializing user-supplied OAS docs with vendor extensions or future spec fields throws |
| `WRITE_DURATIONS_AS_TIMESTAMPS=true` | Duration values in extensions switch from numeric seconds (`2.234`) to ISO-8601 (`PT2.234S`) — changes wire format |
| `WRITE_UTC_AS_OFFSET=true` | `OffsetDateTime` fields emit `Z` instead of `+00:00` — produces diff noise in generated specs |
| `SORT_PROPERTIES_ALPHABETICALLY=false` | Properties reorder alphabetically instead of following declaration order defined by the spec |
| `FAIL_ON_ORDER_MAP_BY_INCOMPARABLE_KEY=true` | `LinkedHashMap` with non-`Comparable` keys silently returns unsorted output instead of failing fast |
| `WRITE_BIGDECIMAL_AS_PLAIN=true` | `minimum`, `maximum`, `multipleOf` values serialize as `1.5E+2` instead of `150` |
| `FAIL_ON_NULL_FOR_PRIMITIVES=false` | Deserializing OAS documents with `null` on a primitive field throws — breaks existing documents that supply `null` for numeric or boolean values |
| `withFirstCharAcceptance(true, true)` | Single-char getters like `getX()` on OAS model beans are not discovered as properties |

## Mapper and caller behavior

The rules apply to Swagger-created JSON, YAML, OpenAPI 3.0, OpenAPI 3.1, and converter mappers. 

## Scope
Other Jackson 3 defaults remain in effect, but some more defaults will be reconsidered. 

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🧪 Tests
- [ ] 📝 Documentation

## Checklist

- [ ] I have added/updated tests as needed
- [ ] The PR title is descriptive
- [ ] The code builds and passes tests locally
- [ ] I have linked related issues (if any)


modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/AbstractModelConverter.java
```diff
@@ -43,8 +42,7 @@ protected AbstractModelConverter(ObjectMapper mapper, TypeNameResolver typeNameR
                     public void setupModule(SetupContext context) {
                         context.insertAnnotationIntrospector(new SwaggerAnnotationIntrospector());
                     }
-                })
-                .configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, false);
```
daniel-kmiecik: Why did you change this?
It should stay as 'false'.
In Jackson 2 'false' was a default value. It's 'true' in Jackson 3.

modules/swagger-core/src/main/java/io/swagger/v3/core/util/ObjectMapperFactory.java
```diff
@@ -238,13 +242,9 @@ public ValueSerializer<?> modifySerializer(
             sourceMixins.put(Discriminator.class, Discriminator31Mixin.class);
         }
         mapperBuilder.addMixIns(sourceMixins);
-        mapperBuilder.configure(StreamWriteFeature.WRITE_BIGDECIMAL_AS_PLAIN, true);
-        mapperBuilder.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, false);
```
daniel-kmiecik: Same as above

modules/swagger-core/src/main/java/io/swagger/v3/core/util/ObjectMapperFactory.java
```diff
@@ -292,25 +292,43 @@ public static ObjectMapper createJsonConverter() {
 
         sourceMixins.put(Schema.class, SchemaConverterMixin.class);
         builder.addMixIns(sourceMixins);
-        builder.configure(StreamWriteFeature.WRITE_BIGDECIMAL_AS_PLAIN, true);
-        builder.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, false);
```
daniel-kmiecik: Same as above

daniel-kmiecik (approved):

